PR #361, opened 2026-07-24
ci: triage-as-bestaxbot, safe robobun-style repro drafting + read-only security scan

allxsmith:
## Description

GitHub Actions / CI automation only (no product code). Four commits, building up from the triage-author fix that motivated the rest:

1. **`ci: post AI triage comments as bestaxbot`** — triage dedupe comments now post under the bestaxbot machine identity (robobun-style) instead of the anonymous github-actions[bot]; `auto-close-duplicates.mjs` + the triage command marker-checks learn the new author.
2. **`ci: add author-only repro and security-scan workflows`** — the two new automations below.
3. **`ci: guard AI entry points against re-trigger and flagged content`** — repo-wide hardening.
4. **`docs: document repro and security-scan automation`**.

- [x] Other (please specify): **GitHub Actions workflows / CI + AI automation**

**The two new automations (robobun-style, hardened for GitHub-hosted runners):**

- **`claude-repro.yml`** (author-only, 4 jobs, **PAT-free**) — a triage+ maintainer labels an issue `claude-repro`; Claude **drafts** a minimal Jest reproduction test and `github-actions[bot]` posts it for a human to run. Nothing executes the draft.
- **`ai-scan.yml`** (separate workflow) — read-only Claude scan of new issues/PRs for malicious code, prompt injection, and social engineering; applies `needs-security-review` **fail-closed**. Holds no PAT and no write tools; the flag is **advisory** (pauses `claude-repro`/`claude-fix`, does **not** block `@claude`/`@bestaxbot`).

**Two security invariants drive the design.** A red-team of the first-draft two-job repro found it would leak `CLAUDE_CODE_OAUTH_TOKEN` to a public comment (secret-masking scrubs logs, not comment bodies) and let an `@mention` re-trigger a write-capable session (`contains()` matches the raw substring):

- **I1** — the model-auth token never shares a job with code execution (repro is author-only; the drafting job has no Bash/gh/network/Task tool, so it can't read env, fetch, or run anything).
- **I2** — no attacker/Claude free text reaches a re-trigger-capable comment (drafts are deterministically sanitized and posted via `GITHUB_TOKEN`, whose comments never re-trigger workflows).

**Repo-wide guards (commit 3):** `claude.yml` bestaxbot sender-exclusion (also closes the re-trigger class for the shipped triage path) + `needs-security-review` refusal; `bestaxbot-reply.yml` / `claude-implement.yml` flag refusal; harden-runner + kill-switch header on `ai-triage.yml`.

## Related Issue(s)

Refs #362

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [x] Documentation
- [x] Build tooling
- [x] Other (please describe): **CI / AI-automation security**

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works — _`scripts/auto-close-duplicates.test.mjs` now covers the auto-close decision logic (15 tests, mutation-checked), and root `pnpm test` runs `node --test "scripts/*.test.mjs"` so root scripts are no longer invisible to CI. The two shell parsers (publish sanitizer, scan-verdict parser) remain untested — they are embedded in workflow YAML and need extracting first: #454. Workflows validated below and need a canary run (see Additional Context)_
- [ ] I have updated Storybook stories as needed (bulma-ui) — _n/a, no UI change_
- [x] My changes require a change to the documentation
- [ ] All new and existing tests passed — _no product tests affected; workflows verified by YAML lint + logic unit tests_
- [ ] Any relevant dependencies are updated — _n/a_

## Screenshots / Demos

n/a — CI automation.

## Additional Context

**Verification done:** YAML valid on all 18 workflows; the publish sanitizer unit-tested (machine markers broken, `@mentions` defanged, legitimate imports untouched); the scan verdict parser tested fail-closed against clean/flagged/crashed/missing/forged inputs; prettier clean; no tabs; commitlint passed all commits.

**Two rollout notes:**

1. **harden-runner on the existing `ai-triage.yml` job ships in `audit`, not `block`.** Flipping a live workflow straight to egress-block risks breaking it if `claude-code-action`'s runtime egress needs an un-allow-listed host (it isn't cleanly documented). Audit monitors without blocking. **Follow-up:** after one real triage run, confirm the endpoint list and flip to `block`. The new jobs already run in `block`.
2. **Canary before relying on these** (event-triggered, security-critical): on a maintainer-authored issue, apply `claude-repro` (confirm no PAT in the jobs, the draft posts as github-actions[bot], the label auto-removes); open a benign fake-injection issue (confirm it flags with **no comment posted**) and a normal issue (confirm it scans clean).

**New labels needed:** `claude-repro` (green `#0e8a16`), `needs-security-review` (red `#b60205`). Both were created ahead of merge, so neither path depends on a label being created on demand.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Added automated, read-only security scans that can flag items for security review.
  - Added an author-only reproduction workflow that drafts tests without executing them.
- **Bug Fixes**
  - Improved duplicate detection and triage marker recognition across current and legacy automation accounts.
  - Prevented automated fix, reply, and investigation workflows from running during security review.
- **Documentation**
  - Updated guidance for reproduction tests, security scans, automation states, and triage behavior.
---

## State at merge (updated after review rounds)

**The security scanner does not start on merge.** `AI_SCAN_MODE` is now explicit opt-in — it
runs only when set to `on` or `y`, and every other value (`off`, unset, empty, a typo)
disables it. Enable it deliberately once you are ready; merging alone changes nothing. Same for
`claude-repro`, which needs its label applied by a triage+ user.

**Labels exist**: `claude-repro` and `needs-security-review` were created ahead of merge, so
neither feature is inert on arrival, and neither depends on a label being created on demand.
(An earlier revision claimed the REST add-labels endpoint auto-creates a missing label. GitHub
does not document that either way, so the claim was removed rather than inverted — deep review
finding 3.)

**Changes made in response to review**, beyond the original four commits:

- Closed the `Read`-based token-exfil path in `claude-repro` (`Collect draft` refuses a draft
  carrying a live credential, literal or base64/hex encoded), and rewrote I1 to state plainly
  that the encoding check is a backstop rather than a proof.
- Scoped the repro comment edit by marker. `--edit-last` could overwrite an unrelated
  `github-actions[bot]` comment — including a historical `<!-- ai-triage:dedupe -->` marker
  that `auto-close-duplicates.mjs` still reads — destroying an auto-close candidate.
- Corrected several security comments that claimed more than the mechanism delivered: the
  `needs-security-review` blast radius (it gates every entry point *this repo controls*; third
  party reviewers are not gated, and there is an open-time window), and the scan session's
  token, which is write-scoped rather than read-only.
- Documented every AI repository variable and its unset default, after finding
  `AI_LOOP_ENABLED` had been documented backwards.

**Final review round** added `.github/CLAUDE.md` — the security contract for anything in
`.github/workflows/`. Copilot's `allowedTools` advisory asked for a comment at one call site;
the durable form is a document covering every workflow, so it states the two invariants,
allowlists as a confinement boundary that never widens casually, SHA pinning, `== 'on'` opt-in
gating, fail-closed verdicts vs fail-open budgets, marker-scoped comment edits, and a reviewer
checklist. Both root `CLAUDE.md` and this PR's other lessons feed it. Same commit also:

- Aligned `actions/checkout` in `ai-scan.yml` and `claude-repro.yml` to the repo-wide pin. Both
  sat on `9c091bb` (**v7.0.0**, twelve commits behind) while the other nineteen usages were on
  `3d3c42e5` (**v7.0.1**) — and both were commented `# v7`, so the drift was invisible to a
  reader. Now exactly one SHA repo-wide.
- Fixed the `auto-close-duplicates.mjs` header, which still described the objection rule as "no
  non-bot user" after the logic moved to `isAutomationAuthor()`.

**`SECURITY.md` and `README.md` describe the scanner as active** while `AI_SCAN_MODE` ships
off. That is deliberate: the variable is turned on shortly after merge, and the steady state is
on, so wording the docs around a temporary off-state would need reverting immediately.

**Known gaps, all filed:** #454 (extract and test the two shell parsers), #455 (split `ai-scan`
so the model session runs with `contents: read`), #457 (publish triage comments from a
structured payload — this PR moves triage to a PAT-authored identity, whose comments *can*
re-trigger workflows, so I2 is currently held by every consumer excluding bestaxbot rather than
by construction; raised by CodeRabbit and verified), and #456 (the two post-merge obligations —
flipping `ai-triage` harden-runner from `audit` to `block` after one real run, and the canary
checklist for both new automations). #456 exists because those two were previously recorded
only in this description, which a squash merge discards.

### DIFF
--- a/.claude/commands/triage-dedupe.md
+++ b/.claude/commands/triage-dedupe.md
@@ -16,9 +16,11 @@ run locally), and `AUTOCLOSE` (`active` or `off`; assume `off` locally). If
 
 1. `gh issue view NUMBER --repo REPO --json state,title,body,comments` — if
    the issue is not open, stop.
-2. Marker check — does any existing bot-authored comment contain
-   `<!-- ai-triage:dedupe -->`? (Match the marker + a bot author, not a
-   specific login — the workflow posts as github-actions[bot].)
+2. Marker check — does any existing comment authored by bestaxbot or a
+   bot account contain `<!-- ai-triage:dedupe -->`? (Match the marker +
+   that author class, never one specific login — the workflow posts as
+   bestaxbot today; older comments are from github-actions[bot] or
+   claude[bot].)
    - `TRIGGER=opened` and marker present → stop (already triaged).
    - `TRIGGER=labeled` and marker present → continue; at the end REFRESH
      that comment instead of posting a new one: if it is your most recent
--- a/.claude/commands/triage-dedupe.md
+++ b/.claude/commands/triage-dedupe.md
@@ -23,10 +23,14 @@ run locally), and `AUTOCLOSE` (`active` or `off`; assume `off` locally). If
    claude[bot].)
    - `TRIGGER=opened` and marker present → stop (already triaged).
    - `TRIGGER=labeled` and marker present → continue; at the end REFRESH
-     that comment instead of posting a new one: if it is your most recent
-     comment on the issue, use
-     `gh issue comment NUMBER --repo REPO --edit-last --body ...`;
-     otherwise post a fresh comment (the old one stays as history).
+     that comment instead of posting a new one. Use
+     `gh issue comment NUMBER --repo REPO --edit-last --body ...` ONLY when
+     your most recent comment on the issue is itself the
+     `<!-- ai-triage:dedupe -->` comment: `--edit-last` selects by author,
+     not by marker, and every automation here posts as bestaxbot now, so a
+     newer repro draft or other machine comment would be the one it
+     overwrites. If anything else is newer, post a fresh comment (the old
+     one stays as history). See `.github/CLAUDE.md` rule 6.
 3. If the issue is too vague to search meaningfully (no error text, no
    component or file name, no concrete behavior), stop.
 
--- a/.claude/commands/triage-find-duplicate-prs.md
+++ b/.claude/commands/triage-find-duplicate-prs.md
@@ -14,9 +14,11 @@ assume `labeled` locally). If `NUMBER` is missing, ask.
 
 1. `gh pr view NUMBER --repo REPO --json state,title,body,files,comments` —
    if the PR is not open, stop.
-2. Marker check — a bot-authored comment containing
-   `<!-- ai-triage:find-duplicate-prs -->` (match marker + bot author, not
-   a specific login — the workflow posts as github-actions[bot]):
+2. Marker check — a comment authored by bestaxbot or a bot account
+   containing `<!-- ai-triage:find-duplicate-prs -->` (match marker + that
+   author class, never one specific login — the workflow posts as
+   bestaxbot today; older comments are from github-actions[bot] or
+   claude[bot]):
    - `TRIGGER=opened` and marker present → stop.
    - `TRIGGER=labeled` and marker present → continue; at the end refresh
      that comment (`gh pr comment NUMBER --repo REPO --edit-last --body ...`
--- a/.claude/commands/triage-find-duplicate-prs.md
+++ b/.claude/commands/triage-find-duplicate-prs.md
@@ -21,8 +21,13 @@ assume `labeled` locally). If `NUMBER` is missing, ask.
    claude[bot]):
    - `TRIGGER=opened` and marker present → stop.
    - `TRIGGER=labeled` and marker present → continue; at the end refresh
-     that comment (`gh pr comment NUMBER --repo REPO --edit-last --body ...`
-     if it is your most recent comment on the PR; otherwise post fresh).
+     that comment with
+     `gh pr comment NUMBER --repo REPO --edit-last --body ...` ONLY when
+     your most recent comment on the PR is itself the
+     `<!-- ai-triage:find-duplicate-prs -->` comment. `--edit-last` selects
+     by author, not by marker, and the other triage commands post as the
+     same account — so a newer `find-issues` marker is what it would
+     overwrite. Otherwise post fresh. See `.github/CLAUDE.md` rule 6.
 
 ## Search — 3 parallel agents
 
--- a/.claude/commands/triage-find-issues.md
+++ b/.claude/commands/triage-find-issues.md
@@ -14,9 +14,11 @@ EXACTLY ONE comment (or nothing). Context from the caller: `REPO`, `NUMBER`
 
 1. `gh pr view NUMBER --repo REPO --json state,title,body,comments` — if the
    PR is not open, stop.
-2. Marker check — a bot-authored comment containing
-   `<!-- ai-triage:find-issues -->` (match marker + bot author, not a
-   specific login — the workflow posts as github-actions[bot]):
+2. Marker check — a comment authored by bestaxbot or a bot account
+   containing `<!-- ai-triage:find-issues -->` (match marker + that
+   author class, never one specific login — the workflow posts as
+   bestaxbot today; older comments are from github-actions[bot] or
+   claude[bot]):
    - `TRIGGER=opened` and marker present → stop.
    - `TRIGGER=labeled` and marker present → continue; at the end refresh
      that comment (`gh pr comment NUMBER --repo REPO --edit-last --body ...`
--- a/.claude/commands/triage-find-issues.md
+++ b/.claude/commands/triage-find-issues.md
@@ -21,8 +21,13 @@ EXACTLY ONE comment (or nothing). Context from the caller: `REPO`, `NUMBER`
    claude[bot]):
    - `TRIGGER=opened` and marker present → stop.
    - `TRIGGER=labeled` and marker present → continue; at the end refresh
-     that comment (`gh pr comment NUMBER --repo REPO --edit-last --body ...`
-     if it is your most recent comment on the PR; otherwise post fresh).
+     that comment with
+     `gh pr comment NUMBER --repo REPO --edit-last --body ...` ONLY when
+     your most recent comment on the PR is itself the
+     `<!-- ai-triage:find-issues -->` comment. `--edit-last` selects by
+     author, not by marker, and the other triage commands post as the same
+     account — so a newer `find-duplicate-prs` marker is what it would
+     overwrite. Otherwise post fresh. See `.github/CLAUDE.md` rule 6.
 3. Note every issue already referenced in the PR body (`#N`, `Fixes #N`,
    `Closes #N`, full URLs) — those are EXCLUDED from the results.
 
--- a/.github/CLAUDE.md
+++ b/.github/CLAUDE.md
@@ -1,0 +1,177 @@
+# .github — CI and AI automation
+
+Everything in `.github/workflows/` is **human-authored by design**. AI agents working in this
+repository are denied write access to this directory (`Edit(.github/**)` and friends sit in
+every write agent's `--disallowedTools`), and the AI fix loop refuses any PR that touches
+`.github/**`. If you are an agent reading this: propose a diff in a comment, do not apply one.
+
+This file is the security contract for these workflows. The rules below are not style
+preferences — each one is load-bearing, and most were written after a review round or a
+red-team found the failure it prevents. Where a rule has a documented origin, it is cited.
+
+## The two invariants
+
+Every AI workflow here is built to preserve these. If a change breaks one, the change is wrong,
+regardless of how convenient it is.
+
+- **I1 — the model-auth token never shares a job with code execution.** `CLAUDE_CODE_OAUTH_TOKEN`
+  pays for a session; anything that can run code in the same job can read it out of the
+  environment. `claude-repro.yml` splits drafting from publishing for exactly this reason: the
+  drafting job has no `Bash`, no `Task`, no network tool, so it cannot read env, fetch, or
+  execute. Note the direction of the guarantee — it raises the cost of exfiltration, it is not a
+  proof. The credential-shape check in `Collect draft` (literal, base64, and hex forms) is a
+  backstop for the same reason, and is documented as one.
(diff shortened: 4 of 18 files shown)


allxsmith:
## Rebased and review findings addressed

Was 70 commits behind (branch point 2026-07-24). Rebased onto main — no conflicts.

### 🟠 Blocking #1 — I1's token-exfil path. Fixed, and the finding was right.

I1 claimed the `author` job's lack of a Bash tool meant an injected session "cannot read
process.env". Removing Bash stops it shelling out, but **`Read` is not confined to the
workspace** — `/proc/self/environ` and `~/.claude/.credentials.json` stay readable, `Write` puts
what it finds into the draft, and `publish` posts that draft to a public comment. The one job
holding `CLAUDE_CODE_OAUTH_TOKEN` is exactly the one feeding a public publish, so the channel the
invariant advertised as closed was open.

`Collect draft` now **refuses** to emit a draft containing a live credential value (compared
against the tokens the job actually holds) or a credential-shaped string (a re-encoded token, or
a secret this job cannot compare against).

Refuse rather than redact, deliberately: a credential in the draft means the session went
somewhere it had no business going, and quietly publishing a scrubbed copy would destroy the
only signal that it happened. The job fails, `cleanup` still removes the label so the button
does not wedge, and the reason stays in the private run log.

I1 is rewritten to say what removing Bash does and does not buy, and to name this check as
load-bearing — not belt-and-braces. Without it the invariant does not hold.

Verified against the actual logic:

| draft contains | result |
|---|---|
| a clean test | allowed |
| the exact OAuth token | **refused** |
| the exact job token | **refused** |
| a *different* `sk-ant-` key | **refused** |
| `github_pat_…` | **refused** |
| a `PRIVATE KEY` block | **refused** |

### 🟡 Blocking #2 — the flag blocks more than the docs said. Fixed.

`CLAUDE.md` and the AI-development guide both said `needs-security-review` "does not block
`@claude`/`@bestaxbot`" — but `claude.yml:49-50` and `bestaxbot-reply.yml:55-56` gate their job
`if:` on that label, so it does. Wrong in the safer direction, but it would send someone hunting
a broken workflow when the label is the answer.

Corrected in both files and in the label table, keeping the accurate half of the original claim:
a **clean verdict** is advisory (it only covers the text as scanned at open time) while the
**flag** is a gate. Added the practical consequence — if `@claude` seems to ignore a mention,
check for the label first.

### 🔵 Advisory #3 — fence break-out. Fixed.

The publish sanitizer now defangs ``` and ~~~ at line start, so a drafted fence can no longer
close the wrapping ```tsx block and spill the rest as rendered markdown.

### 🔵 Advisory #4 — no committed tests. Not addressed.

Still true: the sanitizer, the fail-closed verdict parser and `isAutomationAuthor` have no
in-repo regression guard. That wants a test harness for shell/workflow logic that does not exist
here yet, so I have left it rather than bolt on something token. Worth its own issue — say the
word and I will file it.

All three workflows parse (`claude-repro.yml`, `ai-scan.yml`, `claude.yml`), and the sanitizer
was exercised against fences and mentions directly.

allxsmith:
@coderabbitai Full Review again

allxsmith:
## Review round 3 — including two corrections to my own commits

Three reviewers converged on the same header, and Copilot caught two things I got wrong. Taking
them in order of consequence.

### 🔴 `AI_LOOP_ENABLED` was documented backwards — my error

Copilot on `ai-development.md:178` and `CLAUDE.md:107`. My table said unset means **enabled**.
Every workflow gates on `vars.AI_LOOP_ENABLED == 'true'`, so unset means **disabled**. I
verified across all of `.github/workflows/` rather than re-reading my own note.

That is the worst row to get wrong: someone deleting the variable expecting a no-op would have
disabled the system — or, reading the table, trusted it was off while it was on.

Since I had already got one row wrong I re-derived all seven from the gates rather than trusting
the rest. The other six were right, and the useful thing the audit surfaced is that the gates
come in two styles, which decides whether *deleting* a variable is a kill switch or an
accidental enable:

- `== 'true'` — `AI_LOOP_ENABLED`, `AI_LOOP_COPILOT` → **unset is off**
- `!= 'off'` — `AI_SCAN_MODE`, `AI_TRIAGE_MODE` → **unset is on**, so deleting them *enables* the
  feature

The practical warning still stands, but for the right reason: `AI_SCAN_MODE` unset is enabled
only while the master switch is `true`, which it currently is.

### 🟠 The flag's blast radius was overstated — also mine

CodeRabbit (Major, `ai-development.md:59`) and the deep review's advisory 2, from two angles at
the same overreach. Correcting an *understatement* two commits ago, I overshot into "blocks every
AI entry point". Two limits, both now documented:

- **Third-party reviewers are not gated.** CodeRabbit and Copilot run on their own triggers and
  never see the label — CodeRabbit's own point, and it is right about itself.
- **There is an open-time window.** The scanner and its consumers both fire on `opened`, so a
  same-event path evaluates its `if:` before the label exists. Bounded by every such path
  independently requiring a trusted author, but a window, not a barrier.

### 🟡 `ai-scan.yml`'s header — flagged by all three reviewers

The one place left in the PR claiming the flag does *not* block `@claude`/`@bestaxbot`. I fixed
`CLAUDE.md` and the guide last round and missed the workflow header. Rewritten, and it now
carries both limits above rather than a flat claim in either direction.

### 🟠 The credential backstop was defeatable by encoding

Copilot on `claude-repro.yml:273`, and a fair hit on the fix I wrote: it compared the literal
token only, so base64 or hex round-tripped straight through. Now checks both encodings of the
live values:

| draft contains | before | after |
|---|---|---|
| literal token | refused | refused |
| base64 of the token | **allowed** | refused |
| hex of the token | **allowed** | refused |

I1 now says plainly that this is *not* a proof — an encoding check can never be complete
(chunk it, reverse it, rot13 it) — and that the invariant's strength is the combination of no
execution primitive, harden-runner egress block, and the token never sharing a job with the
publish step, with this check as the last line rather than the first. That is a more honest
claim than the one Copilot was pushing back on.

### Not taken

**`node --test` directory target** (Copilot, `package.json:9`). The concern is reasonable but the
suggested fix is worse here — `node --test scripts/` tries to execute the *directory* as a module
and fails outright (`MODULE_NOT_FOUND`). The quoted glob is the portable form: the quotes stop
shell expansion so Node expands the pattern itself, which is exactly why it finds all 15 tests,
and it matches what `docs/` already runs in CI.

**`author-only` clarification** (CodeRabbit nitpick). Already covered — the label table says
"author-only — CI does not run it" on the same line.

**Deep-review advisory 3** (fail-closed scanner over-flags on infra hiccups, wedging entry points
until a human clears it). Accepted trade-off as designed; worth watching after rollout rather
than changing now.

---

Verified: both workflows parse; the encoded-credential check refuses literal, base64 and hex of
each live token while allowing a clean test; 15 script tests pass; prettier clean.

allxsmith:
@claude another deep review

allxsmith:
## Deep review round 2 addressed

### 🟡 Blocking — unscoped `--edit-last`. Fixed, and it was a real data-loss path.

`--edit-last` edits whatever the token's identity posted most recently on that issue, which is
not necessarily a repro draft. github-actions[bot] also authored the dedupe comments that
predate the move to bestaxbot, and `auto-close-duplicates.mjs` still reads those — it matches
any *automation author*, not one login. So a re-run could overwrite a dedupe marker and destroy
an auto-close candidate along with its `Duplicate of #N`.

Now scoped by marker: find our own last `<!-- ai-repro:draft -->` comment and `PATCH` it, else
post. Tested against the exact shape the finding describes:

| comments on the issue | selected |
|---|---|
| dedupe marker (id 1), our draft (id 2), same marker from another author (id 3), **dedupe marker (id 4, last)** | **id 2** |
| only a dedupe marker | empty → posts a new comment |

The old code would have hit id 4 — the dedupe marker — which is precisely the loss described.

### 🔵 Advisory 3 — write-scoped token. Comment corrected; structural fix filed as #455.

This one had a false claim inside it, so it was not purely advisory: the header said the session
holds "a read GITHUB_TOKEN". The job grants `issues: write` + `pull-requests: write` for the
budget marker and the labeler, and the session is read-only *only* because the Bash allowlist
admits nothing that writes. One control reading as two — widening `allowedTools` later hands an
injected session write access with no permission diff for a reviewer to catch.

The header now says where read-only actually comes from and marks the allowlist as
security-critical rather than cosmetic. The real fix — split gate/scan/label so the session job
drops to `contents: read`, the same pattern `claude-repro.yml` and `test-deploy.yml` already
use — is **#455**. Restructuring a 260-line security workflow at merge time without its own
review pass trades more risk than it removes.

### 🔵 Advisory 4 — the gate fails open. Documented.

The *verdict* fails closed; the *gate* does not. A budget read/charge error or a spent daily
limit means no scan and no flag, so an unscanned item is indistinguishable from a clean one
downstream. Deliberate — a counter problem must not wedge every incoming issue — but the header
advertised only the fail-closed half. Both directions are now stated.

### 🔵 Advisory 2 — encoded exfil. Already owned.

I1 says outright that the encoding check is not a proof and that the invariant rests on the
combination of no execution primitive, egress block, and token/publish separation. The review
agrees this is explicitly owned by the PR; no change.

---

Both workflows parse; the marker selector is verified against the failure case above.

Worth flagging for the merge decision: this is the third review round, and each one has found
something real in the round before it — an overstated invariant, then a backwards default and a
defeatable check, now a data-loss path and a false token claim. All in comments and docs *about*
security properties rather than in the mechanisms themselves, which is its own signal. Nothing
outstanding that I know of, but I would not treat my own confidence as the deciding evidence
here.

allxsmith:
@coderabbitai full review

allxsmith:
@coderabbitai full review

Issue #405, opened 2026-07-29
[Security] main lacks required status checks and required reviews — two in-repo comments claim a gate that doesn't exist


allxsmith:
## Summary

`main`'s branch protection is materially thinner than several in-repo files assume. Two
committed comments describe a review gate that **does not exist**. Before #403 adds README copy
advertising "multiple reviews before merge," the underlying gate has to be real.

## Evidence

`gh api repos/allxsmith/bestax/branches/main/protection` returns only:

```json
{
  "required_signatures": { "enabled": true },
  "enforce_admins":      { "enabled": false },
  "allow_force_pushes":  { "enabled": false },
  "allow_deletions":     { "enabled": false },
  "required_conversation_resolution": { "enabled": false }
}
```

There is **no `required_status_checks`** and **no `required_pull_request_reviews`**.
`gh api repos/allxsmith/bestax/rulesets` returns `[]`, so nothing is enforced via the newer
ruleset path either.

## What that contradicts

1. **`.coderabbit.yaml:10-12`** —
   > Keep false: a CodeRabbit APPROVE could satisfy main's required-1-approval rule, which would
   > gut the human merge gate of the AI loop (issue #215).

   `request_changes_workflow: false` is defending a required-1-approval rule that isn't
   configured. The setting is still correct to keep — but its stated rationale is currently fiction.

2. **`.github/CODEOWNERS:3-4`** — honestly notes it's inert until "Require review from Code
   Owners" is enabled. It is still inert. `/.github/` — described in that same file as "the
   repo's security boundary" — is unprotected by review.

3. **Nothing enforces green CI at the branch level.** A red PR is merge-able today. The AI loop
   has its own convergence gate (`claude-pr-loop.yml:184-197`) requiring `Build and Test`,
   `React N compatibility`, and `Dependency Review` — but that governs only the loop, not merges
   in general.

## Resolution

Apply classic branch protection to `main`:

- **Required status checks** — the same four the AI loop already treats as the definition of
  green: `Build and Test`, `React 18 compatibility`, `React 19 compatibility`, `Dependency Review`.
  `strict: false` (no forced rebase-before-merge).
- **Required approvals: 1**, with `dismiss_stale_reviews: true`.

### Checks deliberately NOT required

- `Publish to npm` — only runs on `main` pushes; requiring it would hang every PR.
- `Test and Preview Deployment` — currently fails on every dependabot PR (#398). Requiring it
  before that's fixed would block all dependency updates.
[...]

Issue #404, opened 2026-07-29
[Security] Adopt OpenSSF Scorecard — workflow, published results, and badge


allxsmith:
## Summary

Add the [OpenSSF Scorecard](https://scorecard.dev) action and publish the resulting score as a
badge. Scorecard is the industry-standard, third-party-verifiable rating of exactly the
supply-chain posture this repo already invests in — it turns "we say we're hardened" into a
number anyone can independently check.

Split out of #403, which deliberately scoped Scorecard out because it touches `.github/**`.

## Why this repo scores well already

Scorecard grades ~18 checks. We already satisfy most of the expensive ones without any new work:

| Scorecard check | Our current state |
| --- | --- |
| `Pinned-Dependencies` | 100% of `uses:` across all 17 workflows are full-SHA pinned |
| `Token-Permissions` | `permissions: contents: read` at workflow level, write re-declared per job (`ci.yml:13-14`) |
| `Signed-Releases` | `publishConfig.provenance` on all published packages + GPG-signed release commits (`ci.yml:209-222`) |
| `Dependency-Update-Tool` | Dependabot, weekly, grouped (`.github/dependabot.yml`) |
| `SAST` | CodeQL (default setup) on JS/TS **and** the workflow files |
| `Security-Policy` | `SECURITY.md` with a disclosure process and contact |
| `Vulnerabilities` | `pnpm audit --audit-level=high` gate in CI (`ci.yml:75-76`) |
| `License` | MIT, declared |
| `Dangerous-Workflow` | no `pull_request_target`; fork PRs never triaged (see #312) |

## What will drag the score down

Being honest up front so the number isn't a surprise:

- **`Branch-Protection`** — tracked separately in the companion issue. Until required status
  checks land on `main`, this check scores low.
- **`Code-Review`** — Scorecard looks for evidence that changes were reviewed before merge. Our
  reviews are AI bots (CodeRabbit + the Claude deep review) plus the owner's merge; Scorecard
  only counts GitHub review approvals, so bot reviews that don't leave an approval don't register.
- **`Fuzzing`** — not applicable to a React component library; expect a 0 here and ignore it.
- **`CII-Best-Practices`** — no OpenSSF best-practices badge; optional, separate effort.

Realistic initial overall: **~7–8 / 10**. It is a live badge and will move on its own as checks
and dependencies change.

## Implementation

New file `.github/workflows/scorecard.yml`, following house conventions (SHA-pinned actions with
[...]

Issue #403, opened 2026-07-29
[Docs] Advertise our supply-chain hardening — Socket.dev + provenance badges and a "Hardened by default" README section


allxsmith:
## Documentation Area

- [x] Other: **README files** (all four) + `SECURITY.md`

---

## Page(s) or Section(s) Affected

| File | Published to npmjs.com? | Current badges | Current security section |
| --- | --- | --- | --- |
| `README.md` (root) | No — GitHub only | 8 | none |
| `bulma-ui/README.md` | **Yes** — the `@allxsmith/bestax-bulma` package page | 7 | none |
| `create-bestax/README.md` | **Yes** — the `create-bestax` package page | 3 | none |
| `bestax-migrate/README.md` | Yes, once published | **0** | none |
| `SECURITY.md` | — | — | see item 3 & 4 below |

---

## Describe the Issue or Suggestion

This project runs an unusually hard supply-chain posture, and **none of it is visible where
people actually decide whether to install a dependency.** All four READMEs have zero security
badges and zero security content, even though `SECURITY.md` and
[the Security guide](https://bestax.io/docs/guides/security) already document the whole thing.

`bulma-ui/README.md` and `create-bestax/README.md` *are* the npmjs.com package pages. That is
the highest-leverage real estate we have for "this dependency is safe to add," and today it says
nothing about provenance, scanning, or review.

**Socket.dev is the sharpest gap.** It is running — every PR gets two passing checks
(`Socket Security: Project Report` and `Socket Security: Pull Request Alerts`, org `bestax`) —
but it is configured entirely in the Socket dashboard. There is no `socket.yml`, no CLI step,
no mention in any file in this repo. Neither a contributor nor an evaluator reading the source
can tell it exists. **CodeQL has the same problem**: it runs via GitHub's default setup, so
there is no `codeql.yml` to point at.

---

### 1. Badges

Socket.dev badge format, verified against live public usage in
[`t3-oss/t3-env`](https://github.com/t3-oss/t3-env/blob/main/packages/nextjs/README.md) and
[`ccusage/ccusage`](https://github.com/ccusage/ccusage/blob/main/apps/ccusage/README.md) —
scoped names work unencoded:

```markdown
[![Socket Badge](https://socket.dev/api/badge/npm/package/@allxsmith/bestax-bulma)](https://socket.dev/npm/package/@allxsmith/bestax-bulma/overview)
[...]

PR #406, opened 2026-07-29
ci: add OpenSSF Scorecard workflow with published results

allxsmith:
Closes part of #404 (the workflow half — the badge lands in the #403 README PR).

## What

Adds `.github/workflows/scorecard.yml`. Runs the [OpenSSF Scorecard](https://scorecard.dev) analysis weekly, on pushes to `main`, and on `branch_protection_rule` changes; publishes results so the public viewer and the README badge have data; uploads SARIF to the Security tab alongside CodeQL.

## Why

Scorecard is a third-party, independently verifiable rating of exactly the posture this repo already invests in. It turns "we say we're hardened" into a number anyone can check — which is the whole premise of #403.

## Notes on the choices

- **No `pull_request` trigger.** Experimental upstream and unsupported on forks.
- **`publish_results: true`** with `id-token: write` — without both, the badge has no data at all.
- **`permissions: read-all`** at workflow level, with the three write scopes re-declared on the job only. Same least-privilege shape as `ci.yml:13-14` (which uses `contents: read`); `read-all` is the value upstream recommends for Scorecard, which needs broad read access to score the repo.
- **All four actions SHA-pinned** with `# vN` comments, per house convention. `actions/checkout` reuses the exact pin already in `ci.yml`; `ossf/scorecard-action` v2.4.4 and `github/codeql-action/upload-sarif` v4 tags were dereferenced to their commit SHAs.
- **Weekly schedule is deliberate** — it keeps scoring after merge, so a regression surfaces without anyone remembering to look.

## Expected score

Roughly **7–8 / 10** initially. High on `Pinned-Dependencies`, `Token-Permissions`, `Signed-Releases`, `SAST`, `Dependency-Update-Tool`, `Security-Policy`. Dragged down by `Fuzzing` (not applicable to a component library — expect 0, ignore it) and `Code-Review` (Scorecard counts GitHub review approvals; our reviews are AI bots that don't leave approvals).

`Branch-Protection` should now score better — required status checks and 1 required approval were applied to `main` under #405.

## Verification

- [x] `prettier --check` passes
- [x] YAML parses; triggers, job permissions, and all four SHA pins confirmed
- [ ] Workflow runs green **after merge to `main`** (the `push` trigger can't fire from a PR branch)
- [ ] Results appear at https://scorecard.dev/viewer/?uri=github.com/allxsmith/bestax
- [ ] SARIF appears in the Security tab

**Merge this before the #403 README PR** — the Scorecard badge 404s until this has run on `main` at least once.

## Summary by CodeRabbit

* **Chores**
  * Added automated OpenSSF Scorecard security analysis on a weekly schedule and whenever changes are pushed to the main branch.
  * Security analysis results are now published to code scanning for visibility and review.

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -33,7 +33,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           persist-credentials: false
 
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,60 @@
+name: OpenSSF Scorecard
+
+# Third-party, independently verifiable rating of this repo's supply-chain
+# posture (pinned actions, token permissions, signed releases, SAST, branch
+# protection...). Publishing the results is what makes the README badge and the
+# public scorecard.dev viewer work — see issue #404.
+#
+# No pull_request trigger: that trigger is experimental upstream and is not
+# supported on forks. The weekly schedule is the point — it keeps scoring us
+# after merge, so a change that regresses the posture surfaces without anyone
+# remembering to look.
+on:
+  branch_protection_rule:
+  schedule:
+    # Mondays 07:41 UTC — off the hour to avoid GitHub's scheduling backlog.
+    - cron: '41 7 * * 1'
+  push:
+    branches: [main]
+
+permissions: read-all
+
+jobs:
+  analysis:
+    name: Scorecard analysis
+    runs-on: ubuntu-latest
+    permissions:
+      # Upload the SARIF so findings land in the Security tab next to CodeQL.
+      security-events: write
+      # Required by publish_results: mints the OIDC token that proves to the
+      # OpenSSF API that this result really came from this repo's CI.
+      id-token: write
+      contents: read
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+        with:
+          persist-credentials: false
+
+      - name: Run analysis
+        uses: ossf/scorecard-action@2d1146689b8cda280b9bc96326124645441f03bc # v2.4.4
+        with:
+          results_file: results.sarif
+          results_format: sarif
+          # Required for the badge and the public viewer to have any data.
+          publish_results: true
+
+      # Retained separately from code scanning: the Security tab prunes old
+      # alerts, and a raw SARIF is easier to diff when investigating a score drop.
+      - name: Upload artifact
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: SARIF file
+          path: results.sarif
+          retention-days: 5
+
+      - name: Upload to code scanning
+        uses: github/codeql-action/upload-sarif@e4fba868fa4b1b91e1fdab776edc8cfbe6e9fb81 # v4
+        with:
+          sarif_file: results.sarif


allxsmith:
Thanks — went through all three. Two are actionable, one I'm pushing back on with evidence.

### Finding 1 (🟡 Minor, pin drift) — **disputing the premise**

Half right: `9c091bb2` *is* v7.0.0, not the latest v7.0.1. But the claim that v7.0.1 is "the repo-wide pin" doesn't hold:

```
$ grep -rhoE 'actions/checkout@[a-f0-9]{40} # v[0-9.]+' .github/ | sort | uniq -c
  16 actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
```

All **16** occurrences across `.github/` are on `9c091bb2`. Nothing in this repo uses `3d3c42e5`. So the PR body's "reuses the exact pin already in `ci.yml`" is accurate, and bumping only this new file to v7.0.1 would *introduce* the drift the finding wants to avoid — it'd be 16 files on v7.0.0 and 1 on v7.0.1.

Repo-wide bumps are Dependabot's job (`.github/dependabot.yml` has the `github-actions` ecosystem on a weekly grouped schedule), so v7.0.1 will arrive everywhere at once. **No change made.**

### Finding 2 (🔵 Advisory, PR-body drift) — **fixed**

Correct, and my error: `ci.yml:13-14` is `contents: read`, not `read-all`. PR body reworded — `read-all` is still the right value here (it's what upstream recommends for Scorecard, which needs broad read access to score the repo), but it isn't "the same pattern."

### Finding 3 (🔵 Advisory, job permissions replace rather than merge) — **on the record, no change**

Agreed on the mechanics and on the conclusion for today. Worth stating the trigger explicitly for whoever reads this later: if this repo ever goes private, the analysis job needs `actions: read` added or Scorecard silently under-reports. Public repo, so correct as written.

allxsmith:
@coderabbitai review

allxsmith:
**Retracting my earlier pushback on finding 1 — the review was right and I was wrong.** Fixed in the latest push.

The pin is now `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7`, matching `main`.

### How I got it wrong

My rebuttal cited this:

```
$ grep -rhoE 'actions/checkout@[a-f0-9]{40} # v[0-9.]+' .github/ | sort | uniq -c
  16 actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
```

That grep ran against the **working tree**, which had been switched to an unrelated feature branch based on an older `main`. So I sourced the pin from a stale branch and then "verified" it against the same stale branch — the confirmation was circular and meant nothing.

Checked properly against `origin/main`:

```
$ git show 5d114e1:.github/workflows/ci.yml | grep -oE 'actions/checkout@[a-f0-9]{40}'
actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
```

`3d3c42e5` in all 16 places on `main`, zero occurrences of `9c091bb2`. The bump landed in #392. My branch point (`5d114e1`) already had the correct pin — this PR was introducing the *only* stale pin in the repo, which is exactly what the finding said.

### Knock-on fix

The same stale-branch mistake hit #411, which additionally had stale `setup-node` (v6 → v7) and `pnpm/action-setup` (v4 → v6.0.9) pins. Corrected there too. Both branches rebased onto current `main`.

Findings 2 and 3 stand as previously addressed.

allxsmith:
deep-review: RE-REVIEW — verify the prior finding was properly addressed. Head is now rebased onto main.

Your previous review flagged (🟡 Minor) that `actions/checkout` was pinned to v7.0.0 (9c091bb2) rather than main’s pin. The author initially DISPUTED this with a grep claiming all 16 pins were 9c091bb2, then retracted — the grep had run against a stale feature branch, not main. The pin is now claimed to be 3d3c42e5.

Verify independently, treating every claim in the PR body and comments as unverified:
1. Resolve EVERY action pin in scorecard.yml via the GitHub API. Confirm each is a real commit (the author separately fabricated a nonexistent SHA on the sibling PR #411, so do not assume any SHA is valid).
2. Confirm each pin matches what main currently uses for that action, where main uses it.
3. For actions new to this repo (ossf/scorecard-action, github/codeql-action/upload-sarif), confirm the SHA corresponds to the version in the trailing comment.
4. Re-check the two prior advisories: the `read-all` top-level permission, and job-level permissions replacing rather than merging.
5. Confirm nothing about the rebase broke the workflow (triggers, cron validity, publish_results + id-token pairing).

allxsmith:
@claude Please re-review this PR to verify that prior review feedback was properly addressed. The `deep-review` label path is dedupe-capped at one review per PR, so it will not re-fire — hence asking you directly.

Context: your earlier deep review flagged (🟡 Minor) that `actions/checkout` was pinned to v7.0.0 (`9c091bb2`) rather than the pin `main` uses. The author initially **disputed** that with a grep claiming all 16 pins in the repo were `9c091bb2`, then retracted — the grep had been run against a stale feature branch instead of `main`. The pin is now claimed to be `3d3c42e5`.

Please verify independently and treat every claim in the PR body and comments as unverified:

1. Resolve **every** action pin in `.github/workflows/scorecard.yml` through the GitHub API. Confirm each is a real commit. The author separately fabricated a nonexistent `download-artifact` SHA on sibling PR #411, so do not assume any SHA is valid.
2. Confirm each pin matches what `main` currently uses for that action.
3. For actions new to this repo (`ossf/scorecard-action`, `github/codeql-action/upload-sarif`), confirm the SHA matches the version in the trailing comment.
4. Re-check the two prior advisories: top-level `permissions: read-all`, and job-level permissions replacing rather than merging.
5. Confirm the rebase onto `main` did not break triggers, cron validity, or the `publish_results` + `id-token: write` pairing.

Report anything still wrong. If something is fine, say so briefly rather than inventing findings.

Issue #402, opened 2026-07-29
[Docs] Offer npm/pnpm/yarn/bun for every command — Tabs in the docs, small PM links on the homepage


allxsmith:
## Problem

Our docs hard-code a single package manager per snippet, and it isn't even consistently the same one. Readers on npm, yarn, or bun have to translate every command by hand, and in at least one place we tell them the wrong thing outright:

`docs/docs/guides/getting-started/installation.md` — the heading says npm, the command is pnpm:

> `### Using npm (recommended)` → followed by a `bash` block containing `pnpm add @allxsmith/bestax-bulma`

(and the "Using pnpm" section below it is a byte-for-byte duplicate of that same block.)

Most snippets elsewhere are pnpm-only (`pnpm add`, `pnpm create bestax@latest`, `pnpm dlx create-next-app`, `pnpm create vite`), while the skills pages use `npx skills add`. pnpm stays our recommended default — the ask here is that npm, yarn, and bun readers stop having to translate by hand.

## Proposal — docs: `<Tabs>` for every install/run command

Use Docusaurus `<Tabs>`/`<TabItem>` with **pnpm / npm / yarn / bun** for each command snippet, with a shared `groupId` (e.g. `groupId="package-manager"`) so the reader's choice persists across tabs and across pages. `<Tabs>` is already used in the repo — see `docs/docs/skills/theming.mdx` and `docs/docs/api/components/tabs.md` for the import pattern.

Tab order: **pnpm, npm, yarn, bun**, with `defaultValue="pnpm"`. pnpm first is load-bearing beyond visual default — see the LLM section below, where first-tab-wins decides what agents copy.

Command equivalents to standardize on:

| pnpm | npm | yarn | bun |
| --- | --- | --- | --- |
| `pnpm add @allxsmith/bestax-bulma` | `npm install @allxsmith/bestax-bulma` | `yarn add @allxsmith/bestax-bulma` | `bun add @allxsmith/bestax-bulma` |
| `pnpm add -D typescript` | `npm install -D typescript` | `yarn add -D typescript` | `bun add -d typescript` |
| `pnpm create bestax@latest` | `npm create bestax@latest` | `yarn create bestax` | `bun create bestax@latest` |
| `pnpm create vite@latest my-app --template react` | `npm create vite@latest my-app -- --template react` | `yarn create vite my-app --template react` | `bun create vite my-app --template react` |
[...]

PR #408, opened 2026-07-29
chore(docs): flatten MDX tab JSX out of the generated LLM artifacts

allxsmith:
Unblocks #402. This is the LLM-pipeline half only — no docs pages change yet, and `PackageManagerTabs` doesn't exist yet. Landing the flattener first means the tabs migration can go page-by-page without regressing the machine-readable docs on the way.

## Why a build step and not a plugin

`docusaurus-plugin-llms` reads the **source** markdown, not rendered HTML. Its only cleanup (`cleanMarkdownContent`, `utils.ts:587`) strips a fixed allowlist of plain HTML tags that doesn't include `Tabs`/`TabItem`, and `PluginOptions` (`types.ts:69`) exposes no transform hook. So tab JSX lands verbatim in `llms.txt`, `llms-full.txt` and every per-page `.md` twin — **112 tag occurrences today**, from the skills pages alone.

It can't be a sibling Docusaurus plugin either: core runs every plugin's `postBuild` under `Promise.all` (`@docusaurus/core/lib/commands/build/buildLocale.js:106`), so it would race the llms plugin nondeterministically. Hence `docusaurus build && node scripts/flatten-llms-tabs.mjs`, which is also the path `deploy.yml` / `test-deploy.yml` already take via `turbo run build --filter=@allxsmith/bestax-docs`.

## Two transforms, because there are two kinds of tabs

| kind | example | rule |
| --- | --- | --- |
| equivalent alternatives | `<PackageManagerTabs command="add …" />` | collapse to one fenced **pnpm** block |
| complementary content | skills pages' `ProfileCard.tsx` / `_profilecard.scss` / `Usage` | linearize to `####` headings, keep **every** body |

pnpm for the first because it's the default tab — what an agent copies out of `llms.txt` has to match what a reader sees. `command` is authored in pnpm's verb vocabulary so this stays a passthrough; npm/yarn/bun translation will live in the React component rather than being duplicated here.

The second rule matters: my first draft of this kept only the first `TabItem`, which would have **silently deleted** the skills pages' other two tabs.

## The trap worth reviewing

**bestax-bulma has its own `Tabs` component.** `<Tabs\b` matches `<Tabs.List` — `\b` sits between the `s` and the `.`. My first pass left 94 stragglers and would have corrupted `docs/api/components/tabs.md`.

Two mitigations, both required and both tested:

- match `<Tabs(?=[\s>])`, never `<Tabs\b`
- **mask** code before transforming — fenced blocks *and* inline spans, since the migration guides write `` `<Tabs>` `` in prose. Mask rather than split, because a `<Tabs>` block *wraps* fences and the document has to stay contiguous for `<Tabs>…</Tabs>` to match across it.

## Verification

`node --test` — 13 tests, including the `<Tabs.List>`-in-fence regression, inline-span survival, idempotency, and placeholder-collision safety. Wired as the docs package's `test` script, so `pnpm test` / `pnpm all` pick it up through turbo with no CI changes.

End-to-end against the current `build/`:

```
theme Tabs/TabItem left outside code: 0     (was 8 before inline-span masking)
library <Tabs.List> preserved       : 19
docs/api/components/tabs.md         : byte-identical
second run                          : 0 files rewritten (idempotent)
```

## Known limitation

A braced-template prop — ``command={`add foo`}`` — is deliberately unsupported: its backticks are indistinguishable from an inline code span, and masking has to run first. It's left untouched in the artifact rather than mangled, and there's a test pinning that behavior. Authors use a plain string.

## Not in scope

Converting the ~18 docs pages, building `PackageManagerTabs`, and the homepage hero links — all still #402.

---

Unrelated, noticed while running the gate: `pnpm format:check` fails locally on 133 files under `.claude/worktrees/` (the skill-loop eval worktree). That path is git-ignored so CI is unaffected, but Prettier only honors `.prettierignore`, so local `pnpm all` is noisy. Happy to add a `.prettierignore` entry in a separate PR if you want it.

## Summary by CodeRabbit

* **New Features**
  * Documentation builds now produce cleaner LLM-friendly output by converting package manager and content tabs into readable commands and headings.
  * Generated documentation preserves tab content while removing rendering-specific markup.

* **Documentation**
  * Added guidance describing the updated documentation pipeline and tab rendering behavior.

* **Tests**
  * Added coverage for tab conversion, code protection, import cleanup, supported commands, and repeated processing.


allxsmith:
@coderabbitai full review

allxsmith:
All three addressed in 867b0e2. Two reproduced as real bugs, one didn't — details below, since the distinction matters for what changed.

### 1. Mask all valid Markdown code delimiters — fixed

Replaced the fence regex with a line-based CommonMark scanner: either fence character, length ≥ 3, closing fence the same character and at least as long, up to 3 leading spaces of indentation, and a backtick fence's info string may not contain backticks. An unterminated fence now runs to end of document and is masked rather than rewritten — when we can't confidently parse, we don't mutate.

**Tilde fences reproduced.** Against the pre-fix code:

```
UNCHANGED (ok)   long fence wrapping short
MUTATED  (bug)   tilde fence
MUTATED  (bug)   import inside fence
MUTATED  (bug)   blank lines in fence
```

**Longer fences did not reproduce.** `docs/api/helpers/theme.md` and `docs/tutorial-basics/markdown-features.mdx` do wrap ``` examples in ```` fences, so I built the test from that real shape — but old and new agree on it exactly (fenced example intact, real tab flattened, no `#### Nope` leak). Fixed anyway, since the scanner removes the class rather than the instance, but I'd rather be straight that I couldn't produce a failing case for that half.

**Multiline inline spans are still single-line on purpose.** CommonMark permits them, but matching across newlines means one unpaired backtick in prose swallows the rest of the document. The failure mode of guessing wrong is far worse than the miss, so this is documented in the code rather than implemented.

### 2. Keep cleanup inside `outsideCode` — fixed, as suggested

Confirmed both halves against the pre-fix code: a fenced example demonstrating `import Tabs from '@theme/Tabs';` lost that line, and `\n{3,}` collapsed deliberate blank lines inside code blocks. Took the suggested diff. Regression tests for each.

### 3. Don't skip import-only artifacts — fixed

Extracted the predicate as an exported `NEEDS_FLATTENING` used by both `main()` and the tests, so the prefilter and `transform` can't drift apart again.

Worth noting for context: `docusaurus-plugin-llms` runs with `excludeImports: true`, and the deployed `llms-full.txt` contains **0** import lines — the plugin strips them upstream before this script ever runs. So the import handling here is defence in depth rather than load-bearing today. Still worth the consistency.

### Verification

21 tests, up from 13. Output over the current `build/` is **byte-identical to the pre-fix run**:

```
theme Tabs/TabItem outside code: 0
library <Tabs.List> preserved  : 19
flattened #### headings        : 165
docs/api/components/tabs.md    : byte-identical
second run                     : 0 files rewritten
```

So this closes latent edge cases without moving real-world output — which is the outcome I wanted from a review fix.

PR #411, opened 2026-07-29
ci: publish SBOMs per release and verify published provenance

allxsmith:
Adds `.github/workflows/supply-chain.yml` — two evidence artifacts our posture previously only *claimed*.

## What

**`sbom` job** — Syft generates SPDX **and** CycloneDX SBOMs (SPDX is the ISO standard procurement asks for; CycloneDX is what most scanners ingest natively). Uploaded as workflow artifacts, and on a `release: published` event attached to the GitHub release as durable, citable assets. Workflow artifacts expire; release assets are what an auditor can link to.

**`verify-provenance` job** — installs the real published tarballs from npm into a scratch tree and runs `npm audit signatures`. `SECURITY.md` tells consumers to run this; until now we never ran it ourselves.

## Why it's decoupled from the release

Triggers are `release: published`, weekly schedule, and `workflow_dispatch` — **never** `ci.yml`'s publish job. Nothing in this workflow can fail a release.

## Verified locally, not just written

```
$ npm install --ignore-scripts @allxsmith/bestax-bulma create-bestax
$ npm audit signatures
audited 17 packages in 2s
17 packages have verified registry signatures
4 packages have verified attestations
```

- [x] `prettier --check` passes
- [x] YAML parses; both jobs' permissions confirmed (`contents: write` on `sbom` only, for the release upload; `verify-provenance` inherits read)
- [x] All three actions SHA-pinned with version comments (`anchore/sbom-action` v0.24.0 dereferenced from its annotated tag)
- [ ] SBOM attachment can only be exercised by a real `release: published` event — verify on the next release

## ⚠️ Found while testing this: `bestax-migrate` is broken on npm

`verify-provenance` originally installed all three published packages. It failed — and not because of this workflow:

```
$ npm install bestax-migrate
npm error code EUNSUPPORTEDPROTOCOL
npm error Unsupported URL Type "workspace:": workspace:^
```

`bestax-migrate@1.0.0`'s published manifest contains `"@allxsmith/bestax-bulma": "workspace:^"` — an unresolved pnpm workspace specifier. **`npm install bestax-migrate` fails for every consumer, on every package manager.** Filed separately; this PR excludes that package from the audit (with a comment) so the check measures signatures rather than that bug. Add it back once fixed.

## Summary by CodeRabbit

* **New Features**
  * Generates SPDX and CycloneDX SBOMs for published releases and attaches them to the corresponding release assets.
* **Improvements**
  * Strengthened provenance verification by improving how provenance is detected and adding retries with backoff before failing.
  * Ensures the verification run fails if any targeted package lacks required provenance.

### DIFF
--- a/.github/workflows/supply-chain.yml
+++ b/.github/workflows/supply-chain.yml
@@ -19,13 +19,14 @@ permissions:
   contents: read
 
 jobs:
+  # Generation is read-only. It runs `pnpm install` and third-party actions, so
+  # it must never hold a write-scoped token — a compromised dependency or
+  # action would inherit it. The release upload is split into its own job below.
   sbom:
     name: SBOM
     runs-on: ubuntu-latest
     permissions:
-      # Only to attach the SBOM files to an existing release; nothing here
-      # creates or edits code.
-      contents: write
+      contents: read
 
     steps:
       - name: Checkout code
@@ -74,10 +75,26 @@ jobs:
           output-file: bestax-sbom.cdx.json
           upload-release-assets: false
 
-      # Workflow artifacts expire; a release asset is the durable, linkable
-      # copy an auditor can actually cite.
-      - name: Attach SBOMs to the release
-        if: github.event_name == 'release'
+  # Workflow artifacts expire; a release asset is the durable, linkable copy an
+  # auditor can actually cite. Isolated from generation so the write-scoped
+  # token exists only on release runs, in a job that installs nothing and runs
+  # no third-party code beyond the artifact download.
+  attach-sbom:
+    name: Attach SBOMs to the release
+    runs-on: ubuntu-latest
+    needs: sbom
+    if: github.event_name == 'release'
+    permissions:
+      contents: write
+
+    steps:
+      - name: Download SBOM artifacts
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
+        with:
+          pattern: bestax-sbom.*
+          merge-multiple: true
+
+      - name: Upload to the release
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           TAG: ${{ github.event.release.tag_name }}
@@ -127,7 +144,13 @@ jobs:
         run: |
           fail=0
           for pkg in @allxsmith/bestax-bulma create-bestax; do
-            predicate=$(npm view "$pkg" dist.attestations.provenance.predicateType 2>/dev/null || true)
+            # --json + a type guard so only a real string counts as present.
+            # Bare `npm view` of a missing field is empty today, but null /
+            # undefined / a non-string would otherwise read as success. Any
+            # failure (network, 404, missing dist-tag) yields empty and fails
+            # closed, which is the right default for a security assertion.
+            predicate=$(npm view "$pkg" dist.attestations.provenance.predicateType --json 2>/dev/null \
+              | jq -r 'select(type=="string")' 2>/dev/null || true)
             if [ -z "$predicate" ]; then
               echo "::error::$pkg has NO provenance attestation on the registry"
               fail=1
--- a/.github/workflows/supply-chain.yml
+++ b/.github/workflows/supply-chain.yml
@@ -58,6 +58,10 @@ jobs:
           format: spdx-json
           artifact-name: bestax-sbom.spdx.json
           output-file: bestax-sbom.spdx.json
+          # Defaults to true: the action would upload its own release asset on
+          # a release event, racing the explicit upload below and naming files
+          # by its own convention. Keep one deterministic uploader.
+          upload-release-assets: false
 
       # Both formats on purpose: SPDX is the ISO standard procurement asks for,
       # CycloneDX is what most scanning tools ingest natively.
@@ -68,6 +72,7 @@ jobs:
           format: cyclonedx-json
           artifact-name: bestax-sbom.cdx.json
           output-file: bestax-sbom.cdx.json
+          upload-release-assets: false
 
       # Workflow artifacts expire; a release asset is the durable, linkable
       # copy an auditor can actually cite.
--- a/.github/workflows/supply-chain.yml
+++ b/.github/workflows/supply-chain.yml
@@ -144,13 +144,23 @@ jobs:
         run: |
           fail=0
           for pkg in @allxsmith/bestax-bulma create-bestax; do
-            # --json + a type guard so only a real string counts as present.
-            # Bare `npm view` of a missing field is empty today, but null /
-            # undefined / a non-string would otherwise read as success. Any
-            # failure (network, 404, missing dist-tag) yields empty and fails
-            # closed, which is the right default for a security assertion.
-            predicate=$(npm view "$pkg" dist.attestations.provenance.predicateType --json 2>/dev/null \
-              | jq -r 'select(type=="string")' 2>/dev/null || true)
+            # --json + a type guard so only a real string counts as present:
+            # null, undefined, [] and non-strings all collapse to empty. The
+            # array branch covers npm aggregating a bare spec into one element.
+            # Any failure (network, 404, missing dist-tag) also yields empty and
+            # fails closed — the right default for a security assertion.
+            #
+            # Retried because this runs weekly and failing closed means a
+            # registry blip would otherwise report the same red as a genuinely
+            # dropped attestation. A gate that cries wolf gets ignored, so
+            # spend three attempts before believing provenance is really gone.
+            predicate=""
+            for attempt in 1 2 3; do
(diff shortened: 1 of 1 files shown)


allxsmith:
The temporary `bestax-migrate` exclusion in `verify-provenance` is now tracked as #416, so it doesn't get forgotten once #412 lands. The in-workflow comment points at the same follow-up.

allxsmith:
Both advisories addressed in 3f123d5 — #1 was a real weakness in the job's core promise.

### Advisory 1 (provenance absence vs. invalidity) — **fixed**

Correct and the sharpest finding here: a step named "Verify published provenance" that stays green when provenance *disappears* is the same failure class as #405 — claiming a gate that isn't enforced. Added an explicit presence assertion:

```bash
predicate=$(npm view "$pkg" dist.attestations.provenance.predicateType)
[ -z "$predicate" ] && fail=1
```

Verified against both controls before committing:

```
ok    @allxsmith/bestax-bulma — https://slsa.dev/provenance/v1
ok    create-bestax           — https://slsa.dev/provenance/v1
FAIL  fs-extra                — no provenance attestation on the registry
```

`fs-extra` is published without provenance, so it's a genuine negative control — the assertion detects absence rather than just passing everything.

### Advisory 2 (SBOM scope) — **documented, not changed**

Also correct: `path: .` over a full dev `pnpm install` enumerates storybook, playwright, docusaurus, jest, turbo — not what a consumer installs (which for bestax-bulma is Bulma alone).

I've scoped the claim rather than the artifact. A repository SBOM is legitimately useful — it's the audit trail for *what builds our releases* — the error would be letting it be read as the consumer closure. Added an explicit `NOTE ON SCOPE` comment saying so. A true per-package SBOM means running Syft over the packed tarballs, which is a bigger change and belongs in its own PR.

### On the residual-risk notes

- **`latest` vs. just-released version** — agreed and worth keeping on the record. semantic-release publishes to npm before creating the GitHub release, so `latest` is normally already current; a propagation lag would audit the prior version, which is harmless for a signature audit and re-runnable via `workflow_dispatch`.
- **Transitive `workspace:^` reintroduction refuted** — matches what I found: `create-bestax` depends only on chalk/commander/figures/fs-extra/prompts, so #412's broken package can't re-enter through it.

allxsmith:
@coderabbitai full review

allxsmith:
Valid finding — fixed in f498320. Verified the claim against `action.yml` at the exact pinned SHA rather than the tag:

```
$ gh api "repos/anchore/sbom-action/contents/action.yml?ref=e22c3899..." | base64 -d | grep -A4 upload-release-assets
  upload-release-assets:
    required: false
    description: "Upload release assets"
    default: "true"
```

Confirmed. On a `release: published` event that meant **three** uploads: one implicit per format step plus the explicit `gh release upload`, with two different naming conventions competing for the same release. Set `upload-release-assets: false` on both steps and kept the explicit uploader, which gives deterministic filenames and `--clobber` idempotency on re-runs.

Worth noting this was latent — nothing on a PR branch can fire a `release` event, so it would only have surfaced on the first real release after merge, as duplicate or oddly-named assets.

allxsmith:
@claude deep review again

allxsmith:
Both Copilot findings were valid and are fixed in 82c062b.

### `contents: write` on non-release runs — fixed as suggested

Correct, and the stronger of the two. The `sbom` job ran `pnpm install` plus third-party actions while holding a write-scoped token on *every* trigger, including the weekly schedule — so any compromised dependency or action inherited write access for no benefit.

Split exactly as recommended:

| job | permissions | when |
| --- | --- | --- |
| `sbom` | `contents: read` | every trigger |
| `attach-sbom` | `contents: write` | `needs: sbom`, `if: github.event_name == 'release'` |
| `verify-provenance` | inherits read | every trigger |

The upload job installs nothing — it only downloads the artifacts and calls `gh release upload`.

### `npm view` non-empty ≠ present — fixed, with a caveat on the mechanism

Adopted. Now `--json` with a type guard:

```bash
predicate=$(npm view "$pkg" dist.attestations.provenance.predicateType --json 2>/dev/null \
  | jq -r 'select(type=="string")' 2>/dev/null || true)
```

Being straight about what I could and couldn't confirm: I **could not reproduce** the specific `null`/`undefined` output. Bare `npm view` printed empty (length 0) for every package I tried without attestations — `fs-extra`, `lodash`, `chalk` — so the original `-z` test wasn't actually producing a false positive.

That said the recommendation is right regardless: the type guard makes presence explicit rather than inferred, and it costs nothing. Verified across all three cases:

```
@allxsmith/bestax-bulma  -> [https://slsa.dev/provenance/v1]  ok
create-bestax            -> [https://slsa.dev/provenance/v1]  ok
fs-extra                 -> []                                FAIL
lodash                   -> []                                FAIL
nonexistent-pkg-xyz-9987 -> []                                FAIL
```

The 404 case matters most: any failure — network, missing package, missing dist-tag — yields empty and fails closed, which is the correct default for a security assertion.

### Also caught in the process

Writing the `attach-sbom` job I initially used a `download-artifact` SHA from memory. It was **fabricated** — `gh api` returned *"No commit found for SHA."* The real pin on `main` is `3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8`. I've now verified every one of the five action pins in this file resolves to a real commit via the API, rather than trusting recall.

allxsmith:
deep-review: RE-REVIEW — verify prior review feedback was properly addressed. Head is 82c062b; no reviewer has examined this commit. Three reviewers raised five findings and the author claims all are fixed. Verify each INDEPENDENTLY against the code, do not trust the author’s reply comments:

1. Deep review advisory: `npm audit signatures` passes when provenance is ABSENT. Author added an "Assert our packages actually carry provenance" step. Does it actually fail closed on a missing attestation, on `npm view` failure, and on a 404?
2. Deep review advisory: SBOM covers the dev monorepo, not the published closure. Author documented the scope rather than changing it. Is the comment accurate and is the artifact honestly labelled?
3. CodeRabbit (Major): `anchore/sbom-action` defaults `upload-release-assets: true`. Author set it false on both steps. Confirm against action.yml at the PINNED SHA, and confirm no duplicate-upload path remains.
4. Copilot: `contents: write` held on non-release runs. Author split `sbom` (read) from `attach-sbom` (write, release-gated). Verify the write token is genuinely unreachable on schedule/workflow_dispatch, and that `attach-sbom` can still get the artifacts.
5. Copilot: `npm view` non-empty treated as present. Author switched to `--json` + a jq string type guard.

CRITICAL — the author made TWO action-pin errors on these PRs: first citing a pin sourced from a stale branch, then writing a `download-artifact` SHA that did not exist at all. Independently resolve EVERY action pin in supply-chain.yml via the GitHub API, confirm each is a real commit, and confirm each matches what main uses for that action. Treat any pin claim in the PR body or comments as unverified.

Also check: does `attach-sbom` correctly obtain both SBOM files given the artifact names used by the generate steps? A silent mismatch there would make the release upload fail only on a real release.

allxsmith:
Advisory 1 is fixed in 32f2f89. Advisory 2 I'm deliberately not fixing here — filed as #424.

### Advisory 1 (fails closed on transient registry error) — **fixed**

Right, and worth acting on: a weekly gate that goes red on a network blip is one people learn to ignore, which defeats the point. The lookup now retries three times with backoff before concluding provenance is gone, so flake and a genuinely dropped attestation no longer produce the same red.

Also handled the residual-risk note about npm returning the predicate as a single-element array — the guard is now `if type=="array" then .[0] else . end | select(type=="string")`.

Verified by extracting the step's `run:` block straight out of the YAML and executing it verbatim under `bash -eo pipefail` (Actions' default shell):

```
ok: @allxsmith/bestax-bulma -> https://slsa.dev/provenance/v1
ok: create-bestax -> https://slsa.dev/provenance/v1
EXIT=0
```

Negative cases still fail closed — `fs-extra`, a nonexistent package, `null`, and `[]` all collapse to empty.

### Advisory 2 (SBOM is the dev closure, not the consumer closure) — **deferred to #424**

Fixable, and worth fixing: `verify-provenance` already builds precisely the right tree — a scratch dir with the *published* packages — so pointing Syft at that yields the true consumer closure. For `bestax-bulma` that's essentially Bulma alone, which is a selling point the current dev-toolchain SBOM actively obscures.

I'm not bolting it on here because it changes the artifact set and the job's purpose, and there are real design questions to settle first (per-package vs. combined; `--omit=dev`; keeping the write-scoped token out of `verify-provenance` so the split Copilot asked for doesn't regress). That deserves its own review pass rather than riding along on a PR four reviewers have already cleared.

### Note for anyone reading the review trail

The re-review reported "PR is already merged" for #406/#407 — correct, both merged at 00:01Z. That is also the real reason the re-triggered deep reviews on those two skipped: `claude-review.yml`'s job requires `pull_request.state == 'open'`. I'd previously blamed the one-review-per-PR dedupe marker; that was wrong. #411 is still open, which is why its re-review ran normally.

Issue #416, opened 2026-07-29
[CI] Re-add bestax-migrate to the published-provenance check once #412 is fixed


allxsmith:
## Summary

Once #412 is fixed, add `bestax-migrate` back to the `verify-provenance` job in `.github/workflows/supply-chain.yml` (added by #411). That two-line change converts a temporary workaround into a permanent regression guard.

**Blocked by:** #412 (the fix) and #411 (the workflow that this edits).

## Background

#411's `verify-provenance` job installs the published packages from npm and runs `npm audit signatures`. `bestax-migrate` is currently excluded, because `npm install bestax-migrate` fails outright with `EUNSUPPORTEDPROTOCOL` (#412) and would mask real signature failures.

The exclusion is marked in the workflow with a comment pointing here:

```yaml
# bestax-migrate is deliberately absent: its published manifest still
# carries an unresolved `workspace:^` specifier, so `npm install` of it
# fails outright and would mask real signature failures here. Add it back
# once that is fixed.
```

## The change

In `.github/workflows/supply-chain.yml`, restore the third package and drop the comment:

```diff
-          npm install --ignore-scripts \
-            @allxsmith/bestax-bulma create-bestax
+          npm install --ignore-scripts \
+            @allxsmith/bestax-bulma create-bestax bestax-migrate
```

## Why this matters beyond tidiness

This is the **only** automated check that installs the *published* artifact. Everything else in CI operates on the workspace, where `workspace:^` resolves correctly — which is precisely why #412 shipped undetected and stayed broken across releases.

With all three packages included, the job fails the moment any published package becomes uninstallable, from any cause: another `workspace:` leak, a missing `files` entry, a broken `exports` map, an unpublished transitive dependency.

Worth considering while making this change: the job currently proves installability only as a side effect of running `npm audit signatures`. An explicit assertion that each published package installs cleanly would state the intent directly, and would keep working even if the signature audit is later scoped differently.

## Verification

Run the same commands the job runs, against the real registry — not the workspace:

```bash
cd "$(mktemp -d)" && npm init -y >/dev/null
npm install --ignore-scripts @allxsmith/bestax-bulma create-bestax bestax-migrate
npm audit signatures
```

Expect a clean install and all packages reporting verified registry signatures. Before #412 is fixed this fails at the install step; that failure is the regression this guards against.

## Acceptance Criteria

- [ ] #412 fixed and a corrected `bestax-migrate` published to npm
[...]

PR #417, opened 2026-07-29
fix(bestax-migrate): resolve workspace: specifiers before publishing

### LINKED ISSUE #412: [Bug] bestax-migrate is uninstallable from npm — published manifest contains "workspace:^"
## Summary

`bestax-migrate@1.0.0` is **uninstallable from npm**. Its published manifest contains an unresolved pnpm workspace specifier.

## Reproduction

```bash
npm install bestax-migrate
```

```
npm error code EUNSUPPORTEDPROTOCOL
npm error Unsupported URL Type "workspace:": workspace:^
```

Confirmed against the live registry:

```bash
$ npm view bestax-migrate dependencies --json
{
  "@allxsmith/bestax-bulma": "workspace:^",   <-- never rewritten
  "@babel/parser": "^7.26.0",
  "chalk": "^5.6.2",
  "commander": "^15.0.0",
  "jscodeshift": "^17.3.0"
}
```

This affects **every** package manager — `workspace:` is only meaningful inside a workspace, so npm, yarn, and pnpm all reject it for an external consumer. The package has been broken since it was first published.

## Root cause

`pnpm publish` rewrites `workspace:^` to the resolved semver range at pack time. **`npm publish` does not.**

The release pipeline uses `@semantic-release/npm` (`bestax-migrate/release.config.js`), which shells out to `npm publish` — so the rewrite never happens. `@allxsmith/bestax-bulma` and `create-bestax` are unaffected because neither declares a workspace dependency; `bestax-migrate` is the only published package that does.

Note this is invisible to CI: the workspace install resolves `workspace:^` correctly, tests pass, and the tarball uploads fine. Nothing in the pipeline installs the *published* artifact.

## Suggested fixes

1. **Rewrite at pack time** — add a `prepack`/`postpack` pair that substitutes the real version, mirroring the existing pointer-file approach in `bulma-ui/scripts/pack-pointer-files.mjs`. Keeps `workspace:^` for local development.
2. **Publish with pnpm** — replace `@semantic-release/npm`'s publish step with `pnpm publish` via `@semantic-release/exec`, so the rewrite happens natively. Larger blast radius: it changes the release path for a package, and OIDC trusted publishing needs to keep working.
3. **Drop the workspace protocol** — declare a plain semver range (`^5.0.0`). Simplest, but loses the guaranteed local link unless `linkWorkspacePackages` is configured to match.

Option 1 is the most contained.

## Regression guard

PR #411 adds a `verify-provenance` job that installs the published packages from npm. Extending it to assert every published package is installable would have caught this, and would catch the next occurrence. `bestax-migrate` is currently excluded there with a comment pointing at this issue.

## Related

- #411 — found while testing that workflow
- #407 — adds README badges and a supported-versions row for `bestax-migrate`; worth landing this fix before advertising the package more prominently

allxsmith:
Fixes #412.

## The bug

`bestax-migrate@1.0.0` is uninstallable by anyone:

```
$ npm install bestax-migrate
npm error code EUNSUPPORTEDPROTOCOL
npm error Unsupported URL Type "workspace:": workspace:^
```

`pnpm publish` resolves `workspace:^` to a real semver range at pack time. `npm publish` does not — and the release pipeline runs `@semantic-release/npm`, which shells out to `npm publish`. `bulma-ui` and `create-bestax` are unaffected because neither declares a workspace dependency.

As #412 notes, nothing in CI installs the *published* artifact, which is why this shipped undetected.

## Two problems, two changes

**1. `@allxsmith/bestax-bulma` was never a runtime dependency.** Nothing under `bestax-migrate/src` or the built `dist/` imports it — every reference is a *string* the codemod writes into the user's code (`transform.ts:34`) or the range it writes into the user's manifest (`deps.ts:10`, `BESTAX_RANGE = '^5'`). Its only real use is `e2e/kitchen-sink.test.ts`, which typechecks migrated output against the workspace build. Keeping it in `dependencies` made every consumer of a *codemod CLI* install the whole React component library. It moves to `devDependencies`.

`turbo.json` already pins `bestax-migrate#test` → `@allxsmith/bestax-bulma#build` explicitly, so the e2e never relied on the `dependencies` graph for ordering.

**2. Belt and braces: `scripts/pack-manifest.mjs`.** Resolves any remaining `workspace:` specifier in any dependency section during `prepack`, and restores the repo manifest in `postpack`. Deliberately mirrors the proven `bulma-ui/scripts/pack-pointer-files.mjs` pattern — the restore matters because `@semantic-release/git` commits `package.json`. The version comes from the linked `node_modules/<name>/package.json`, so at release time it picks up the version `bulma-ui`'s semantic-release just wrote (it runs first in `ci.yml`).

## Regression guard

New `publishable-manifests` sub-check in `scripts/check-conformance.mjs` (CI already runs `pnpm check:conformance`). For every non-private workspace package:

- no `workspace:` in `dependencies` / `peerDependencies` / `optionalDependencies` — the sections npm resolves for consumers;
- a `workspace:` left in `devDependencies` requires the pack-time hooks, so it can never be published verbatim.

Both failure modes were confirmed to fail with an actionable message before being fixed.

This is the in-repo, fast guard. It complements rather than replaces #416 — the published-provenance check in #411 still gets `bestax-migrate` added back once a corrected `1.0.1` is on the registry.

## Verification

`pnpm all` green. Then the repro from the issue, against the actual packed tarball:

- packed `package.json` has `"@allxsmith/bestax-bulma": "^5.8.0"` in `devDependencies`, no `workspace:` anywhere, and no `bestax-bulma` in `dependencies`;
- `npm install --ignore-scripts <tarball>` into a scratch project → 98 packages, 0 vulnerabilities;
- `bestax-migrate --help` runs;
- `node_modules/@allxsmith` is absent — the component library is no longer dragged in;
- the repo manifest is restored after packing and the backup file removed.

## Also here

- Root `CLAUDE.md` never listed `bestax-migrate/` in the monorepo package list. One-line fix, since that file is what CodeRabbit and the `@claude` action read as project instructions.
- `.gitignore` entry for `package.json.pack-backup`, next to the existing `bulma-ui/CLAUDE.md.bak`.
- `pnpm-lock.yaml` regenerated: the importer block moves sections, plus ~30 lines of incidental in-range drift (`@types/node` 26.0.1→26.1.1, `picomatch` 4.0.4→4.0.5) from re-resolving the subtree. `pnpm install --frozen-lockfile` passes.

## Release impact

`fix(bestax-migrate)` → patch release of `bestax-migrate` only, per its `releaseRules`. That publishes `1.0.1` with a correct, installable manifest.

## Summary by CodeRabbit

- **New Features**
  - Improved package publishing by resolving workspace dependency references before publication and restoring manifests afterward.
  - Ensured required migration-tool dependencies are included for package consumers.
  - Added validation for catalog references in publishable packages.

- **Bug Fixes**
  - Prevented published packages from exposing unresolved workspace or catalog references.

- **Chores**
  - Added automated pre-release checks for package manifest conformance.
  - Added safeguards for temporary packaging artifacts and clarified release documentation.


allxsmith:
@coderabbitai review

allxsmith:
@coderabbitai full review

allxsmith:
## Review follow-up

All five findings addressed across `7fda9db` and `5315efe`; branch also merged up to `main` (`444bff0`). `pnpm all` green — 21/21.

### Fixed

**CodeRabbit — bare `workspace:` specifier** (`pack-manifest.mjs`). Real bug, good catch. A bare `workspace:` is pnpm's shorthand for `workspace:*`, but it fell through to the "already carries its own range" branch and got unwrapped to `""`, packing an empty specifier. Reachable in practice — `pnpm add --save-workspace-protocol=rolling` writes exactly that form. Now resolved like `workspace:*`:

```
devDependencies.@allxsmith/bestax-bulma: workspace: -> 5.8.0
```

**CodeRabbit — the conformance claim was too broad** (`bestax-migrate/CLAUDE.md`). Correct: `checkPublishableManifests()` never checks *which section* `@allxsmith/bestax-bulma` sits in, so "enforces both" overclaimed. Narrowed to what the check actually does, and it now says plainly that re-adding the library as a plain-semver runtime dependency would pass CI and is caught only in review.

**Deep review #1 — `catalog:`** (`check-conformance.mjs`, `pack-manifest.mjs`). Closed rather than left as a future gap, since it's the same asymmetry that caused #412: `pnpm publish` resolves `catalog:`, `npm publish` ships it verbatim. The conformance guard now flags both protocols, and `pack-manifest.mjs` exits 1 rather than packing a `catalog:` it can't resolve — the range lives in `pnpm-workspace.yaml`, not in the linked package, so resolving it properly is real work with no current consumer. Failing loudly at pack time beats shipping another broken tarball. Both paths verified by temporarily injecting `"chalk": "catalog:"`.

**Deep review #2 — hard-coded `pack-manifest.mjs` name.** The premise was right and I'd initially only added a comment; there was a genuine hole underneath it. The check matched the *string* `pack-manifest.mjs` in the hooks, so a hook left pointing at a moved or deleted script passed conformance and failed at `npm publish` instead — the most expensive moment to find out. It now resolves the path out of each hook command and confirms the file exists, with a message distinct from the missing-hook case:

```
bestax-migrate/package.json points its prepack/postpack hooks at
"scripts/moved/pack-manifest.mjs", but bestax-migrate/scripts/moved/pack-manifest.mjs
does not exist. The workspace:/catalog: specifier in devDependencies would go out
unresolved (#412), and the failure would surface during the release rather than in CI.
```

Both failure modes were confirmed to fire (hook pointing at a moved path; hooks deleted entirely), and the clean tree still passes.

The name match itself stays, deliberately, and the reasoning is now a comment in the check. Accepting any `prepack`/`postpack` pair would wave through `bulma-ui/scripts/pack-pointer-files.mjs`, which doesn't perform this rewrite at all. The two failure directions aren't symmetric: a false positive costs one line of config, a false negative ships another uninstallable tarball. On the guard whose entire job is preventing a silent bad publish, that trade only runs one way.

### Not changed, deliberately

**Deep review #3 — published manifest references `scripts/`, which `files: ["dist"]` excludes.** Agreed it's inert, and left alone. npm runs `prepack`/`postpack` on pack and publish, never on install from a tarball, so a consumer never follows the path. Both available fixes are worse than the problem: adding `scripts/` to `files` ships ~120 lines of dead code to every install, and stripping the hooks during `prepack` risks npm not running `postpack` and leaving the repo's `package.json` rewritten for `@semantic-release/git` to commit — precisely the failure the backup/restore exists to prevent. The reasoning is now a "Note for reviewers" block in the `pack-manifest.mjs` header so this doesn't get re-raised each pass.

### Also

Merged `origin/main` in. The branch was 13 commits behind and `pnpm all` was flaking on `e2e/kitchen-sink.test.ts` — `test` and `test:coverage` run concurrently in one turbo invocation and raced over a shared `.e2e-tmp`. That's what #419 (`2211ea5`) fixed with a per-process scratch dir; this branch predated it. Clean merge, no conflicts.

### Re-verified end to end after the merge

Packed the real tarball again: `"@allxsmith/bestax-bulma": "^5.8.0"` in `devDependencies`, runtime deps still just `@babel/parser`, `chalk`, `commander`, `jscodeshift`, no `workspace:` or `catalog:` anywhere in the manifest, and the repo manifest restored clean afterward.

allxsmith:
@coderabbitai rate limit

allxsmith:
deep-review: Second pass requested. Since your first review the PR gained a `catalog:` guard, a pack-script existence check, a pnpm alias-form guard, and a merge up to `main`. Focus on `scripts/check-conformance.mjs` and `bestax-migrate/scripts/pack-manifest.mjs` — specifically the coupling between them, which is where both post-review defects have appeared.

**The invariant to attack.** These two files encode the same rule twice and must not drift: `pack-manifest.mjs` decides which specifier shapes it will resolve, and `checkPublishableManifests()` decides which shapes it will let past on the strength of the pack hooks being wired. Any shape the script *refuses* but the check *excuses* is a silent CI-green / release-red inversion — the exact failure the check exists to prevent. That inversion has already happened twice here:

1. CodeRabbit caught it for `catalog:` (fixed in `4127ead`) — verified live at head before fixing, conformance exit 0 vs pack exit 1.
2. I caught it for `workspace:<name>@<range>` (fixed in `de6a900`) — the alias form was unwrapped to a bare `@scope/pkg@^5`, which nothing can install, and exited 0.

Assume a third instance exists. Enumerate every specifier form `pnpm publish` rewrites and check each against both files. `workspace:*`, `workspace:^`, `workspace:~`, bare `workspace:`, `workspace:<range>`, `workspace:<name>@<range>`, and `catalog:`/`catalog:<name>` are the ones I know of — if pnpm has others, that is the gap.

**Specific things worth doubting:**

- The alias detection is `rest.includes('/') || rest.lastIndexOf('@') > 0` in both files. I claim no semver range contains `/` or a non-leading `@`. Find a counter-example that either false-positives a legitimate range or false-negatives an alias.
- `UNRESOLVABLE_AT_PACK` in the check and the two `process.exit(1)` branches in the script are hand-synced. Is there a shape one has and the other doesn't?
- `parseWorkspacePackages()` is a hand-rolled YAML line parser. It only reads the `packages:` block. Consider anchors, quoted forms, flow sequences (`packages: [a, b]`), comments mid-block.
- The check iterates `pnpm-workspace.yaml`'s `packages:` list and skips `private` packages. A publishable package outside that list is unscanned.
- `postpack` restoring the manifest is what keeps `@semantic-release/git` from committing a rewritten `package.json`. I argue the only gap is a hard process kill between hooks. Check whether `npm publish` failure paths (registry rejection, auth failure, a failing `prepublishOnly`) all still reach `postpack`.

**Two deliberate non-fixes** — argue against them if you disagree, but they are decisions, not oversights:

- The devDeps rule matches the hook script by the literal name `pack-manifest.mjs`. Name-agnostic matching would wave through `bulma-ui/scripts/pack-pointer-files.mjs`, which does not perform this rewrite. False positive costs one line of config; false negative ships an uninstallable tarball.
- The published manifest keeps `prepack`/`postpack` pointing at `scripts/`, which `files: ["dist"]` excludes. npm runs those hooks on pack/publish, never on install, so the path is never followed by a consumer. Reasoning is in the `pack-manifest.mjs` header.

**Not worth your tokens:** the `bestax-migrate` → `devDependencies` move and the turbo ordering argument (settled in your first pass), and the `pnpm-lock.yaml` drift.

Verification so far is `pnpm all` green (21/21) plus targeted negative tests — I mutated `bestax-migrate/package.json` into each failure shape and confirmed the right violation fires, including three at once. There are no automated tests for `checkPublishableManifests()` itself; if you think this logic has earned them, say so and where.

PR #418, opened 2026-07-30
chore(docs): make the LLM flattener indentation-aware before the #402 rollout

allxsmith:
First PR of #402. No docs pages change here — this fixes a latent defect in the flattener that #408 shipped, which would have corrupted the LLM artifacts the moment a `<PackageManagerTabs>` landed inside a numbered step.

## The bug

The flattener replaces JSX in place, but only the *opening* fence inherited the tag's indentation. Running the merged `transform()` on a nested component:

```
1. **Create it:**

   <PackageManagerTabs command="add foo" />
```

produced:

```
1. **Create it:**

   ```bash
pnpm add foo
```
```

Opening fence at 3 spaces, body and closer at column 0. Per CommonMark the under-indented lines terminate the list item: the command becomes prose, and the following step gets swallowed into the block that the stray ``` opens. `transform()` reports the corrupted output as idempotent, so nothing self-heals, and re-running masks the damage as code.

**20 of #402's target sites are nested** — 18 in `react-setups.md` at 3 spaces, `migration/bulma-ui-3-to-4.md:40` at 2 — and because `llms-full.txt` is one concatenated document, `bulma-ui-3-to-4.md` (the only fence on its page) would have bled into the next page. Nothing in CI would have caught it: the HTML site builds perfectly and only the machine-readable artifacts are wrong.

## The fix

Both `flattenPackageManagerTabs` and `flattenGenericTabs` now capture the tag's indentation and carry it onto every emitted line. Two consequences worth reviewing:

- The pattern is **anchored to the start of a line**, since only a tag that starts its own line has an indentation to inherit. A component mid-sentence therefore no longer matches and survives as visible JSX — a loud failure instead of an inline fence that corrupts the document. The build gate below catches it.
- `flattenGenericTabs` **dedents then re-indents** the tab body rather than prefixing it, so a body that already carries the source indentation doesn't end up doubly indented — at 4+ spaces that would silently become a markdown code block.

## Shared translation module

Adds `docs/src/components/PackageManagerTabs/translate.mjs` — the pnpm → npm/yarn/bun vocabulary, which the component in the next PR will consume. The flattener now imports `renderCommand` from it instead of its own `toPnpm`, so the "duplicating them here would be a second source of truth" claim in its header comment is mechanically true rather than aspirational. A test asserts the flattened body equals `renderCommand(cmd, 'pnpm')` for every row of the table — the single assertion keeping llms.txt and the default tab from drifting.

`.mjs` because `docs/package.json` has no `"type": "module"`, so node would load a `.js` as CommonJS and `node --test` couldn't import it.

A `command` may now hold several `;`-separated lines. Segments whose first token is a known verb get translated; anything else (`cd my-app`, a `#` comment) passes through byte-identically. That's what the Quick Start scaffold sequences need.

## Build gate

`main()` now throws if a rewritten artifact still contains tab JSX outside code, has an unterminated fence, or has **a fenced block dedented below its opening fence**.

That third check is the one that catches the bug above, and it's worth explaining why the obvious check doesn't: I first wrote this as a fence-parity assertion and the test failed. CommonMark lets a closing fence sit at 0–3 spaces *regardless* of the opening indent, so the corrupted output pairs its fences happily and parity sees nothing wrong. The damage is structural, not lexical. Comparing body indentation against the opening fence is what actually detects it.

## Verification

- **The LLM artifacts are byte-identical to a pre-change baseline build.** `diff -r` over `*.md` twins and `llms*.txt` is empty, so this closes the latent defect without moving any published output.
- Coverage 21 → **64 tests**. The new ones were written failing first against the merged code: 11 failed before the fix.
- Includes a **corpus test** that runs every shipped page through `transform()` and asserts no surviving tab JSX, no dedented or unterminated fence, no empty emitted block, and idempotency — so a future nested usage fails in `pnpm test` rather than silently in `llms-full.txt`.
- Gates: `check:conformance` 7/7, `gen:catalog:check` clean, prettier clean, full docs build green with the new verifier active.

One incidental cleanup: removed a duplicated `outsideCode` doc comment that #408 left sitting above `maskFences`.

## Not in scope

The component, the global `MDXComponents` registration, the homepage, and any docs page conversion — those follow in the next PRs of #402.


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **New Features**
  * Documentation now renders package-manager commands consistently for pnpm, npm, Yarn, and Bun.
  * Multi-step commands are displayed clearly across separate lines.
  * Flattened documentation preserves indentation, list formatting, headings, and code blocks.
* **Bug Fixes**
  * Improved handling of tabbed documentation and nested code examples.
  * Build validation now detects unflattened tabs, malformed fences, and swallowed commands before release.
* **Tests**
  * Added comprehensive coverage for documentation transformations and command rendering.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/docs/scripts/flatten-llms-corpus.test.mjs
+++ b/docs/scripts/flatten-llms-corpus.test.mjs
@@ -1,0 +1,77 @@
+/**
+ * Run the flattener over the real docs corpus.
+ *
+ * The unit tests use hand-written fixtures, so they only ever cover shapes
+ * someone thought to write down. This walks every page we actually ship and
+ * asserts the invariants that matter for the published artifacts — which means a
+ * future page that nests a component somewhere awkward fails here rather than
+ * silently corrupting llms-full.txt.
+ *
+ * Source-only: no build required, so it runs in the same fast `pnpm test`.
+ */
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { readFile, readdir } from 'node:fs/promises';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { transform, verifyArtifact } from './flatten-llms-tabs.mjs';
+
+const DOCS_DIR = path.resolve(
+  path.dirname(fileURLToPath(import.meta.url)),
+  '../docs'
+);
+
+async function* walk(dir) {
+  for (const entry of await readdir(dir, { withFileTypes: true })) {
+    const full = path.join(dir, entry.name);
+    if (entry.isDirectory()) yield* walk(full);
+    else if (/\.mdx?$/.test(entry.name)) yield full;
+  }
+}
+
+const pages = [];
+for await (const file of walk(DOCS_DIR)) pages.push(file);
+
+test('the corpus is non-empty (guards a bad DOCS_DIR)', () => {
+  assert.ok(pages.length > 100, `expected >100 pages, found ${pages.length}`);
+});
+
+test('every page survives the flattener cleanly', async () => {
+  const failures = [];
+
+  for (const file of pages) {
+    const src = await readFile(file, 'utf8');
+    const out = transform(src);
+    const rel = path.relative(DOCS_DIR, file);
+
+    for (const problem of verifyArtifact(out)) {
+      failures.push(`${rel}: ${problem}`);
+    }
+
+    // An empty emitted block means a command was swallowed. Anchored to line
+    // starts and to the indentation we emit — a loose regex false-positives on
+    // the inner fence of a ```` block (docs/tutorial-basics/markdown-features).
+    if (/^([ \t]*)```bash\n\1```$/m.test(out)) {
+      failures.push(`${rel}: emitted an empty bash fence`);
+    }
+
+    // Idempotency on real content, not just fixtures.
+    if (transform(out) !== out) {
+      failures.push(`${rel}: transform is not idempotent`);
+    }
+  }
+
+  assert.deepEqual(failures, []);
+});
+
+test('pages with no tab JSX are passed through untouched', async () => {
+  const changed = [];
+
+  for (const file of pages) {
+    const src = await readFile(file, 'utf8');
+    if (/<Tabs|<TabItem|<PackageManagerTabs/.test(src)) continue;
+    if (transform(src) !== src) changed.push(path.relative(DOCS_DIR, file));
+  }
+
+  assert.deepEqual(changed, []);
+});
--- a/docs/scripts/flatten-llms-gate.test.mjs
+++ b/docs/scripts/flatten-llms-gate.test.mjs
@@ -81,6 +81,53 @@ test('the gate fails on a dedented fence body', async () => {
   assert.match(result.stderr, /dedented below its opening fence/);
 });
 
+test('the gate fails on a dedented fence body nested past 3 spaces', async () => {
+  // Two list levels deep, and with a following page whose fence pairs with the
+  // stray column-0 one — the shape that used to reach publication silently,
+  // because a fence opened at 4+ spaces was invisible to the check entirely.
+  const result = await runGate({
+    'llms-full.txt': [
+      '1. Outer step:',
+      '',
+      '   - Inner step:',
+      '',
+      '     ```bash',
+      'pnpm add foo',
+      '```',
+      '',
+      '# Next page',
+      '',
+      '```ts',
+      'const a = 1;',
+      '```',
+      '',
+      '<PackageManagerTabs command="install" />',
+    ].join('\n'),
+  });
+
+  assert.equal(result.ok, false, 'expected the build to fail');
+  assert.match(result.stderr, /dedented below its opening fence/);
+});
+
+test('the gate passes an indented code block that quotes a fence', async () => {
+  // The other side of that widening. At the top level this is indented code, not
+  // a nested fence, so neither its unclosed ``` nor its column-0 neighbours may
+  // be read as damage — otherwise closing the gap above would redden valid pages.
+  const result = await runGate({
+    'docs/guide.md': [
+      'An indented code block showing a fence:',
+      '',
+      '    ```bash',
+      '    pnpm add foo',
+      '',
+      '<PackageManagerTabs command="install" />',
+    ].join('\n'),
+  });
+
+  assert.ok(result.ok, `expected success, got: ${result.stderr}`);
+  assert.match(result.stdout, /rewrote 1 file/);
+});
+
 test('artifacts with no tab JSX are not even read for problems', async () => {
   // The NEEDS_FLATTENING prefilter still applies — an unrelated malformed page
   // must not start failing the docs build.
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by allxsmith** [comment]
Both addressed in 84dcbea.

### `main()` skipped verification on a no-op transform — fixed

Correct, and worse than it reads: this exempted precisely the shape the gate was built for. The line-anchored pattern deliberately skips a mid-line tag, which means `transform()` returns the input unchanged, which meant `after === before` short-circuited before `verifyArtifact` ever ran. The "loud and visible" failure I claimed in the PR description was silent.

Confirmed against the previous commit, same fixture both times:

```
OLD (a6aa4bf)  flatten-llms-tabs: rewrote 0 file(s)          exit 0
NEW (84dcbea)  llms-full.txt: unflattened tab JSX outside code: <PackageManagerTabs
               1 artifact(s) failed verification              exit 1
```

Verification now runs for every artifact matching `NEEDS_FLATTENING`; only the write stays conditional.

Worth noting why the existing tests missed it: `verifyArtifact` had unit coverage and was correct. The bug was in the wiring around it, which a pure-function test cannot see. So the new coverage is `flatten-llms-gate.test.mjs`, which runs the real script against a temp build dir and asserts the **process exit code** — plus a case pinning that the prefilter still spares unrelated malformed pages, so this doesn't turn every pre-existing unterminated fence into a build failure.

### "byte-identically" was inaccurate — fixed

Also correct. `splitSegments` trims and collapses whitespace, so `#   Remove    the  old  dep` renders as `# Remove the old dep`. I took the comment fix rather than the behaviour fix: the normalization is deliberate (it's what lets authors pad around the `;` separators), and the guarantee that actually matters is that a passthrough line is identical *across the four tabs*, not identical to the source. Reworded to say that explicitly, and pinned with a test asserting all four managers emit the same passthrough line.

### Verification

70 tests, up from 64. LLM artifacts still byte-identical to the pre-change baseline build, so neither fix moved published output.

**Comment by allxsmith** [comment]
deep-review: The prior review predates 84dcbea and never saw the gate fix. Be adversarial about the gate itself, not the flattener.

1. **`verifyArtifact`'s wiring.** 84dcbea made verification unconditional and added `flatten-llms-gate.test.mjs`, which shells out to the real script over a temp build dir and asserts exit codes. Does that test actually fail if the wiring regresses, or can it pass vacuously (wrong cwd, swallowed stderr, temp dir not seeded, exit code masked by the harness)? The original bug was invisible to pure-function tests — confirm the new test is not invisible in the same way.

2. **The two advisories you raised.** Both survive at head. `maskFences` only matches opens at 0–3 spaces so `open.indent` is never >= 4; state plainly whether #402's 20 nested sites can ever produce a fence the gate structurally cannot see, and whether the corpus test compensates or shares the blind spot.

3. **The false-positive direction.** The dedent predicate flags *any* less-indented non-blank body line. What shipped page shape would trip it and turn a valid doc into a red build?

4. **"Byte-identical to a pre-change baseline."** Try to refute it — is the `diff -r` scope actually complete over `llms*.txt` and the `.md` twins?

5. **`translate.mjs` semantics.** Multi-segment `;` commands plus whitespace normalization: find an input where a passthrough segment differs across tabs, or where a verb-leading segment is translated when it should not be.

**Comment by allxsmith** [comment]
Advisory #1 closed in `05766e6`. Advisory #2 is a won't-fix, reasoning below.

### #1 — the gate was blind past 3 spaces: closed

Confirmed as stated. `open.indent` could only ever be 1–3, so the dedent signal could not see a fence nested deeper than one list level. Worth being precise about *why* that mattered, because the review's own "usually trips the unterminated-fence check instead" is the part that doesn't hold in the artifact we actually publish:

```
1. Outer step:

   - Inner step:

     ```bash        <- 5 spaces: invisible to the check
pnpm add foo         <- corruption
```                  <- stray closer at column 0

# Next page

```ts                <- info string, so not a *closer* ...
const a = 1;
```                  <- ... this pairs with the stray one instead
```

Fences balance, the whole region masks as code, and `verifyArtifact` returned `[]`. Not a weaker signal — no signal. That fixture is the new regression test, and it is the one of the four new cases that fails against `84dcbea`.

The fix follows the reason the hole existed: CommonMark's 0–3 allowance is relative to the *enclosing container*, not the document, so seeing a nested fence means knowing the container's column. `maskFences` gains a `listAware` mode tracking the innermost list item's content column, and the closing fence is now measured against that column rather than against the opening fence's own indent — which is what lets a deep fence pair at all, while still pairing a closer that sits *below* it, since that is the corruption signature.

Two deliberate limits, both because a gate that reddens valid builds is worse than one with a blind spot:

- **`verifyArtifact` runs two passes.** Residual tab JSX keeps the transform's view of code. Widening it there would mask a region the flattener *does* rewrite and call a genuine miss clean — the opposite of the bug being fixed. Only the dedent signal comes from the `listAware` pass.
- **The unterminated-fence signal stays with the narrow pass.** Under the wide reading, an unclosed fence at 4+ spaces is usually just a top-level indented code block quoting a ```` ``` ```` line — valid markdown. My first cut trusted the wide pass with that signal too, and the test I wrote for it failed: a plain indented code block reddened the build. Trading a blind spot for a false positive is not a fix.

Incidental: indentation is now measured in columns with tabs expanded to the next multiple of four, so a tab-indented fence still reads as the indented code block it is. Without that, widening the pattern from `^ {0,3}` to `^[ \t]*` would have quietly started treating one as a fence at column 1.

### #2 — dedent predicate false-positives on flush-left content: won't fix

Correct as described, and I'm leaving it. The two failure modes are the same bytes: "an author wrote a flush-left line inside an indented fence" and "a rewrite dropped the tag's indentation" are indistinguishable to any predicate, because the second *is* the first — that's what makes the corruption silent in the first place. So there is nothing to narrow on except the thing being detected.

The asymmetry decides it. A false positive is a red build naming the file and the problem, fixed by indenting the block — which the source wanted anyway, since flush-left content inside an indented fence is fragile in CommonMark regardless of this script. A false negative is corrupted `llms-full.txt` shipped to agents, invisible in the rendered site, which is the entire class of defect this PR exists to close.

It also isn't the build that hits it first: the corpus test runs every shipped page through `transform()` + `verifyArtifact()` in `pnpm test`, so an author who writes that shape sees it locally before CI. Corpus and a full docs build are clean today.

Noting honestly that #1's widening extends #2's surface — the predicate can now also fire on a deep fence — and that this is the trade I'm taking knowingly, not an oversight.

### Verification

- **77 tests**, up from 74. Four new: the deep-nesting regression (fails against `84dcbea`), a correctly-indented 6-space fence, the indented-code-block non-false-positive, the tab-indent rule, and one pinning that the residual-JSX scan keeps the transform's view of code.
- **Transform behaviour unchanged, measured rather than asserted.** With `listAware` off the container column is 0, which is the old rule exactly — and `transform()` output is byte-identical across all 144 shipped pages *and* their concatenation, `verifyArtifact`'s verdict is unchanged on every page, and both script versions run over a real build's artifacts with 0 rewrites and exit 0. The published LLM artifacts still match the pre-change baseline build.
- `pnpm all` green; `check:conformance` 7/7; `gen:catalog:check` clean; prettier clean.

⚠️ The commit is on `claude/pr-418-resolution-ubeqlm`, not on this PR's head — it's `docs/pm-tabs-flattener-indent` plus one commit, so it fast-forwards. Say the word and I'll push it to the PR branch.

---

PR #399, opened 2026-07-28
docs(bulma-ui): generate API page Props, Import, Overview and CSS variables (layout pilot)

allxsmith:
# Pull Request

> **Please do not merge** — this is a pilot opened for review and preview comparison.

## Description

The API reference pages are hand-written, and the sections that merely restate the source drift
from it. Nothing in CI compared a props table to its interface, so the tables diverged in **shape**
(3-column in `elements/button.md`, 4-column `| Field |` in `layout/footer.md`, `| Prop |` in the
`components/avatar.md` exemplar) and in **content** (`button.md` renders `textColor` as
`` `'primary'` | ... (see docs) ``; `hero.md` and `card.md` instead dump 19-member colour unions).

This takes machine ownership of the four derivable regions and leaves everything else alone.
Regions are delimited by `<!-- bestax:generated <id> -->` markers:

| Region | Source |
| --- | --- |
| `overview` | the summary sentence, from the component's TSDoc |
| `import` | the public barrel |
| `props` | the `<X>Props` interfaces, via the TypeScript compiler API |
| `cssvars` | **new section**, parsed from the SCSS |

Prose outside the markers is preserved byte-for-byte, so pages keep their admonitions and their
hand-written Usage / Accessibility / Related / Additional sections. Section order is now
Overview → Import → Usage → Accessibility → Related → Additional → **Props → CSS & Sass Variables**.
Deleting a marker pair opts a region out.

**This is not typedoc.** Three choices keep the output reading as hand-written:

- **Own members only.** Resolving the heritage yields ~250 DOM attributes plus ~65 helper props, so
  the generator reads the interface's declared members and emits the one catch-all `...` row the
  pages already wrote by hand.
- **Types from AST source text, never `checker.typeToString`** — the checker expands
  `(typeof validColors)[number]` into 19 literals. Wide colour unions instead become a link to the
  existing [Valid values](https://bestax.io/docs/api/helpers/valid-values) page.
- **Prop descriptions were seeded from the existing docs tables**, not the `@property` blocks,
  because the hand-written ones are consistently richer. Nothing curated was lost.

`@property` blocks migrate onto interface members as inline TSDoc, so descriptions are now
verifiable by position and show up in users' editors. Component TSDoc summaries were seeded **from**
the pages for the same reason — the existing ones ("Bulma Hero component root.") were worse than
the prose they'd have replaced.

Scope: `layout/` only (6 pages). `MANAGED_CATEGORIES` in `scripts/lib/api-sources.mjs` gates the
[...]

Issue #350, opened 2026-07-22
[Feature] create-bestax: the "never inline style={{}}" rule needs a concrete style-to-helper-prop mapping


allxsmith:
## Affected Package(s)

- [ ] bulma-ui (`@allxsmith/bestax-bulma`)
- [x] create-bestax (`create-bestax`)
- [ ] docs (`@allxsmith/bestax-docs`)
- [ ] Other (please specify):

---

## Is your feature request related to a problem? Please describe.

**Background** (same A/B prompt experiment as the sibling issues): an AI agent built a full "Skynet" marketing site inside a create-bestax scaffold whose generated CLAUDE.md includes the House style rule:

> Never inline `style={{}}` — use the helper props every component accepts (`m*`/`p*` spacing, `textColor`/`bgColor`, `display="flex"`, `flexDirection`, `alignItems`).

The run violated it **47 times** across 8 components — and the model demonstrably *knew* the helper props, using `mb="4"`, `py="6"`, `textAlign="centered"`, `textColor="white"` correctly elsewhere in the same files. Most violations had direct equivalents:

| It wrote | Helper equivalent it used elsewhere |
|---|---|
| `style={{ marginTop: '2rem' }}` | `mt="5"` |
| `style={{ textAlign: 'center', marginBottom: '3rem' }}` | `textAlign="centered"` + `mb="6"` |
| `style={{ color: '#b9a9ab' }}` | `textColor="grey-light"` |
| `style={{ display: 'flex', flexDirection: 'column', height: '100%' }}` | `display="flex"` + `flexDirection="column"` |
| `style={{ fontWeight: 900 }}` | text-weight helper |

The rule states the prohibition and names prop *families*, but gives no lookup path from "the inline value I'm about to write" to "the prop I should write instead." Under generation pressure, the shortest path wins and the rule loses — especially mid-JSX where the model already has `style={{` half-typed in its head.

---

## Describe the solution you'd like

1. **Add a compact mapping table to the House style block** in `create-bestax/src/constants.ts` (`CLAUDE_MD` template), or a small reference file it links: common inline-style patterns on the left, the helper prop / Bulma helper on the right — spacing (`m*`/`p*`), `textColor`/`bgColor`, `textAlign`, text size/weight helpers, `display`/flex helpers, visibility/responsive helpers.
2. **Mirror the mapping into the skills that generate the most JSX** — `bestax-layout-scaffold` and `bestax-custom-component` — so it's in context exactly when pages are being scaffolded.
[...]

PR #420, opened 2026-07-30
docs: add the API page generator, its CI gates and review aids (1/4)

allxsmith:
# Pull Request

> **1 of 4** — replaces #399, which grew to 199 files and is now past CodeRabbit's 100-file review limit ("Review skipped: 192 files exceed the limit of 100"). Same final tree, split so each part is reviewable. Merge in order: this → #421 → #422 → #423.

## Description

The 87 API reference pages are hand-written, and the sections that merely restate the source drift from it. Nothing in CI compared a props table to its interface, so the tables diverged in **shape** (3-column in `elements/button.md`, 4-column `| Field |` in `layout/footer.md`, `| Prop |` in the `components/avatar.md` exemplar) and in **content** (`button.md` renders `textColor` as `` `'primary'` | … (see docs) ``; `hero.md` and `card.md` instead dump 19-member colour unions that blow the table 300 columns wide).

This lands the tooling **with no page generated yet**. `MANAGED_CATEGORIES` and `ORDERED_CATEGORIES` both start empty and grow one category per follow-up, so the gates are live from this commit and simply have nothing to police.

Four regions per page, delimited by `<!-- bestax:generated <id> -->` markers:

| Region | Source |
| --- | --- |
| `overview` | the component's TSDoc summary sentence |
| `import` | the public barrel |
| `props` | the `<X>Props` interfaces, via the TypeScript compiler API |
| `cssvars` | **new section**, parsed from the SCSS |

Everything outside a marker pair is hand-written and preserved byte-for-byte. Deleting a pair opts that region out, and `docs-section-order` makes that visible rather than silent.

### This is not typedoc

Past typedoc attempts produced the ugly, confusingly-structured output that led to these pages being hand-written. Three choices keep it reading as hand-written:

- **Own members only.** Resolving the heritage yields ~250 DOM attributes plus ~65 helper props. The generator reads the interface's declared members and emits the one catch-all `...` row the pages already wrote by hand.
- **Types from AST source text, never `checker.typeToString`** — the checker expands `(typeof validColors)[number]` into 19 literals, which is exactly the unreadable cell in `hero.md` today.
- **Wide colour unions become a link** to the existing [Valid values](https://bestax.io/docs/api/helpers/valid-values) page. That link is the single biggest readability lever.

### CI gates without touching `.github/**`

Two new sub-checks in `check-conformance.mjs`, which already runs in CI: `docs-generated` recomputes each managed page **in memory** and diffs; `docs-section-order` holds the order and marker presence. Both verified by tampering.

### Two gate gaps this exposed, also closed here

- `pnpm all` ran `turbo run format:check`, which only runs the **per-package** scripts — none cover `scripts/`, `docs/scripts/` or any `.md`. CI runs the **root** script, which does. So CI could be red while `pnpm all` reported green, and on #399 it was.
- **Nothing linted `scripts/` at all.** `pnpm lint` is `turbo run lint`, and only bulma-ui/create-bestax/bestax-migrate define one, each scoped to its own `src`.

`all` now calls the root format check, and `lint` also runs eslint over `scripts` and `docs/scripts` — so CI picks it up with no workflow change. Verified by planting an unused variable and confirming the gate fails. That gap had been hiding 7 eslint errors, fixed here.

### Review aids

Both one-shot, both used to validate the migrations that follow:

- `check-docs-parity.mjs` — fails on any prop, default, description word, code span, URL, live example or prose line a generated page drops relative to its hand-written self.
- `check-docs-wording.mjs` — buckets every previously-documented prop by how its description changed (`unchanged` / `extended` / `merged` / `replaced`).

## Related Issue(s)

Refs #399

## Type of Change

- [x] Build tooling
- [x] Documentation
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance

Affected packages:

- [ ] bulma-ui (`@allxsmith/bestax-bulma`) — no source changes in this part
- [x] docs (`@allxsmith/bestax-docs`) — build chains the marker-stripping step
- [ ] create-bestax

## Checklist

- [x] I have added/updated documentation as needed (`docs/CLAUDE.md`, `bulma-ui/CLAUDE.md`, `CONTRIBUTING-COMPONENTS.md`)
- [ ] I have added tests that prove my fix is effective — the generator's test is its idempotence gate in CI
- [ ] Any relevant dependencies are updated — none added

## Additional Context

**Markers and the LLM surface.** `docusaurus-plugin-llms` copies markdown from disk and does not strip HTML comments, so the markers would land ~600 times in `llms-full.txt` and in every per-page `.md` twin — the files agents ingest. `docs/scripts/strip-generated-markers.mjs` runs after `docusaurus build` and removes them from the built output only. It is a build step rather than a plugin because `postBuild` hooks run under `Promise.all`, so a plugin declared after the llms plugin still races it — the first attempt did exactly that and stripped nothing.

**New requirement:** generating reads `node_modules/bulma` for the stock components' SCSS, a deliberate departure from `gen-component-catalog.mjs`'s no-install rule. CI installs first, and the script fails with an actionable message rather than an `ENOENT`. A Bulma version bump will now produce a real diff here, which is the point.

**`registerVarsKeys`** moved to `scripts/lib/scss-vars.mjs` so the generator shares one SCSS parser; key extraction is unchanged, verified key-for-key across all 26 repo partials (366 keys, 0 diffs).

---


allxsmith:
@coderabbitai full review

---

allxsmith:
## Review of the stack — props tables verified page-by-page in a browser

`bestax.io` and `*.bestax.pages.dev` are blocked by this session's network policy, and #421–#423 get no preview anyway (`ci.yml` and `test-deploy.yml` only fire on PRs targeting `main`). So I built both sides locally and drove them in Chromium: `origin/main` as the original — byte-for-byte what bestax.io serves — and the stack tip as the new. All 87 pages, weighted to the props tables.

**The headline: the zero-loss claim holds.** Across all 87 pages exactly one prop row is dropped (`card.md`'s `m`/`p`, now covered by the catch-all). Defaults, sub-component tables, the `Omit`/alias/union extractor fixes, live-example counts and rendered-node counts all check out; every props table is 4 columns with 4 `<td>`s per row, no split cells, no marker leakage, no NEW-only console errors.

What did not hold up is a set of rows describing props the components reject, and some type cells that got **less** informative than the hand-written originals. Those are fixed and pushed.

### Fixed on this branch

`gen:catalog:check` was red here for a real reason, not a flake: `scripts/gen-component-catalog.mjs` carried a copy of the preamble predating #365, so regenerating deleted 21 lines of shipped guidance (the per-component value-union warning, where to find the installed `.d.ts`, the `Skeleton` exception, which sub-component families take helper props, the composition spine). `#421` had already committed that loss into `component-catalog.md`. Restored here and regenerated down the stack.

Four extractor defects, all found by diffing generated pages against the interfaces they claim to describe:

| Defect | Symptom |
|---|---|
| A one-member "union" (`type CellSpanValue = number`) is not a union node | `cell.md` `colSpan`/`rowSpan` said `CellSpanValue` where main said `number` |
| A union naming another alias failed the all-members-simple test | `grid.md` `fixedCols` opaque beside five siblings expanded to `0 … 12` |
| `children` synthesized for any DOM-based interface | `divider.md` documented the one thing React throws on — `<hr>` is void |
| `never` / `false \| true` merged from discriminated-union branches | `slider.md` `minDistance: never \| number`, `range: false \| true` |

Plus sub-components whose props are an inline DOM type were dropped entirely — which is *why* navbar and pagination still carried hand-written duplicate lists. `Navbar.Divider` and `Pagination.Ellipsis` are now generated.

Verified after: generator idempotent, 10/10 conformance, `check-docs-parity` still 0 prop/default/description losses, docs build clean, bulma-ui typecheck + lint + tests green.

> I closed and reopened this PR once to force CI — pushes made with the session's app token don't emit `synchronize`, so the red run you see was pinned to the pre-fix SHA.

### Remaining, all generator-level — @claude please pick these up

1. **Catch-all understates `Omit<React.*HTMLAttributes<*>, keyof BulmaClassesProps>`.** `menu.md` (4 tables) and `dropdown.md` (2) render `... | Bulma helper props`, dropping the HTML attributes. They do pass through — `<Menu.Item id="MID" title="MTITLE">` → `<li id="MID" title="MTITLE">`. A regression: main said ``All `<li>` and helper props``. `catchAllRow` never sees a `kind:'dom'` base for that shape.
2. **`DOM_ELEMENT_LABELS` has no entry for `HTMLHRElement`, `HTMLPreElement`, `HTMLProgressElement`**, so `divider`, `pre`, `progress` (and 5 others) lost the concrete element name from the catch-all.
3. **`impliedFalse` shouldn't apply to controlled-mode booleans.** `collapse.md` documents `open` as defaulting to `false`, but `open === undefined` is what selects uncontrolled mode — the row's own description says "If provided, component is controlled".
4. **`skeleton.md`'s catch-all links Helper Props**, but `SkeletonProps` only extends `React.HTMLAttributes` and the component never calls `useBulmaClasses`. Main said "Passed through to the root element."
5. **Bare internal aliases with no `**Types:**` footnote:** `PickerLabels` and `DateFormatOption` on the three datetime pages, `IconTextItem` on `icontext.md`. None is exported from the barrel, so a reader can neither import nor look them up — and `DateFormatOption` is a regression from main's `string | Intl.DateTimeFormatOptions`.
6. **`title.md` and `subtitle.md` render the same unsplit 16-row CSS-variable table** (9 `title-*` + 7 `subtitle-*`). The prefix rule already splits `generic.scss` correctly into Code/Pre/Strong, so this is an inconsistency. `icon`/`icontext` share a milder version.
7. **`navbar.md`'s `aria-expanded` default** is documented `false`; the real default is `!!active`.

### Two claims in the PR bodies that don't match the code

- **`LevelItem` forwarding `rel` is not a behaviour change.** It already forwarded `rel` on `main` (`Level.tsx:238`) and is already tested (`Level.test.tsx:105-127`). The full `Level.tsx` diff is comment-only. The docs addition is correct; the framing isn't.
- **`file.md` had three descriptions changed, not one.** `label`, `isFullwidth` and `className`. All three are now verbatim source JSDoc and the originals were drift, so the edits are right — only the count is wrong.

---

allxsmith:
@coderabbitai full review

---

allxsmith:
## Round 2 — manifests rebased onto main, review findings addressed

### The manifests were reverting main, not just trailing it

Worth separating those two, because only one of them matters at merge time. These branches were cut before **19 commits** landed on main. For most files that is harmless — `bestax-migrate/package.json`, `bulma-ui/package.json` and `create-bestax/package.json` are untouched by the stack, so a merge takes main's side cleanly.

But `package.json`, `docs/package.json` and `pnpm-lock.yaml` **are modified by the stack**, so relative to its own merge base (`69f9fe7`) it actively downgrades:

| | main | this stack |
|---|---|---|
| `@commitlint/cli` | ^21.2.1 | ^21.0.2 |
| `@commitlint/config-conventional` | ^21.2.0 | ^21.0.2 |
| `@semantic-release/github` | ^12.0.9 | ^12.0.2 |
| `semantic-release` | ^25.0.8 | ^25.0.2 |
| `prettier` | ^3.9.6 | ^3.9.4 |
| `react` / `react-dom` | ^19.2.8 | ^19.2.3 |
| `@docusaurus/*` (6 packages) | ^3.10.2 | ^3.9.2 |
| `@fortawesome/fontawesome-free` | ^7.3.1 | ^7.2.0 |

and **drops `docs`' `"test": "node --test \"scripts/*.test.mjs\""`**, which silently disables the `flatten-llms-tabs` tests #408 added. Because these are edits rather than absences, merging would have carried the downgrades into main.

Same root cause as the `gen-component-catalog.mjs` preamble revert from the last round — a wholesale copy of an older tree. The PR body notes that taking `docs/package.json` wholesale "would have dropped the `flatten-llms-tabs` build step"; that step was restored, but the version block and the `test` script were not.

**Fixed:** `package.json`, `docs/package.json`, `bestax-migrate/package.json` and `pnpm-lock.yaml` are now main's, with only this stack's own script additions re-applied — `gen:api-sources`, `gen:api-docs`, `gen:api-docs:check`, `gen`, the `lint` widening over `scripts`/`docs/scripts`, the root `format:check` fix in `all`, and the marker-stripping step chained after the llms flattener. The stack adds no dependency of its own, so main's lockfile is exactly right; `pnpm install --frozen-lockfile` succeeds against it on all four branches. Every manifest now differs from main by those script lines and nothing else.

### Review findings addressed

- **The alias fixpoint** resolved mixed unions and `indirect` renames in two separate passes, so a mixed union naming a forward-only alias — or a rename pointing at a mixed alias — stayed opaque forever. Both CodeRabbit and the Claude review flagged this independently, which is why I took it first. They now interleave in one bounded fixpoint.
- **`rendersChildren`** matched `/\bchildren\b/` against raw source text, so the word in a comment would have counted. It is an AST walk now.
- **A small local interface named in a type cell** now renders its object shape. `icontext.md`'s `items` is back to main's `{ iconProps: IconProps; text?: string }[]`, and `slider.md`'s `marks` improves on *both* to `{ value: number; label?: React.ReactNode }[]` — neither interface is exported from the barrel, so the bare name was unlookupable.

Both refactors are behaviour-preserving: the generator emits byte-identical output for them today. That's deliberate — they remove latent fragility without moving the 87 pages.

### Refuted

- **"`listOnly` passes the initializer, not the declaration, to `jsdocText`"** (CodeRabbit) — not a live defect. `ts.getJSDocCommentsAndTags` walks up to the variable statement, and both list-only subs do render their summaries: `Navbar.Divider`: *Divider in dropdown menus.* and `Pagination.Ellipsis`: *Ellipsis separator.* I left it alone rather than churn working code.
- **"The new extractor branches are untested-by-output"** (#421 finding 3) — true of #421's pages in isolation, but `never`/`boolean` fires on `slider.md`, `listOnly` on `navbar`/`pagination`, and `rendersChildren` on `divider`/`icon`/`delete` — all at the stack tip. It's an artefact of reviewing one slice of a stack, not a gap.
- **`useprefixedclassnames.md`'s `createPrefixedClassNames` rest param** — the line already carries `// args: same union as classNames`. Spelling the union out twice on adjacent lines reads worse.

### @claude — still open

1. Catch-all understates `Omit<React.*HTMLAttributes<*>, keyof BulmaClassesProps>` on `menu.md` (4 tables) and `dropdown.md` (2) — they render `... | Bulma helper props`, dropping the HTML attributes, which is a regression against main.
2. `collapse.open` documented as defaulting to `false`, but `undefined` is what selects uncontrolled mode.
3. `navbar.md`'s `color` rows on Brand/Burger/Menu/Start/End describe an `is-*` modifier; at runtime the prop emits `has-text-*` and `NavbarBrand` never reads it.
4. `title.md` / `subtitle.md` share one unsplit 16-row CSS-variable table.
5. `control.md`'s orphaned `\* [Bulma color] options include: …` footnote, which also names values not in `validColors`.
6. `card.md`'s dropped `m`/`p` row — the one genuine loss across all 87 pages.

Verified before pushing: generator idempotent, 10/10 conformance, `gen:catalog:check` exits 0, `pnpm install --frozen-lockfile` clean, and the 87-page cell-by-cell diff against main unchanged at one lost prop row.

---

allxsmith:
@coderabbitai full review

---

allxsmith:
## The four one-shot migration scripts should come out before merge

Flagging this now rather than at merge time so there's room to disagree.

Four scripts this PR adds are invoked by **nothing** — not `package.json`, not `ci.yml`, not `check-conformance.mjs`, and not a single `.md` in the repo (I grepped all four surfaces):

| script | lines | state once this lands on main |
|---|---|---|
| `scripts/check-docs-parity.mjs` | 421 | defaults to `--base=origin/main`, so it compares main against main and reports "0 losses" |
| `scripts/check-docs-wording.mjs` | 197 | every prop buckets as `unchanged` |
| `scripts/migrate-api-pages.mjs` | 490 | migration complete — all 87 pages are managed |
| `scripts/codemod-property-tsdoc.mjs` | 782 | `@property` blocks are all migrated |

That's **1,890 lines nothing calls**. `check-docs-parity.mjs:21-23` already says as much:

> *This is a migration tool, not permanent CI: once the migration lands, `origin/main` becomes the new baseline and `docs-generated` in check-conformance.mjs holds the line instead.*

Two things push this past "harmless dead code":

1. **This PR widened `lint` to `turbo run lint && eslint scripts docs/scripts`** — deliberately, and it was the right call. But it means these 1,890 unreferenced lines get linted and prettier-checked on every CI run, and have to be carried through every eslint / prettier / Node upgrade, forever, for zero ongoing value.
2. **After merge they don't fail, they mislead.** A green `check-docs-parity` run reads as "0 losses across 87 pages" while it is diffing a tree against itself. That's a worse failure mode than a script that errors out — it's a check that looks like evidence and isn't.

For contrast, `scripts/` on main holds five files and every one is invoked: `check-conformance.mjs` (CI), `check-pointer-urls.mjs` (`check:urls`), `gen-component-catalog.mjs` (`gen:catalog`), `auto-close-duplicates.mjs` (its workflow), plus the conformance baseline. The convention there is that everything in the directory earns its place by being called. `gen-api-docs.mjs`, `gen-api-sources.mjs` and everything under `scripts/lib/` clearly do; these four don't.

Deleting loses nothing that matters — git history keeps them, and this PR is the durable record of how the migration was validated (which is the part worth preserving, and it's in the description and the review threads, not the code).

**Plan:** remove all four in a single commit on this branch, propagated up the stack so the tip doesn't resurrect them. **Not yet, though** — `check-docs-parity.mjs` is what I've been running each round of the current review loop to catch dropped prose, URLs and live examples, so it stays until the loop converges and goes in the last commit before merge.

Shout if you'd rather keep the two codemods around for a future category — but `MANAGED_CATEGORIES` covers all 87 pages now, and a brand-new component's page comes from `gen-api-docs.mjs` plus hand-written prose, not from the migration codemod.

---

allxsmith:
@claude please do a full adversarial review of this PR.

Flagging why I'm asking here rather than relying on the label: **the `deep-review` label has never fired `claude-review.yml` for this PR.** I've applied it three times (01:57, ~03:22, ~04:07 UTC) and across the last 30 runs of that workflow there are **zero** for `claude/api-docs-tooling` — while `claude/api-docs-layout`, `claude/api-docs-elements` and `claude/api-docs-components-form` each fired 2–3 times from the same batched label operations. The other three have all now converged at 0 blocking findings; this one has never been reviewed at all.

That matters because #420 carries the highest-risk code in the stack, and it is the only part that is *not* generated output:

- `scripts/lib/props-extract.mjs` (~1,345 lines) — the whole TypeScript-compiler-API extractor
- `scripts/gen-api-docs.mjs`, `scripts/lib/{api-page,api-sources,scss-vars}.mjs`
- the two new `check-conformance.mjs` sub-checks (`docs-generated`, `docs-section-order`) that are supposed to hold the line after merge
- `docs/scripts/strip-generated-markers.mjs`, which rewrites built output
- the root `package.json` / `docs/package.json` / `pnpm-lock.yaml` rebase

Please focus on:

1. **The extractor changes made during review** — the alias fixpoint now interleaves mixed unions with `indirect` renames; `namesChildren` is an AST walk gating the synthesized `children` row; `interfaceShape` inlines small local interfaces; `catchAll` returns `{text, helpers}`; `listOnly` subs; the `never` / `false|true` normalisation. Are any of these wrong in a way the generated pages don't currently expose? The reviews on #421/#422 both noted these branches are partly dead code *on those PRs* — they only produce output further up the stack.
2. **Can the CI gates pass vacuously?** `MANAGED_CATEGORIES` is empty on this branch, so `docs-generated` has nothing to police here. Confirm it is genuinely inert-but-correct rather than silently broken, and that it fails loudly on drift once categories are added.
3. **The manifest rebase.** These branches were *reverting* main's dependency versions rather than trailing them (they modify those lines relative to merge base `69f9fe7`), and had dropped `docs`' `"test": "node --test \"scripts/*.test.mjs\""`. I restored `package.json`, `docs/package.json`, `bestax-migrate/package.json` and `pnpm-lock.yaml` from main and re-applied only this stack's script additions. Please verify nothing else was lost and that no *new* divergence was introduced.
4. **`strip-generated-markers.mjs`** — it runs over built output. What it globs, whether it can corrupt non-generated files, and whether it fails loudly if it matches nothing.

Worth knowing: `ci.yml` has not run against any commit on this branch since the original failing one. Pushes from this session's app token do not emit `synchronize` events, so the red "Build and Test" you see is pinned to the pre-fix SHA. I verified the exact failing step locally on the current head — `pnpm run gen:catalog:check` exits 0 — along with 10/10 conformance, generator idempotence, `pnpm install --frozen-lockfile`, typecheck, lint, tests and a clean docs build. That still wants a real CI run before merge.

---

allxsmith:
@claude all three findings were real and are fixed in `9d682dd` — please re-review.

**1. The `BulmaClassesProps` substring test.** Confirmed exactly as you traced it. `MenuProps extends Omit<React.HTMLAttributes<HTMLElement>, keyof BulmaClassesProps>, BulmaClassesProps` matched on the *first* clause because the name appears in the omit key list, the function returned before the `Omit` branch could unwrap `React.HTMLAttributes<HTMLElement>`, and `HTMLElement` never reached `catchAllRow`. Matched structurally now — the node itself must name `BulmaClassesProps`, or be an `Omit`/`Pick` whose **first type argument** does. After regenerating, menu/dropdown read `All standard <ul> attributes and Bulma helper props`, `<li>`, `<p>`, `<div>` — the element names are back.

Worth recording that you found the *mechanism*: I had reported the menu/dropdown symptom two rounds earlier from diffing pages against main, and had it filed as "catch-all understates the Omit shape" without knowing why. Your trace is what made it a one-line fix rather than a hunt.

**2. `impliedFalse` on controlled booleans.** Confirmed. `collapse.md` asserted `open` defaults to `false` while its own description says "If provided, component is controlled" — mutually exclusive claims. A prop whose description says it controls the component now keeps an empty Default. `collapse.open` renders `—`.

**3. `pnpm all` skipping the widened lint.** Confirmed, and this was the best catch of the three because nothing was going to surface it. `turbo.json` declares no `//#lint` root task and only bulma-ui / create-bestax / bestax-migrate define `lint`, each scoped to its own `src` — so the bare `turbo run … lint` inside `all` never reached the root script where `eslint scripts docs/scripts` lives. Exactly the asymmetry you identified against the `format:check` fix. Routed through `pnpm run lint`; verified by planting an unused variable in `scripts/`, which `pnpm run lint` now rejects and previously passed.

Blast radius was 3 pages (menu, dropdown, collapse). Generator idempotent, 10/10 conformance, `gen:catalog:check` exits 0, and the 87-page cell-by-cell diff against main is unchanged at one lost row.

**On your caveat about not being able to execute anything:** noted, and it is worth saying that all three findings were reproduced from static analysis alone and all three were real — that is a better hit rate than the rounds where the sandbox could run things. For this pass, please focus on:

1. The `isHelpers` predicate itself. `Omit<BulmaClassesProps, 'size'>` (Rate) and bare `BulmaClassesProps` must still short-circuit to helpers-only; anything else must fall through to the `Omit`/`Pick` unwrapping. Is there a heritage shape in `bulma-ui/src` that now takes the wrong branch — an intersection, a nested `Omit<Omit<…>>`, or a `Pick` whose first argument is itself an alias for the helper props?
2. The `controlled` detection is a description regex (`/\bcontrolled\b/i` over the member's TSDoc). That is a heuristic on prose, which is the same class of thing you rightly criticised `rendersChildren` for. Which props does it currently hit across all 87 pages, and is any of them a plain flag that merely mentions the word?
3. Whether `all` now covers every gate `ci.yml` runs, or whether another step is still routed through a bare `turbo run` that misses a root-only script.

---

allxsmith:
@claude deep review again — this branch has moved substantially since your last pass and I want a clean read before merge.

Head is now `dc147d3`. Since you reviewed `9d682dd`:

- **All three of your findings are fixed.** The `BulmaClassesProps` substring test is now a structural `isHelpers` predicate; `impliedFalse` no longer applies to controlled-mode booleans; `pnpm all` routes lint through `pnpm run lint` so `eslint scripts docs/scripts` actually runs (verified by planting an unused variable).
- **Main was merged in.** It had moved to `5acafc3` (#417's 188 lines into `check-conformance.mjs`, plus #442) and this PR had gone `dirty`.
- **A deleted conformance check was restored.** Resolving that conflict surfaced that this branch had removed `style-mapping-sync` (#350) outright — header entry, `MAPPING_FILES`, `checkStyleMappingSync`, and its `CHECKS` registration. It exists at this branch's own merge base and on main, is unmentioned in the PR description, and **merged silently** because only the header comment conflicted. Restored verbatim from main. All twelve checks now run and pass.

Please focus on:

1. **The merge resolution itself.** Is `check-conformance.mjs` now a correct union of main's checks and this PR's two new sub-checks, with nothing else lost? The only two removals I believe are intentional are `basename` (now unused) and `registerVarsKeys` (moved to `scripts/lib/scss-vars.mjs` so the generator shares one SCSS parser) — please confirm those and flag anything else.
2. **Whether the same silent-deletion pattern exists elsewhere in this branch.** This is the third instance (after the `gen-component-catalog.mjs` preamble and the manifest downgrades), so the prior is no longer low. Diff the branch against its merge base and look for content that main has and this branch dropped, rather than for things it added.
3. **The `isHelpers` predicate under merge.** `Omit<BulmaClassesProps, 'size'>` and bare `BulmaClassesProps` must still short-circuit to helpers-only; everything else must fall through to the `Omit`/`Pick` unwrapping.

On advisories: I'd rather have a short, high-signal list than a complete one. Please omit notes that only record what the sandbox stopped you running — I run the full gate locally every round and report it, so those cost more than they're worth. Raise something only if you'd act on it.

---

allxsmith:
@claude one correction before you start — **you are not actually blocked from executing things.** Four rounds in a row have spent an advisory on "node execution is gated in this sandbox", and that's a false constraint you can drop.

`claude-review.yml` runs `pnpm install --frozen-lockfile` before handing over, so `node_modules` is present. Its `--allowedTools` grants:

```
Bash(pnpm run:*)   Bash(pnpm test:*)   Bash(pnpm exec:*)   Bash(pnpm --filter:*)
Bash(gh pr view:*) Bash(gh pr diff:*)  Bash(gh api:*)
```

What it does **not** grant is `Bash(node:*)`. Every command in the PR descriptions, `CLAUDE.md` and my own comments is spelled `node scripts/…`, so that's what you've been reaching for, and it gets denied. The `pnpm run` aliases are equivalent and permitted:

|---|---|
| the whole gate | `pnpm run all` |

`pnpm run gen:api-docs:check` is specifically the generator-idempotence check that three separate reviews logged as unverifiable — it regenerates and `git diff --exit-code`s. Please run it rather than filing it as residual risk again.

The only two that genuinely have no alias are `check-docs-parity.mjs` and `check-docs-wording.mjs`, so those really are out of reach — and they're slated for deletion before merge anyway.

Please verify empirically this round instead of statically. If a command still gets denied, say which one and I'll get it allowlisted.

---

PR #421, opened 2026-07-30
docs(bulma-ui): generate the layout, columns and grid API pages (2/4)

allxsmith:
# Pull Request

> **2 of 4**, stacked on #420 — review that first, and merge it first. The diff shown here is incremental.

## Description

First categories through the generator: `layout` (6 pages), `columns` (2), `grid` (2). Plus `helpers/` **reordered but not generated** — four of its six pages document hooks with `## API` and no `## Props` at all, and `config.md`/`theme.md` keep their hand-written tables (`theme.md`'s Props section is a ~350-line prose sub-API, not a table). That is what `ORDERED_CATEGORIES` being wider than `MANAGED_CATEGORIES` buys: order becomes uniform without forcing markers onto pages that cannot fill them.

`@property` blocks move onto the interface members as inline TSDoc, so a description is verifiable by position and reaches users' editors and the shipped `.d.ts`.

**Descriptions were seeded FROM the docs pages, not the other way round.** The hand-written tables are consistently the richer text — `button.md` explains what `ghost` and `text` render as where the source comment only says "Bulma color modifier" — so migrating in the other direction would quietly downgrade every page. Where the two disagree, the page stays authoritative.

Component TSDoc summaries were likewise seeded from the pages. The existing ones ("Bulma Hero component root.", "Container component for Bulma.") were worse than the prose they would have replaced.

## Review focus

**`skills/bestax-custom-component/references/component-catalog.md` is the canary.** The catalog's one-liners come from each page's first Overview sentence, so that diff is where a bad TSDoc seed shows up. It also ships in the `.d.ts` and users' editor tooltips — worth reading rather than skimming.

Worth eyeballing on the preview:

- `layout/container` — **no** CSS & Sass Variables section (Container has SCSS but registers no CSS variables)
- `layout/hero` — `Hero.Head`/`Body`/`Foot` sub-tables, the inlined `HeroSize` union, and exactly 8 CSS variables. Bulma registers `navbar-item-*`, `tabs-*`, `title-*` and `subtitle-*` *inside* Hero's per-colour loop; those are correctly excluded
- `layout/level` — one row, proving the singular `cv.register-var` path
- `layout/media` — the six-line `hsla()` collapsed to one line

## Verification

- `check-docs-parity` — **0 losses across all 87 pages**
- Generator idempotent; `pnpm all` green
- Six props were *added* that the hand-written tables had missed: Hero `textColor`; Level `as`, `hasTextCentered`, `href`, `target`, `rel`

## Related Issue(s)

Refs #399

## Type of Change

- [x] Documentation
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance
- [ ] Build tooling

Affected packages:

- [x] bulma-ui (`@allxsmith/bestax-bulma`) — comments only, but the summaries ship in the `.d.ts`
- [ ] create-bestax

## Checklist

## Additional Context

One decision worth an opinion: `footer.scss` declares `$footer-color: false !default`. Rendered verbatim as `` `false` `` rather than `—`, since sass-false means the `color:` declaration is skipped entirely and `—` would read as "no default". Easy to flip.

---

## Summary by CodeRabbit

- **Documentation**
  - Improved API documentation across columns, grid, helpers, and layout components.
  - Added clearer generated overviews, import guidance, prop references, defaults, valid-value links, subcomponent details, and standard HTML attribute support.
  - Added CSS/Sass variable documentation for relevant components.
  - Clarified responsive behavior, spacing options, color types, and fixed-grid settings.
  - Reorganized helper and configuration documentation for easier navigation.
  - Refined component prop descriptions without changing runtime behavior or public APIs.


allxsmith:
@coderabbitai full review

---

allxsmith:
## Review — layout / columns / grid verified against main in a browser

Built `origin/main` and the stack tip locally and compared all of this PR's pages in Chromium (bestax.io and the `*.pages.dev` previews are blocked by this session's network policy, and stacked PRs get no preview deploy anyway).

**Clean on the substance.** No prop lost on any of the 16 pages, no invented props, defaults correct, and the six props this PR *adds* that the hand-written tables had missed are all genuine interface members. Hero's CSS-variable table is exactly the 8 you claimed — `navbar-item-*`, `tabs-*`, `title-*` and `subtitle-*` from the per-colour loop are correctly excluded and correctly appear on their own pages instead. `container.md` correctly has no CSS section. `columns.md`'s `gap` renders fully expanded, not a bare `BulmaGapValue`. All 62 CSS tables are 3 columns with 3 `<td>`s per row, every variable is genuinely registered in the SCSS, and live-example counts match main page-for-page.

### Fixed and pushed here

- **The catalog regression** — this branch had committed the shortened preamble into `skills/bestax-custom-component/references/component-catalog.md`, deleting 21 lines of shipped guidance that #365 added. Generator restored from main and regenerated; the preamble matches main again and the component one-liners are untouched.
- **`cell.md`'s `colSpan`/`rowSpan`** rendered as a bare `CellSpanValue` — an alias for exactly `number`, so strictly less than main's `number`, with no `**Types:**` footnote anywhere on the page. The same defect class as the `gap`/`BulmaGapValue` one you fixed; this instance was missed because a one-member "union" is not a union node.
- **`grid.md`'s `fixedCols`** was the only opaque cell in a table where its five per-breakpoint siblings expand to `0 | … | 12`, because `BulmaFixedGridColsProp = BulmaFixedGridCols | 'auto'` has a member that names another alias.

Extractor fix is on #420; this branch just carries the regenerated pages. 10/10 conformance, parity still 0 losses.

### One judgement call for you

`footer.md` renders `--bulma-footer-color | $footer-color | false`. That is verbatim-faithful, but `false` is not a colour, and `footer.scss:20-22` only emits `color:` when the variable is truthy — a reader sees a boolean where a colour belongs. A short footnote ("`false` = Bulma emits no `color` declaration") would read better than either `false` or `—`. Same pattern on `$title-family`/`$subtitle-family`.

Minor, same family: `grid.scss`'s `$grip-gap` is upstream Bulma's real typo so reproducing it is correct — but it carries no `!default`, so listing it under "Sass Variable" implies an override point that doesn't exist.

@claude the `title`/`subtitle` CSS-variable split is tracked on #420 (item 6) since it's a generator fix.

---

allxsmith:
@coderabbitai full review

---

allxsmith:
## Round 2 — findings addressed

Both reviewers' findings on this PR are handled. Full detail of the manifest issue is on #420; the short version is that these branches were reverting main's dependency versions rather than merely trailing them, and that is fixed across the stack.

**Fixed here:**

- **The stacked JSDoc on the six `gapSize*` props** (`Columns.tsx:60`, the one item either reviewer marked blocking) — the migration added the new block on top of the old rather than replacing it, so six dead comments shipped in the source and the `.d.ts`. Collapsed into one block each, keeping the tie-break clause the original had and the page had lost: the cells now read *"@deprecated Use `gap` instead — `gap` wins if both are set."*
- **The alias fixpoint** (finding 2, and CodeRabbit's `props-extract.mjs:399-413`) — mixed unions and `indirect` renames resolved in two separate passes, so a mixed union naming a forward-only alias stayed opaque. They interleave now.
- **`rendersChildren`** (finding 4) is an AST walk instead of a text regex.
- **Two `helpers/` defects** CodeRabbit flagged: `theme.md` had a stray four-backtick fence trapping the "Theme with Styling" heading *and* its example inside a code block, and `usebulmaclasses.md` claimed a `className` the example never passes plus `has-text-info-mobile` for `color: 'link'`. Both are **pre-existing on main** and merely relocated by this PR's reorder — so they render broken on bestax.io today — but they're on pages this PR already touches, so fixing them here is cheaper than a follow-up.

**Refuted:**

- **Finding 3, "the new branches are untested-by-output"** — true of this PR's pages in isolation, but `never`/`boolean` fires on `slider.md`, `listOnly` on `navbar`/`pagination`, and `rendersChildren` on `divider`/`icon`/`delete`, all at the stack tip. It's an artefact of reviewing one slice of a stack.
- **CodeRabbit on `useprefixedclassnames.md`** — the line already says `// args: same union as classNames`; spelling the union out twice on adjacent lines reads worse.
- **Finding 1, the catalog preamble** — already fixed in the previous round; the trimmed guidance is restored, not merely duplicated elsewhere.

**Acknowledged, not changed:** `grid.md`'s `$grip-gap` (finding 2 on the earlier review) is upstream Bulma's real typo and *is* the actual override name, so reproducing it is correct — though it carries no `!default`, so it isn't overridable at all, which is worth a footnote someday. `footer.md`'s `false` default is the same judgement call raised in the PR body; I'd still favour a short footnote over either `false` or `—`.

Verified: generator idempotent, 10/10 conformance, `gen:catalog:check` exits 0, `pnpm install --frozen-lockfile` clean.

---

allxsmith:
## Round 4 — the blocking finding was right, fixed in `86a2c4d`

**`Level.isMobile` was documented backwards.** Verified against Bulma's own source: `bulma/sass/layout/level.scss:27-30` gives `.level.is-mobile { display: flex; flex-direction: row; }` — `is-mobile` keeps the level **horizontal** on mobile, and a level *without* it stacks vertically below tablet. The row said "Enables mobile layout (stacks vertically on mobile)", and the Usage prose repeated it.

Worth being precise about whose bug it is: **the wording is pre-existing on `main`**, in both the props table and the prose, so this stack didn't introduce it. What the stack does is *launder* it — the migration seeds TSDoc from the doc pages, so a page-level error becomes a shipped `.d.ts` comment and an editor tooltip. That's a wider blast radius than the page alone, which is exactly why it's worth fixing here rather than filing separately. Corrected at the source so the generated row follows, plus the hand-written paragraph.

Good catch on checking the other behavioural props for the same inversion (`Columns.isMobile`/`isDesktop`, `Container.fluid`/`widescreen`/`fullhd`/`isMax`, `Hero.fullheightWithNavbar`) and finding `Level.isMobile` the lone outlier — that's the check that makes the finding trustworthy rather than a one-off.

**Advisory 2 — the missing `strip-generated-markers.mjs` guard — was correct and is already resolved.** You diagnosed it as stack drift rather than intentional removal, and that's exactly what it was: the guard landed on `claude/api-docs-tooling` after this branch was last synced. Merging main down the stack (`f6d9da1`) brought it. Confirmed present on all four branches now.

While here: main had moved and #420 had gone `dirty`. Merging it down surfaced something bigger than the conflict itself — the stack had **deleted the `style-mapping-sync` conformance check** (#350) outright: header entry, `MAPPING_FILES`, `checkStyleMappingSync` and its `CHECKS` registration. Present at the stack's own merge base and on main, absent from the branches, unmentioned in any PR description, and it merged *silently* — only the header comment conflicted. Restored verbatim. All twelve checks now run and pass on every branch, including main's `publishable-manifests` (#412) which the stack also hadn't seen.

Gate at the tip: generator idempotent, 12/12 conformance, `gen:catalog:check` exits 0, `pnpm install --frozen-lockfile` clean, 87-page cell diff unchanged.

---

allxsmith:
@claude deep review again — head is now `86a2c4d`, well past the `5711850` you last read.

Since your review:

- **Your blocking finding is fixed.** `Level.isMobile` was documented backwards; verified against `bulma/sass/layout/level.scss:27-30` (`is-mobile` sets `flex-direction: row`, keeping the level horizontal on mobile) and corrected in both the source TSDoc and the hand-written Usage prose. Worth noting the wording is pre-existing on main — this stack laundered it into the shipped `.d.ts`, which is why it was worth fixing here.
- **Your advisory 2 is resolved.** You diagnosed the missing `strip-generated-markers.mjs` guard as stack drift rather than intentional removal, and that was exactly right — merging the base branch down brought it. Present on all four branches now.
- **Main was merged down the stack**, and doing so surfaced that the stack had deleted the `style-mapping-sync` conformance check (#350) entirely. Restored; twelve checks now run and pass here.

Please focus on:

1. **Description accuracy on the remaining behavioural props.** You caught `Level.isMobile` by checking siblings for the same inversion, and that method is what made it trustworthy. Please extend it — the migration seeds TSDoc *from* the doc pages, so any pre-existing page error becomes a shipped type comment. `columns`, `grid` and the remaining `layout` props are the surface.
2. **Whether the merge disturbed any generated page.** Everything regenerated after the merge; the pages should be a fixed point.
3. `grid.md`'s `$grip-gap` — you flagged it before as faithful-to-Bulma's-typo. It also carries no `!default`, so it is not actually overridable. Is listing it under "Sass Variable" misleading enough to change?

On advisories: please keep the list short and actionable. Skip notes that only record what the sandbox blocked you from executing — I run the full gate locally each round and report the result.

---

allxsmith:

Use the aliases:

|---|---|

Your last review closed with "Confirm the rebase onto the final #420 restores that guard" — you can now check that yourself: the merge landed and `docs/scripts/strip-generated-markers.mjs` carries the guard on this branch.

Please verify empirically this round. If a command still gets denied, name it and I'll get it allowlisted.

---

PR #422, opened 2026-07-30
docs(bulma-ui): generate the elements API pages (3/4)

allxsmith:
# Pull Request

> **3 of 4**, stacked on #421. Merge #420 → #421 → this. The diff shown here is incremental.

## Description

30 pages. Two fixes to the extractor that this category forced — both found by `check-docs-parity`, not by reading pages:

- **Heritage through `Omit`/`Pick` and imported (alias) symbols.** `LinkButtonProps extends Omit<ButtonProps, …>` resolved the symbol of `Omit` — a lib type — so all 14 of its inherited props landed in the catch-all row instead of a table. An imported base arrives as an *alias* symbol whose `declarations` are the import specifier, not the interface, so `getAliasedSymbol()` is now followed.
- **`Table`'s six sub-components are imported rather than declared locally**, so the page rendered no sub-tables at all and `Thead`/`Tbody`/`Tfoot`/`Tr`/`Th`/`Td`'s props vanished. Cross-module subs now resolve through the barrel.

Props the old tables documented but no interface declares — `href`/`target`/`rel` on Link, `value` on ListItem, the `<ol>` attributes on OrderedList, `skeleton` from the helper props — are parked as `@extraProp`, carrying their type and default. Lossless by construction, and it puts the text in the source where the rest of it now lives.

`Button`'s component JSDoc was attached to an unrelated `const validButtonColors` rather than to `Button`, so its summary never reached the page. Moved.

## Verification

- `check-docs-parity` — **0 losses across all 87 pages**
- `check-docs-wording`, restricted to these pages: of **215** previously-documented props, **178 read identically** and **37 keep their sentence with more appended**. None replaced.
- `pnpm all` green

## Related Issue(s)

Refs #399

## Type of Change

- [x] Documentation
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance
- [ ] Build tooling

Affected packages:

- [x] bulma-ui (`@allxsmith/bestax-bulma`) — comments only, plus the `Button` JSDoc relocation
- [ ] create-bestax

## Checklist

## Additional Context

`elements/block.md` is the page that exposed why a phased rollout is worse than either order applied consistently — it still had Props near the top while `layout/` had moved it down. From this PR its category is uniform with the rest.

---


allxsmith:
@coderabbitai full review

---

allxsmith:
## Review — all 30 elements pages verified against main in a browser

Built `origin/main` and the stack tip locally and compared every page in Chromium (bestax.io and the previews are unreachable from this session, and stacked PRs get no preview deploy).

**Zero prop losses across all 30 pages**, and every `@extraProp` in `bulma-ui/src/elements/` is runtime-real — I rendered each one: `Link` href/target/rel → `<a href="/x" target="_blank" rel="noopener">`, `ListItem` value → `<li value="7">`, `OrderedList` type/start/reversed → `<ol type="a" start="3" reversed>`, `Title`/`SubTitle` skeleton → `class="title is-skeleton"`. `linkbutton.md` resolves `extends Omit<ButtonProps, …>` correctly — the three omitted props are absent and `color` is narrowed to the 10-member override. The 19-member colour union is gone from every generated table. Live-example counts match main page-for-page (button 23/23, table 7/7, icon 11/11), all mounting into shadow roots with no NEW-only console errors.

### Fixed and pushed here

- **`table.md` and `figure.md` still carried their pre-migration tables** immediately after the generated block. `table.md` rendered **11 props tables where 7 are intended**, with the two copies disagreeing — `isSelected` defaulted to `false` in the generated table and `—` in the legacy one, and the legacy copies used a bare `TableColor` with no explanation while the generated ones explain it. `figure.md`'s stale `Figure.Caption Props` still dumped the 19-member colour union inline — precisely what this migration exists to delete. Both removed; the generated tables supersede them and keep the element names in their catch-all rows.
- **Three synthesized `children` rows for props the components don't render.** `divider.md` was the serious one: `Divider` spreads onto `<hr>`, so the row documented the one thing React refuses — verified live, `hr is a void element tag and must neither have children nor use dangerouslySetInnerHTML`. `icon.md`'s was silently discarded (`Icon` always supplies its own JSX children — `render(<Icon name="star">HELLO</Icon>)` drops `HELLO`), and `delete.md`'s renders but lands text on top of Bulma's pseudo-element ✕.

The extractor gate that fixes this lives on #420; this branch carries the regenerated pages. 10/10 conformance, parity still 0 prop losses.

### Remaining on these pages — @claude, details and the rest of the list are on #420

- **`skeleton.md`'s catch-all links Helper Props**, but `SkeletonProps` only extends `React.HTMLAttributes` and `Skeleton` never calls `useBulmaClasses`. Verified: `<Skeleton m="4" textColor="primary" />` → `<div class="skeleton-block" m="4" textcolor="primary">` plus a React unknown-prop error. Main said "Passed through to the root element", which was right.
- **Eight pages lost the concrete element name** from the catch-all (`<code>`, `<hr>`, `<em>`, `<figure>`, `<pre>`, `<progress>`, `<strong>`). Three are a one-line fix — `DOM_ELEMENT_LABELS` has no entry for `HTMLHRElement`, `HTMLPreElement` or `HTMLProgressElement`.
- **`icontext.md`'s `items`** regressed from the self-describing `{ iconProps: IconProps, text?: string }[]` to a bare `IconTextItem[]` with no footnote, and the interface isn't exported. `tag.md` handles the analogous `TagColor` correctly with a `**Types:**` block.

---

allxsmith:
@coderabbitai full review

---

allxsmith:
Picked up 8a6e640 — the `DOM_ELEMENT_LABELS` additions and the `catchAll → {text, helpers}` change are both right, and returning `helpers` rather than sniffing the rendered string is the better shape.

One stacking hazard worth flagging, now handled: `scripts/lib/props-extract.mjs` and `scripts/gen-api-docs.mjs` are shared by all four branches, so landing the fix here alone left #420 and #421 without it and made **#423 revert it** — the tip is stacked above this branch and carried the older file. I've applied the identical content to #420, #421 and #423 and regenerated; all four now hash-match on both files.

Regenerating on #423 turned up that the Helper Props mislink was wider than the elements pages — `card`, `message`, `modal`, `navbar`, `pagination`, `panel`, `sidebar` and `tabs` were mislinking it too, and are fixed by the same change.

General note for this stack: a fix to anything under `scripts/` needs applying to every branch (or the stack rebasing), otherwise the tip silently reverts it and the merged result loses the fix.

---

allxsmith:
## Round 2 — the three advisories addressed

The second deep review on this PR came back **0 blocking · 3 advisory** against my fix commit, and it verified the extractor claims empirically rather than from the prose — that's the right kind of check. All three advisories are handled or answered.

**Fixed:**

- **Advisory 1, `rendersChildren` is a raw-text regex** — correct, and the false-positive case (the word `children` in a comment) would silently restore a row for a component that doesn't render it. It's an AST walk now: a `forEachChild` descent looking for an `Identifier` named `children`, which cannot see comments.
- **Advisory 2, the mixed-alias fixpoint resolving before `indirect` renames are folded in** — also flagged by CodeRabbit on #421, and by the #421 deep review. Three independent hits, so I took it first. The two deferred kinds now interleave inside one bounded fixpoint, so a mixed union naming a forward-only alias and a rename pointing at a mixed alias both settle.
- **Bonus, same class:** `icontext.md`'s `items` was a bare `IconTextItem[]` where main spelled the shape out. Small local interfaces named in a type cell now render as their object shape, so it's back to `{ iconProps: IconProps; text?: string }[]`.

**Answered, not changed — advisory 3, `Table.Tfoot`'s catch-all reading "All standard `<thead>`/`<tbody>` attributes":** correct diagnosis, and it is `DOM_ELEMENT_LABELS` keying on `HTMLTableSectionElement`, which all three sections share. Adding `<tfoot>` to that label would fix `Tfoot` and make `Thead`/`Tbody` read wrong instead — the label is per-DOM-interface, not per-component, so there's no value that is right for all three. Fixing it properly means labelling by the rendered element rather than the React type, which is a bigger change than this PR should carry. Left as-is and noted.

Also on this branch: the manifest rebase described on #420 — these branches were reverting main's dependency versions rather than trailing them, and dropping `docs`' `test` script.

Both refactors are behaviour-preserving; the generator emits byte-identical output for them today. Verified: idempotent, 10/10 conformance, `gen:catalog:check` exits 0, `pnpm install --frozen-lockfile` clean.

@claude the remaining items for this category (`skeleton`'s Helper Props catch-all is now fixed by your `8a6e640`; the element-name labels likewise) are tracked on #420.

---

allxsmith:
## Round 4 — 0 blocking, and both advisories answered with measurements

Thanks for attacking the two things I asked you to attack. One held, one didn't, and I'd rather record both.

**Refuted by your own check — `isHelpers`.** You confirmed exactly one page (`skeleton.md`) lost "Bulma helper props", and that `Skeleton.tsx` genuinely extends only `React.HTMLAttributes<HTMLDivElement>` with no `useBulmaClasses`, while components reaching helpers transitively (`LinkButton` via `Omit<ButtonProps>`) keep the link. That's the right test and it's the answer I wanted — the structural predicate does what the substring test was supposed to.

**Advisory 2, the `controlled` regex — measured, and it is currently exact.** You were right to flag the heuristic class; I flagged it against myself for the same reason. So I counted rather than argued. Across all 87 pages, exactly **three** optional-boolean rows have `controlled` in their description:

| page | prop | default | description |
|---|---|---|---|
| `components/collapse.md` | `open` | — | Controlled open state. If provided, component is controlled. |
| `components/dropdown.md` | `active` | — | Whether the dropdown is open (controlled). |
| `form/switch.md` | `checked` | — | Controlled checked state. |

All three are genuinely controlled props, so there are zero false positives today. Two of them (`collapse.open`, `dropdown.active`) are *restored* to what `main` says — the synthesized `false` was the regression. `switch.checked` was already `—`.

The fragility you describe is real but latent: a future plain flag whose prose happens to say "controlled" would silently lose its `false`. I've left the heuristic in place because the alternative — an explicit opt-out tag — adds ceremony to three props, and the failure mode is a missing default rather than a wrong one. Worth revisiting if the count grows; recorded here so the next person has the number.

**Advisory 1, `namesChildren` matching any `children` identifier** (including an unrelated `node.children` access) — accepted as a real latent issue, same shape. Your own sweep found no false positive in the current set, and the containment is that it can only *add* a row the catch-all already covers, never drop one. Left as-is.

Also on this branch since your review: main had moved and the stack had gone stale. Merging main down surfaced that the stack had **deleted the `style-mapping-sync` check** (#350) entirely — silently, since only the header comment conflicted. Restored; all twelve checks now run and pass here.

---

allxsmith:
@claude deep review again — head is now `4789784`, past the `c29a6a2` you last read.

Since your review:

- **Your two advisories are answered rather than silently kept.** The `controlled` heuristic: I counted instead of arguing, and it hits exactly three optional-boolean rows across all 87 pages — `collapse.open`, `dropdown.active`, `switch.checked` — all genuinely controlled, zero false positives, and two of them *restored* to what main says. `namesChildren` matching any `children` identifier: accepted as latent, contained by the fact that it can only add a row the catch-all already covers, never drop one. Both left in place, both recorded.
- **Main was merged down the stack**, and that surfaced a deleted conformance check — `style-mapping-sync` (#350) had been removed from the stack outright and merged silently. Restored; twelve checks now run and pass here.

Please focus on:

1. **Your own refutation, re-checked under the merge.** You confirmed `skeleton.md` is the only page that correctly loses "Bulma helper props" and that `LinkButton` keeps it via `Omit<ButtonProps>`. The merge brought main's `check-conformance.mjs` changes in — please confirm that still holds on the merged tree.
2. **The `elements` pages after the merge** — everything regenerated, so they should be a fixed point.
3. **Description accuracy**, the failure class that produced the one blocking finding on #421: `Level.isMobile` was documented backwards (pre-existing on main, laundered into the shipped `.d.ts` by the migration). The same risk applies to every behavioural prop on the elements pages. Sibling-comparison against Bulma's SCSS is the check that caught it.

On advisories: short and actionable, please. Skip anything that only records what the sandbox blocked from running — I run the full gate locally each round and report it, so those notes displace signal.

---

allxsmith:

Use the aliases:

|---|---|

Your review said "wrong-but-present output escapes the parity gate, so I traced the highest-leverage paths by hand" — that reasoning was right, and it's exactly why running the generator matters: `pnpm run gen:api-docs:check` regenerates and `git diff --exit-code`s, which catches wrong-but-present output that hand-tracing can miss.

If a command still gets denied, name it and I'll get it allowlisted.

---

PR #423, opened 2026-07-30
docs(bulma-ui): generate the components and form API pages (4/4)

allxsmith:
# Pull Request

> **4 of 4**, stacked on #422. Merge #420 → #421 → #422 → this. The diff shown here is incremental.

## Description

The last 41 pages — `components` (22) and `form` (19) — completing the migration. All 87 pages now share one section order and one table shape.

Four extractor fixes this pair forced, each surfaced by `check-docs-parity`:

- **Props types that are type ALIASES, not interfaces.** `ControlProps` and `SliderProps` are unions, and treating a props type as necessarily an interface rendered both pages with **no table at all** — 42 props. Unions and intersections now contribute every prop they mention, which is what a catalogue should do.
- **Defaults destructured in the function body** rather than the parameter (`Slider`, `TimeInputBase`), and defaults belonging to a **base module a thin wrapper renders** (`TimeInput` → `TimeInputBase`). `editable` read as `false` on `timeinput.md` where the source says `true`.
- **No bare alias names in type cells.** Short unions inline (including `(typeof xs)[number]` and `AvatarProps['size']` lookups); long ones keep the name and get a `**Types:**` footnote built from the alias's own TSDoc. `columns.md`'s `gap` had regressed from `number | string (0-8)` to a bare `BulmaGapValue`.
- **`**Subcomponents:**` renders each sub's summary sentence**, which the hand-written pages carried ("Top bar for navigation or branding") and a bare name list dropped. A sub-component with its own page is linked rather than restated, so `Columns.Column` does not duplicate `column.md`'s 25 rows.

Two pages needed a hand edit first: `components/tabs.md` and `form/taginput.md` had prose sitting **between** prop tables, where the marker design has nowhere to put it.

## Verification, across all 87 pages

- `check-docs-parity` — **0 losses**: no prop, default, description word, code span, URL, live example or prose line dropped.
- `check-docs-wording`, of **1083** previously-documented props:

| | |
|---|---|
| unchanged | **953** |
| extended (sentence intact, more appended) | 117 |
| merged (sentence intact inside a longer one) | 12 |
| **replaced** | **1** — `form/file.md` `label`, where I added a comma |

- `pnpm all` green; generator idempotent

## Related Issue(s)

Refs #399

## Type of Change

- [x] Documentation
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance
- [ ] Build tooling

Affected packages:

- [x] bulma-ui (`@allxsmith/bestax-bulma`) — comments only
- [ ] create-bestax

## Checklist

## Additional Context

**The wording rule, worth knowing before reviewing descriptions.** The merge originally preferred whichever text had the larger word set, which let a stale source comment overwrite a hand-corrected page — "Offset for mobile." became "Mobile column offset." on six rows, and 114 props changed that way. The page is now authoritative for any prop it already documented; the comment only wins when it fully *contains* the page's sentence. Where they disagree the codemod prints what it discarded, so the difference is reviewed rather than lost.

**CSS variables.** Six components had none where they should have: `Skeleton`, `Code`, `Pre`, `Strong`, `Delete` and `Column`. Bulma keeps their variables in `sass/base` and `sass/utilities` (which the scan skipped), registers them at `$variables-host` or inside a `@mixin` (not on the component's own selector), and the semantic wrappers own a variable family without rendering a Bulma class at all. Host registrations are claimed by key prefix, which keeps `body-*` and `hr-*` out of `Code`'s table. `Block` genuinely has none — `block.scss` is four lines that `@extend %block`.

---


allxsmith:
@coderabbitai full review

---

allxsmith:
## Review — all 41 components + form pages verified against main in a browser

Built `origin/main` and this branch locally and compared every page in Chromium (bestax.io and the previews are unreachable from this session; stacked PRs get no preview deploy).

**The high-risk claims in the description all hold.** `control.md` and `slider.md` — the alias-union pages that used to render no table at all — now emit complete tables, and both union members contribute (`nameLow`/`nameHigh` exist only on `SliderRangeProps` and are present). `timeinput.md`'s `editable` reads `true`, matching `TimeInputBase.tsx:186`, and all 20 function-body defaults verify 1:1, as do `dateinput`/`datetimeinput` against their `*Base` — including the deliberately differing ones. `taginput.md` and `tabs.md` lost no prose in the hand-edit. Live-example counts, mounted-node counts and interactive-control counts are identical to main on every page, with zero NEW-only console errors.

### Fixed and pushed here

**`@extraProp` tags on the wrong interface, publishing props the components reject.** `RateProps` carried `index`, `isActive` and `isHovered` — fields of `RateIconProps`, the object handed to `customIcon`, which `Rate` never destructures. `CarouselProps` carried `active`, which belongs to `CarouselItemProps`; `<Carousel active>` gives React *"Received `true` for a non-boolean attribute `active`"* and the root node has no such attribute. Both sets were already documented correctly in their own sub-tables. This is the failure mode `@extraProp` invites, so it's worth a second look at the rest.

**`DateTimeInput`'s three `@extraProp` tags omitted the `{type}` braces**, so `name`, `form` and `required` rendered with an empty Type cell and `required` lost its `false` default — the `DateInput` and `TimeInput` siblings had them.

**Five fabricated `children` rows** (`avatar`, `dialog`, `toast`, `Panel.Icon`, `Panel.InputBlock`) plus the form-side set — none of those components render `children`; I rendered each with a sentinel and none reached the DOM. Gated in the extractor on #420; `Modal`'s eight, `Sidebar.Close` and `Steps.Step` are genuine and kept.

**Duplicate content removed.** `navbar`, `panel` and `pagination` each carried a hand-written Subcomponents list beside the generated one, and they disagreed. `sidebar.md` rendered every subcomponent heading twice, emitting duplicate anchors (`sidebarheader-1`, `sidebartitle-1`, …) — that breaks the right-hand TOC and deep links. Removed only after making the generated output strictly richer: `Navbar.Divider` and `Pagination.Ellipsis` now generate (they were dropped because their props are an inline DOM type, and survived *only* via those duplicate lists), and the prose the pages carried but the source TSDoc had lost is restored into the TSDoc — `Panel.Icon`'s "accepts all Icon props" note, both descriptions above, and the five `Sidebar.*` summaries. Per this stack's own rule, where page and comment disagree the page wins.

**Section order** restored to Accessibility → Related on `collapse`, `loading`, `tooltip` and `switch`, which had it inverted against the other 83 pages.

Verified after: generator idempotent, 10/10 conformance, `check-docs-parity` 0 prop/default/description losses, docs build clean, bulma-ui typecheck + lint + tests green.

### Remaining — @claude please pick these up

1. **`navbar.md`'s new `color` rows on Brand/Burger/Menu/Start/End describe behaviour that doesn't exist.** They say "Bulma color modifier", but `color` falls through to `useBulmaClasses` and emits the *text* helper: `<Navbar.Brand color="primary">` → `class="navbar-brand has-text-primary"`, never `is-*`. `NavbarBrand` never reads the prop, and it collides with the separately-documented `textColor`. The text is pre-existing TSDoc, but this PR is what publishes it — main's navbar had no subcomponent tables.
2. **`control.md` carries an orphaned, inaccurate footnote.** The table now links `[Bulma color]` and has no `*` marker, so the surviving `\* [Bulma color] options include: …` line renders as literal text with a dangling reference link. Its list is also wrong — it names `'white-bis'`/`'white-ter'`, which aren't in `validColors`, and omits `'light'`/`'dark'`. Pre-existing inaccuracy, newly orphaned; `control.md` is the only page still carrying it.
3. **`card.md` dropped the `m`/`p` row** — the single genuine prop-row loss across all 87 pages. Reachable via the catch-all's Helper Props link, so arguably fine, but it's the one exception to the zero-loss claim.
4. **`panel.md`'s `Panel.Icon` restates 15 `IconProps` rows** instead of linking `elements/icon.md`. `ownPage` keys on the implementing component name and `PanelIcon` has no page. Defensible, but it duplicates a table that will drift.
5. **Two Import-block comments dropped:** `control.md` lost `// ...and other form elements as needed`, `field.md` lost `// Field.Label, Field.Body are available as subcomponents`.

The remaining generator-level items (catch-all understating `Omit<…, keyof BulmaClassesProps>` on menu/dropdown, `collapse.open`'s controlled-mode default, the bare `PickerLabels`/`DateFormatOption` aliases) are listed on #420 since the fix belongs in the extractor.

---

allxsmith:
@coderabbitai full review

---

allxsmith:
## Exhaustive markdown diff — all 87 pages, every prop cell

Following up the browser pass with a mechanical cell-by-cell comparison of every generated page against its `origin/main` original: prop sets, Type cells and Default cells, normalising only the two substitutions this migration makes deliberately (the 19-member colour union → the Valid values link, and `ReactNode` → `React.ReactNode`). It reads main's 3-column tables too, which is where `card.md` and `button.md` live.

**Result across all 87 pages:**

| | |
|---|---|
| Prop rows lost | **1** |
| Type cells that got *less* informative | **4**, on 2 pages |
| Default cells changed | 39 — none wrong |
| Prop rows added | 237 |

**The single lost row** is `card.md`'s `m`/`p`. Everything else that first looked like a loss was main's combined row correctly split — e.g. `datetimeinput.md`'s `min/max` is now separate `min` and `max` rows.

**The four type regressions** were `DateFormatOption` on the three datetime pages and `IconTextItem` on `icontext.md`. I've fixed `DateFormatOption` — a union member naming an unresolvable type (`Intl.DateTimeFormatOptions` is a qualified name, not an indexed identifier) now falls back to its own source text, restoring main's `Intl.DateTimeFormatOptions | string`. Blast radius is exactly those three cells; no other page moved. `IconTextItem` still needs a hand — it's an `interface`, so neither the inline path nor the footnote path (which only indexes type aliases) reaches it.

Worth recording: every *other* bare alias in a type cell — `AvatarColor`, `BadgeColor`, `TagColor`, `BulmaColumnSize` (12 cells on `column.md`), `BulmaMinColValue` — **does** carry a `**Types:**` footnote on its own page. I checked each rather than assuming.

**The 39 default changes** are 38 × `—` → `false` on optional booleans (documenting the effective default — a convention change, applied consistently) plus one real correction, `autocomplete.data` → `[]` per `Autocomplete.tsx:129`. None disagrees with the source. The only one I'd still question is `collapse.open`, where `undefined` is what selects uncontrolled mode — already noted above.

So the zero-loss claim is essentially true, and now measured rather than asserted.

> Also note: `scripts/lib/props-extract.mjs` and `scripts/gen-api-docs.mjs` are now byte-identical across all four branches. Because they're shared, a fix landing on one branch alone gets reverted by whichever branch sits above it — worth checking before each merge.

---

allxsmith:
## Round 2 — findings addressed at the stack tip

The manifest rebase described on #420 lands here too: these branches were **reverting** main's dependency versions rather than trailing them (they modify those lines relative to their own merge base), and had dropped `docs`' `"test": "node --test \"scripts/*.test.mjs\""`, silently disabling the `flatten-llms-tabs` tests. All manifests are now main's plus only this stack's script additions, and `pnpm install --frozen-lockfile` succeeds.

**Fixed here since the last round:**

- The alias fixpoint now interleaves mixed unions with `indirect` renames — flagged independently by CodeRabbit and by the deep reviews on both #421 and #422.
- `rendersChildren` is an AST walk rather than a raw-text regex.
- `slider.md`'s `marks` was a bare `SliderMark[]` — which main also showed bare, and which isn't exported from the barrel — and now renders `{ value: number; label?: React.ReactNode }[]`.
- The six `gapSize*` props on `Columns.tsx` no longer carry two stacked JSDoc blocks each.

**Earlier finding, now settled:** the first deep review here flagged the Subcomponents list omitting inline-typed subs *and* a duplicate hand-written list surviving on navbar/panel/pagination. Both are resolved, and in the right order — `Navbar.Divider` and `Pagination.Ellipsis` were made to generate **first**, so deleting the duplicate lists lost nothing. The slider `never | number` advisory from that review is also fixed; it reads `number` now, and `range` reads `boolean` instead of `false | true`.

**Still open — @claude, these are the highest-value ones left:**

1. **`navbar.md`'s `color` rows on Brand/Burger/Menu/Start/End** describe a Bulma colour modifier, but at runtime the prop falls through to `useBulmaClasses` and emits `has-text-*`, never `is-*` — and `NavbarBrand` never reads it. Publishing that text is new here, since main's navbar had no subcomponent tables. Either the description should say what it does, or the prop shouldn't be on those interfaces.
2. **`control.md`'s orphaned footnote** — `\* [Bulma color] options include: …` survives with no `*` marker in the table, so it renders as literal text with a dangling reference link, and its list names `'white-bis'`/`'white-ter'` (not in `validColors`) while omitting `'light'`/`'dark'`.
3. **`collapse.open`** documented as defaulting to `false`, where `undefined` is what selects uncontrolled mode.
4. **`card.md`'s `m`/`p` row** — the one genuine prop-row loss across all 87 pages.
5. **`panel.md`'s `Panel.Icon` restates 15 `IconProps` rows** instead of linking `elements/icon.md`, because `ownPage` keys on the implementing component name.

Verified: generator idempotent, 10/10 conformance, `gen:catalog:check` exits 0, docs build clean, `pnpm install --frozen-lockfile` clean, and the 87-page cell-by-cell diff against main still shows exactly one lost prop row.

---

allxsmith:
@coderabbitai full review

---

allxsmith:
@claude deep review again — head is now `783deff`, past the `4d46fbe` you last read.

Since your review:

- **Main was merged down the whole stack**, and that surfaced a deleted conformance check: `style-mapping-sync` (#350) had been removed from the stack outright — header, `MAPPING_FILES`, function and registration — and it merged **silently**, because only the header comment conflicted. Restored verbatim. Twelve checks now run and pass here.
- **`Level.isMobile` was documented backwards** (your counterpart on #421 caught it): Bulma's `is-mobile` keeps a level horizontal on mobile, the description claimed it stacks vertically. Pre-existing on main, laundered into the shipped `.d.ts` by this migration. Fixed at source.

Your advisory 2 last round is still open and I'd like your read on whether it's worth acting on: **`SliderMark.value` went from "Position of the mark along the slider range." to "Controlled value."**, and single-mode `minDistance?: never` picked up a range-only `@defaultValue 0`. You noted neither surfaces on the rendered pages, so it is IDE-tooltip-only — but it is still a shipped `.d.ts` regression, and the `Level.isMobile` finding is a reminder that type-comment accuracy has a wider reach than the page.

Please focus on:

1. **Whether that `SliderMark`/`minDistance` wording should be corrected**, or whether the union-merge behaviour makes it not worth churning.
2. **Description accuracy across `components` and `form`** — the `Level.isMobile` failure class. The migration seeds TSDoc from the doc pages, so any pre-existing page error becomes a shipped type comment. Sibling comparison against Bulma's SCSS is what caught the one real instance.
3. **The merged tree** — all 41 pages regenerated after the merge and should be a fixed point.

On advisories: keep them short and actionable. Please skip notes that only record what the sandbox stopped you executing — I run the full gate locally every round (generator idempotence twice, twelve conformance checks, catalog gate, `--frozen-lockfile`, typecheck/lint/tests, docs build, and an 87-page cell-by-cell diff against main) and report the result, so those notes cost signal rather than adding it.

---

allxsmith:

Use the aliases:

|---|---|

Your last two reviews both logged generator idempotency as unverifiable residual risk and deferred it to CI — `pnpm run gen:api-docs:check` is exactly that check and it's permitted. Please run it rather than filing it a third time. This branch has all 41 `components`/`form` pages, so it's the one where a non-idempotent generator would actually show.

If a command still gets denied, name it and I'll get it allowlisted.

---

PR #433, opened 2026-07-30
feat(docs): add PackageManagerTabs and register it globally

allxsmith:
Second PR of #402. #418 (the flattener groundwork) has merged, so this targets `main` directly.

The component half. One install command, four package managers, pnpm first and default.

## What's here

- `docs/src/components/PackageManagerTabs/index.js` — consumes the `translate.mjs` vocabulary that landed with the flattener in #418, so the pnpm tab and `llms.txt` cannot drift.
- `docs/src/theme/MDXComponents.js` — global registration.
- One conversion, `guides/intro.md` "Install Dependencies", to prove the pipeline rather than landing the component unused.

## Two decisions worth reviewing

**Global registration over per-file imports.** ~20 pages will use this and none of the 135 `.md` files carries an import line today. Per-file imports would put churn in exactly the diffs a reviewer needs to read closely — the ones nested inside numbered steps. Forgetting a global registration is still a loud failure (MDX throws *"Expected component `PackageManagerTabs` to be defined"* during the SSR prerender, naming the file), so this costs no safety. The tradeoff accepted: reading raw source on GitHub shows an element with no visible definition.

**Shared `groupId="package-manager"`.** This is the reason to use `@theme/Tabs` rather than hand-rolling a switcher: Docusaurus persists the choice in localStorage, so picking npm on one page selects npm on every other page. The homepage hero switcher will read the same slot in the next PR. Note the tab `value`s must stay exactly the `PACKAGE_MANAGERS` strings — Docusaurus silently discards a stored value that isn't valid for the group.

## Verified in the browser

Dev server, `/docs/guides/intro`:

- Four tabs render, `pnpm` active by default.
- Each tab shows the right translation:
  ```
  pnpm add @allxsmith/bestax-bulma
  npm install @allxsmith/bestax-bulma
  yarn add @allxsmith/bestax-bulma
  bun add @allxsmith/bestax-bulma
  ```
- Clicking `npm` switches the panel and writes `docusaurus.tab.package-manager`; the selection survives a reload.
- No console errors and no hydration warnings.
- The swizzled `CodeBlock`'s copy button still appears inside the tab panel (non-`live` blocks pass through untouched, as expected).
- At 375px in dark mode all four labels fit on one line, tab bar doesn't overflow, no horizontal page scroll.

## Verified in the artifacts

The flattened `build/docs/guides/intro.md` twin reproduces the original fence exactly:

````
### Install Dependencies

```bash
pnpm add @allxsmith/bestax-bulma
```
````

So `diff -r` over `*.md` twins and `llms*.txt` against a pre-conversion baseline build is **empty**. The conversion is invisible to agents — that's the invariant every later batch gets checked against, and the reason it's worth establishing on one boring fence first.

`verifyArtifact` reports clean on `llms.txt`, `llms-full.txt` and the converted twin. 70 tests pass, `check:conformance` 8/8, prettier clean.

## Next

PR 2 is the homepage hero switcher; PRs 3-6 convert the docs pages, starting with the two riskiest shapes (the 2-space bulleted case and the fake-numbered-list case) before the 18-fence `react-setups.md`.

## Summary by CodeRabbit

* **New Features**
  * Added synchronized package-manager tabs for npm, Yarn, pnpm, and Bun installation commands.
  * Documentation can now use canonical pnpm commands that automatically translate to equivalent package-manager instructions.
  * Added consistent styling and tab selection persistence across documentation pages.

* **Documentation**
  * Updated authoring guidance for package-manager tabs and JSX formatting.
  * Clarified how generated documentation handles tab content and code examples.

* **Refactor**
  * Simplified documentation builds by removing post-build tab conversion and related validation workflows.

### CODE REVIEW

**Comment by allxsmith** [comment]
## Rebased onto main, review fixes applied — and one blocker found

Rebased cleanly onto `main` (22 commits; zero file overlap, no conflicts).

### Review fixes — `5b0d3e6`

**Copilot 1 + deep review advisory 1 (`command` undefined / empty).** Both close together: an
omitted or empty `command` now throws. Deliberately *not* gated on `NODE_ENV` — `docusaurus
build` prerenders every page with `NODE_ENV=production`, so an unconditional throw is what
turns an authoring slip into a failed build naming the page; a dev-only check would miss it in
CI entirely. The flatten gate already catches a *missing attribute* (its regex requires a
literal `command="…"`), but cannot see `command=""` or `command=" ; "`, which flatten to a
silent empty ` ```bash ` fence. Because the prerender covers every page, this can never fire in
a browser on a site that built successfully.

**Copilot 2 (default tab coupled to array order).** Now a named `DEFAULT_PACKAGE_MANAGER`.
Worth stating why it isn't cosmetic: the flattener collapses every tab group to the pnpm
rendering, so the default tab is what makes the rendered page agree with the LLM artifact.

**Deep review advisory 2 (`[class*='tabItem_']`).** Left as-is with a comment naming the theme
internal it couples to. `.tabs__item` is a stable Infima class but the tab *panel* only has a
hashed CSS-module name, so a loose match is the honest option; the failure mode is a regained
bottom margin, visible on sight.

**Not from review, but it would have bitten #434:** `groupId="package-manager"` here and
`'docusaurus.tab.package-manager'` in #434 were about to be two hardcoded strings with nothing
tying them together. A rename would have silently degraded to "hero and docs no longer share a
choice", with nothing failing. Both now come from `TAB_GROUP_ID` / `TAB_STORAGE_KEY` in
`translate.mjs`, which both files already import.

Also added the authoring convention to `docs/CLAUDE.md` — it documented the component but never
said how to write one — plus a note that `.md` renders JSX here because `markdown.format`
defaults to `mdx`, which `intro.md` is the first page to rely on.

### Blocked on #451

While verifying the build I found that this PR's `intro.md` conversion **silently deletes the
install command from the LLM artifacts**. Not this PR's fault: `docusaurus-plugin-llms` 0.5.0
strips PascalCase JSX tags, and content carried in a *prop* goes with the tag. Since
`<PackageManagerTabs command="…" />` is self-closing, its whole payload is the prop.

The same bug is already live on `main` for `<TabItem label="…">` — those labels are missing
from `llms-full.txt` today. #451 fixes the ordering (flatten into a mirror *before* the plugin
reads it) and adds tests for the failure class.

**Merge #451 first.** Until it lands, this PR would ship a conversion whose entire point —
readers and agents seeing the same command — is defeated for agents.

**Comment by allxsmith** [comment]
@claude please re Review, lots have changed

**Comment by allxsmith** [comment]
## Review round 2 — 2 fixed, 2 acknowledged

### Deep review 1 — `install --frozen-lockfile` → wrong for npm *and* yarn. Fixed.

Good catch, and worse than advisory once you follow it through. npm has no
`--frozen-lockfile`; `npm ci` **is** the frozen install, so `npm install --frozen-lockfile`
did a normal, lockfile-mutating install — the opposite of what the reader asked for, silently.
And `guides/security.md:40` already tells readers the npm equivalent is `npm ci`, so converting
that page would have made it contradict itself on the same screen.

Yarn was wrong too, which the review didn't flag: `--frozen-lockfile` is Yarn Classic. Berry
spells it `--immutable`, and `translate.mjs` already targets Berry for `dlx` ("Yarn Classic has
no `dlx`"), so it was internally inconsistent. bun takes the flag as written.

| | before | after |
|---|---|---|
| pnpm | `pnpm install --frozen-lockfile` | unchanged |
| npm | `npm install --frozen-lockfile` | **`npm ci`** |
| yarn | `yarn install --frozen-lockfile` | **`yarn install --immutable`** |
| bun | `bun install --frozen-lockfile` | unchanged |

Plain `install` is unaffected. The table test covers all four.

### Copilot (suppressed) — a leaked fence delimiter. Fixed, and it was real.

This one deserved not to be suppressed. If MDX doesn't parse the block — the missing blank
lines, which is *the* likely authoring slip and something `docs/CLAUDE.md` now warns about —
the ` ``` ` delimiters arrive as text. I reproduced it:

```
authored  : "```bash\npnpm add foo\n```"
roundTrip : "```bash\npnpm add foo\n```"   ← equal, so the assertion passes
npm tab   : "```bash\nnpm install foo\n```"
```

The round trip provably **cannot** catch it: ` ```bash ` isn't a known verb, so it passes
through untouched on every tab and the equality still holds — while the tabs render fence
markers inside the code block. Now checked explicitly, with a message naming the real mistake.
There's a test asserting the round trip does *not* catch this, so the guard can't later be
deleted as redundant.

### Deep review 3 — component coverage. Partly addressed.

The pnpm inverse moved into `translate.mjs` as `unrenderPnpm` and is now unit-tested; it was
inlined in a JSX file where `node --test` can't reach it, which was the weakest spot given the
round trip is the design's central invariant. Component and tests now share one definition.

`textOf` and the throws still have no direct unit test — that needs a jest/RTL harness `docs`
doesn't have, and adding one is out of scope here. Mitigation worth stating: the throws are
unconditional precisely so the production prerender exercises them across all 144 pages on
every build, so a broken `textOf` fails the build rather than shipping.

### Deep review 2 — `<Tabs>` label loss. Acknowledged, no change.

Known and deliberate, documented in `docs/CLAUDE.md` with the two affected pages named, and
filed upstream as rachfop/docusaurus-plugin-llms#64.

---

21 tests pass, build clean, lint and format clean.

PR #444, opened 2026-07-31
perf(docs): cut live-example CSS cost with shared, deferred stylesheets

### LINKED ISSUE #443: [Performance] Live-example shadow roots reparse ~1.85 MB of CSS per preview
## Affected Package(s)

- [ ] Other (please specify):

---

## Component(s) or Area(s) Affected

The live-example pipeline for the docs site:

- `docs/src/components/ShadowPreview/shadowStyles.js` — the raw CSS set
- `docs/src/components/ShadowPreview/index.js` — homepage/showcase previews
- `docs/src/theme/CodeBlock/index.js` — the swizzled `CodeBlock` that renders every ` ```jsx live ` fence

---

## Description of Issue

Every `live` code fence renders into its own shadow root so Bulma can style the preview without Infima leaking in. That requires injecting Bulma, the icon fonts and the bestax extras as raw CSS strings with `:root` rewritten to `:host`, because `:root`-level custom properties and document stylesheets do not cross a shadow boundary.

The correctness of that approach is fine. The cost model is not — three separate problems:

**1. Nothing is shared between shadow roots.** Each root is given its own eleven inline `<style>` children carrying the whole set. The browser therefore reparses the entire stylesheet set once per preview, and the parsed `CSSStyleSheet` objects are never reused.

**2. Every route pays, even with zero live examples.** `shadowStyles.js` is statically imported by the swizzled `CodeBlock`, and `ShadowPreview` is additionally pulled in by `HomepageCarousel` and `PickerShowcase`. The whole CSS-as-JS payload therefore sits in the chunk every route loads, including the landing page and pages with no live fence at all.

**3. `bulma-prefixed.min.css` is dead weight.** At 726 KB it is 39% of the payload. It exists to serve `classPrefix` demos, but the only `classPrefix="bulma-"` left in the docs is in `docs/guides/getting-started/migration/bulma-ui-4-to-5.md` inside a non-live ` ```tsx ` fence. Every other `classPrefix` in the docs is `bestax-`/`mylib-`/`outer-`, which stock Bulma's prefixed sheet does not cover anyway.

---

## Steps to Reproduce

1. `pnpm exec turbo run build --filter=@allxsmith/bestax-docs && pnpm --filter @allxsmith/bestax-docs run serve`
2. Open `/docs/api/form/datetime/timeinput` (42 live fences — the heaviest page).
3. In the console:

```js
document.querySelectorAll('.live-preview').length;
[...document.querySelectorAll('.live-preview')]
  .reduce((n, e) => n + e.shadowRoot.styleSheets.length, 0);
```

4. Compare against `/docs/guides/getting-started/installation`, which has no live fence, and observe the CSS still ships in the initial JS.

---

## Performance Metrics / Evidence

Raw CSS pulled into **each** shadow root via `!!raw-loader!`:

| Sheet | Raw bytes | Gzip |
| --- | ---: | ---: |
| `bulma@1.0.4/css/bulma.min.css` | 677,931 | 66,406 |
| `bulma@1.0.4/css/versions/bulma-prefixed.min.css` | 726,625 | 67,658 |
| `@fortawesome/fontawesome-free@7.3.1/css/all.min.css` | 90,336 | 22,500 |
| `@mdi/font@7.4.47/css/materialdesignicons.min.css` | 346,626 | 54,407 |
| `material-icons`, `material-symbols`, `extras.css`, `profilecard.css` | ~7,000 | ~1,500 |
| **Total per shadow root** | **~1.85 MB** | **~212 KB** |

Scale: **1,175 live fences across ~110 pages.** Worst offenders — `timeinput.md` 42, `datetimeinput.md` 40, `dateinput.md` 38, `usebulmaclasses.md` 37, `guides/library/elements/index.md` 30, `input.md` 28, `button.md` 23.

Measured on the production build in Chromium:

- `timeinput.md` builds **462 distinct `CSSStyleSheet` objects from ~78 MB of CSS text** at load.
- Initial scripts, per page:

| Page | Initial JS |
| --- | ---: |
| no live fences | 3,210 KB |
| 42 live examples (`timeinput.md`) | 3,254 KB |
| homepage | 2,914 KB |

Note the first row: a page with **no live examples at all** still downloads essentially the same JS as the heaviest one, which is problem 2 above.

---

## Expected Behavior

Per-page CSS parse cost should be **O(1)**, not O(number of examples) — the sheets are identical across previews, so they should be parsed once and shared.

Pages with no live fence should not download the live-example CSS at all.

Rendering must be unchanged in both color modes.

---

## Environment

- Package version(s): `@allxsmith/bestax-docs` (private, current `main`)
- React version: 19.2.8
- Browser (if applicable): Chromium 141
- OS: Linux
- Node version: 22 (CI runs 24)
- Other relevant info: Docusaurus 3.10.2, `react-shadow@20.6.0`, `bulma@1.0.4`

---

## Additional Context

`react-shadow@20.6.0` already supports the fix: it accepts a `styleSheets` prop typed `arrayOf(instanceOf(CSSStyleSheet))` and assigns it to `shadowRoot.adoptedStyleSheets` immediately after `attachShadow`. Parsing once at module scope and adopting everywhere needs no new dependency.

Any fix needs a fallback path — Safari < 16.4 has no constructable stylesheets, and Docusaurus server-renders the component tree — so the existing `<style>`-children rendering has to stay as the fallback.

allxsmith:
# Pull Request

## Description

Every `live` fence renders into its own shadow root, and each root was given its own eleven inline `<style>` children carrying the whole Bulma + icon-font set (~1.85 MB of CSS). Nothing was shared between roots, so the browser reparsed the lot per preview — `api/form/datetime/timeinput` built **462 `CSSStyleSheet` objects from ~78 MB of CSS text** at load. The same payload also rode in the chunk *every* route loads, including pages with no live fence at all.

Three commits, each independently reviewable:

1. **`perf(docs)` — share one parsed stylesheet set.** Parse the sheets once at module scope into constructable `CSSStyleSheet`s and hand the same objects to every shadow root via `react-shadow`'s `styleSheets` prop (which assigns `shadowRoot.adoptedStyleSheets`). Only the per-instance theme-variable sheet is unique per preview. That page now parses **9 shared sheets plus 42 tiny `:host{}` rules**. Also drops `bulma-prefixed.min.css` (726 KB, 39% of the payload, styling nothing), and moves the shadow root behind a dynamic import so the CSS no longer rides every route.
2. **`perf(docs)` — defer previews until they scroll into view.** An `IntersectionObserver` slot mounts each preview once it is within 600px of the viewport. Since the shadow root is already behind a dynamic import, nothing above the fold pulls the CSS chunk at all.
3. **`refactor(docs)` — serve ionicons from this origin, prune dead assets.** ionicons v8 was fetched from `unpkg.com` at runtime despite being a declared dependency; it is now copied into the build by a small plugin. Deletes ~700 KB of Ionicons **v4** CSS/fonts referenced only by a commented-out config block, adds a `_headers` cache policy, and declares `bulma` + `copy-webpack-plugin` in `docs/package.json` (both previously resolved only via hoisting).

Please provide a summary of the changes and which package(s) are affected.

- [ ] Other (please specify):

## Related Issue(s)

Closes #443

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Documentation
- [x] Performance
- [x] Build tooling
- [ ] Other (please describe):

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated Storybook stories as needed (bulma-ui)
- [ ] My changes require a change to the documentation
- [x] Any relevant dependencies are updated

Notes on the unchecked boxes, since they are deliberate rather than skipped:

- **Tests** — this is a docs-site rendering change with no unit-testable surface; it is verified by pixel-diffing the built site (below) rather than by jest. `docs` has no component test suite.
- **Storybook / docs content** — no library component or public API changed, so no story or API page needed updating. `pnpm check:conformance` (all 10 sub-checks) passes.
- **CLAUDE.md** — no command, convention, or package-structure change that the existing files describe.

## Screenshots / Demos

No screenshots, because the point is that there is **nothing to see** — rendering is byte-identical.

Full-page screenshots at 1280×2400 of `button`, `field`, `usebulmaclasses`, `color-palette` and the homepage, in **both color modes**, differ by **zero pixels** before and after. Verified three times over:

1. after the shared-stylesheet rework,
2. again with constructable stylesheets force-disabled, to exercise the Safari < 16.4 / SSR `<style>` fallback,
3. again after deferral, with every preview scrolled into view.

30 comparisons, 0 differing pixels.

## Additional Context

**Measured against the production build** (Chromium 141, per page, scripts only):

| Page | Before | After | |
| --- | ---: | ---: | ---: |
| no live fences | 3,210 KB | **1,076 KB** | −66% |
| 42 live examples (`timeinput.md`) | 3,254 KB | **1,120 KB** | −66% |
| homepage | 2,914 KB | 2,204 KB | −24% |

Total build JS 10,097,034 → 8,692,021 bytes. Stylesheets parsed on `timeinput.md`: **462 → 9** shared.

The shadow CSS stays inline-in-JS rather than becoming a `.css` file, which is what keeps it cheap: it compiles to one content-hashed chunk (~198 KB gzip) shared by all ~110 live-example pages, fetched **once per visit** and then served from cache — and `assets/*` is now marked `immutable` by the new `_headers`, so there is not even a revalidation round-trip.

**Two things I want to flag rather than bury:**

- The `@font-face` strip applied to the shadow sheets is **runtime-only** — it runs at module init, so the raw text still ships in the chunk. It drops ~14 inert rules from the shared sheets once per page load, not per preview. Its value is uniformity (`materialIconsStyles` already did this; the other three sheets never got the same fix), not bytes. Those rules were **inert, not broken**: browsers do not register `@font-face` declared inside a shadow tree, verified in Chromium 141, and the previews issue no failing font requests before or after.
- Deferral means a preview far below the fold is not in the DOM until you scroll near it. `.live-preview-slot:empty` reserves 6rem so a page of deferred examples keeps a sensible scroll length, and previews genuinely on screen still mount on load (`guides/library/elements` mounts two immediately, first slot at 788px).

**Out of scope, worth a separate issue:** `assets/fonts/` is 18 MB, of which Material Symbols alone is 12.8 MB. It is lazy — browsers fetch only the faces actually used — but subsetting those fonts is a far bigger lever on page weight than anything in the CSS pipeline. Untouched here.

**Gates:** `pnpm all` green (21/21) and `pnpm check:conformance` green (all 10 sub-checks — note it is a separate CI step, not part of `pnpm all`). `pnpm check:urls` was not run locally; it needs network access to bestax.io, which the sandbox blocks. It will run in CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DJCcqoU4YxnQrrs2BQqG6g)_

## Summary by CodeRabbit

* **New Features**
  * Added improved live code previews with theme-aware rendering and shadow DOM isolation.
  * Added Ionicons support for documentation previews.
  * Added responsive loading of previews as they enter the viewport.
* **Bug Fixes**
  * Improved preview styling consistency, including Bulma styles, fonts, icons, and color modes.
  * Added fallback support for browsers without advanced stylesheet capabilities.
* **Performance**
  * Reduced initial page work by loading live previews and related assets only when needed.
* **Security**
  * Added stricter response headers and caching rules for documentation assets.


allxsmith:
@claude fix the bad reference spoitted in pr request

allxsmith:
## Rebased onto main — conflicts resolved

Was 28 commits behind (branch point `ccaf512`, 2026-07-30). Rebased rather than merged, so the
four commits stay intact and linear.

**One real conflict**, in `docs/package.json`. Both sides wanted a change to the same
`devDependencies` block:

- this branch adds `copy-webpack-plugin: ^11.0.0` (the `local-ionicons` plugin requires it)
- `main` bumped `docusaurus-plugin-llms` `^0.4.0 → ^0.5.1` in #447

Resolved by keeping both. `pnpm-lock.yaml` auto-merged, and `pnpm install --frozen-lockfile`
passes — which is the version CI runs, so the lockfile genuinely matches the merged manifests.
Verified `copy-webpack-plugin@11.0.0` is present in the lockfile and installed, not just
declared.

`docs/package.json`'s `build` script also changed on `main` while this was open — the
`flatten-llms-tabs` step was removed in #433 — but this branch never touched it, so it merged
cleanly to main's version.

## The composition worth checking

`main` gained `PackageManagerTabs` (#433/#434) since this branch was cut, and that component
renders through `@theme/CodeBlock` — which is exactly the file this PR rewrites (+84/-250). So
they meet. Verified on the rebased build:

- all four tabs render (`pnpm add` / `npm install` / `yarn add` / `bun add`)
- the LLM artifact still reduces to the pnpm fence

Non-live code blocks fall through to `OriginalCodeBlock`, which is why the rewrite doesn't
disturb them.

`pnpm all` passes on the rebased branch: build, typecheck, tests, coverage, bundle stats, lint,
format check, and the Storybook build.

PR #446, opened 2026-08-01
ci: push releases with a scoped GitHub App token

allxsmith:
# Pull Request

## Description

Restores npm publishing. `main` was migrated from classic branch protection to a repository
ruleset; `github-actions[bot]` is not a bypass actor, so `@semantic-release/git`'s commit-back
was rejected with `GH006: Protected branch update failed` and every release since 2026-07-30
12:25 UTC failed in the `prepare` step — *before* `@semantic-release/npm` publishes, which is
why nothing was half-published (npm and the repo agree exactly at 5.8.0 / 3.8.0 / 1.0.0).

The publish job now mints a short-lived GitHub App token, which is the ruleset's only
automation bypass.

- Job permissions drop from `contents/issues/pull-requests: write` to **`contents: read`**;
  `id-token: write` stays for npm OIDC trusted publishing.
- Checkout gains `persist-credentials: false`. The absence of this flag is why the failing
  push authenticated via the `.git/config` extraheader — swapping only the env var would have
  left the old credential in place.
- The token is minted **after** `Build`. `pnpm install` and `pnpm run build` execute
  repo-owned code, and a bypass-capable token must not be in scope while they run — the same
  reasoning as the Cloudflare credential split in #442.

Requires repo secrets `RELEASE_BOT_APP_ID` and `RELEASE_BOT_PRIVATE_KEY` (both already set)
and the `bestax-release-bot` App as a bypass actor on the `main` ruleset (already configured).

- [x] Other (please specify): CI / release infrastructure

## Type of Change

- [x] Bug fix
- [x] Build tooling

## Compatibility notes

- **The release flow is otherwise byte-identical.** The three `release.config.js` files are
  untouched, npm auth stays on OIDC with no `NPM_TOKEN`, and the GPG import step stays — it is
  what actually signs the release commit (`git config --global commit.gpgsign true`).
- **No release is triggered by this PR** — `ci:` is a non-releasing type in all three configs.
- App tokens live one hour; the release steps run ~2 minutes after the build.

## Checklist

Notes on the unchecked boxes: workflow-only change with no unit-testable surface — the
verification is the first release after merge. `prettier` does not cover `.yml`; the three
edited markdown files pass `prettier --check`.

Docs updated: `SECURITY.md` now discloses the bypass and its scope/lifetime, `VERSIONING.md`
names the App as the pusher, and `CLAUDE.md` records it as the one exception to "no direct
pushes to `main`".

## Additional Context

Touches `.github/**`, so it is outside the AI loop by design and requires Code Owner review.
The first release after merge is the end-to-end proof: watch for a `chore(release)` commit
landing on `main` as **Verified**.

## Summary by CodeRabbit

* **Security**
  * Improved release automation security by using short-lived, scoped credentials for publishing.
  * Reduced default workflow permissions to read-only access.

* **Documentation**
  * Documented release automation permissions, token lifetime, and protected-branch behavior.
  * Clarified release scope guidance and verified release commit signing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,21 +220,23 @@ jobs:
       # repo-owned code (postinstall hooks, build config), and a bypass-capable
       # token must not be in the environment while they run. Same reasoning as
       # the Cloudflare credential split in test-deploy.yml (#441).
+      #
+      # How the push authenticates, since it is not obvious: semantic-release
+      # takes its repositoryUrl from each package.json's `repository.url`, not
+      # from the `origin` remote (lib/get-config.js). It first tries pushing to
+      # that URL as-is; that is why the pre-fix failure logged a credential-free
+      # `git push https://github.com/allxsmith/bestax.git` — the checkout's
+      # .git/config extraheader was satisfying auth. With persist-credentials
+      # false that attempt fails, and lib/get-git-auth-url.js falls back to the
+      # one credential in the environment, rewriting the URL as
+      # `x-access-token:$GITHUB_TOKEN@…` (the App-token form). So GITHUB_TOKEN
+      # below is the load-bearing credential for both the push and the API.
       - name: Mint release token
         id: app-token
         uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3
         with:
           app-id: ${{ secrets.RELEASE_BOT_APP_ID }}
           private-key: ${{ secrets.RELEASE_BOT_PRIVATE_KEY }}
-
-      # Checkout persisted no credential, so point the remote at the App token
-      # explicitly rather than relying on semantic-release's auth-URL fallback.
-      # create-github-app-token registers the token as a log mask.
-      - name: Authenticate git as the release app
-        env:
-          TOKEN: ${{ steps.app-token.outputs.token }}
-          REPO: ${{ github.repository }}
-        run: git remote set-url origin "https://x-access-token:${TOKEN}@github.com/${REPO}.git"
 
       - name: Import GPG key
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,19 +170,23 @@ jobs:
     needs: build-and-test
     if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     # OIDC trusted publishing needs id-token: write to mint the short-lived
-    # token npm exchanges with the registry. The remaining scopes are what
-    # @semantic-release/git and @semantic-release/github require to push the
-    # release commit/tag and create the GitHub release.
+    # token npm exchanges with the registry. Everything that writes to the
+    # repository — the release commit, the tag, the GitHub release — runs on a
+    # short-lived GitHub App token minted after the build (see "Mint release
+    # token"), so this job's own GITHUB_TOKEN stays read-only.
     permissions:
-      contents: write
-      issues: write
-      pull-requests: write
+      contents: read
       id-token: write
     steps:
       - name: Checkout code
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           fetch-depth: 0
+          # The steps below run repo-owned code (install, build) before any
+          # release happens; don't leave a pushable credential in .git/config
+          # for them to reach. The release remote is configured after the
+          # build instead.
+          persist-credentials: false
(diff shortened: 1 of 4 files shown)


allxsmith:
## Review response — 2 addressed in `d4749ea`, 1 needs a human check

**Copilot — `SECURITY.md` understates the App's usage.** Valid, and fixed. The bullet claimed
the App was used "solely to push" the `chore(release)` commit and tag; the same token also
authenticates `@semantic-release/github`, which creates the GitHub Release and comments on and
labels the issues a release closes. Understating the privileges of a bypass actor in the
security policy is the wrong direction to be wrong in, so it now names all three permissions
(`contents`, `issues`, `pull-requests`) and distinguishes the writes that actually bypass the
ruleset (the commit and tag) from the API calls that don't.

**Advisory 2 — the `git remote set-url` step is a no-op.** Confirmed and removed. I checked
the source rather than reasoning from the diff: `lib/get-config.js:73` resolves
`repositoryUrl` as `pkgRepoUrl(...) || repoUrl(...)`, so each package.json's `repository.url`
wins and the `origin` remote is never consulted. The step also wrote the token into
`.git/config` — precisely what `persist-credentials: false` was added to prevent — so it was
working against the PR's own rationale.

The real mechanism is now documented where the token is minted. `lib/get-git-auth-url.js`
tries the plain URL first (which is why the pre-fix failure logged a credential-free
`git push https://github.com/allxsmith/bestax.git` — the checkout extraheader was satisfying
auth), and on failure rewrites it using the single credential in the environment. Worth noting
the exact entry, since it is what makes an App token work at all:

```js
GITHUB_TOKEN: isNil(env.GITHUB_ACTION) ? undefined : "x-access-token:",
```

`GITHUB_ACTION` is always set in Actions, and `x-access-token:` is the required prefix for a
GitHub App installation token. Exactly one credential is in scope, so it takes the
`envVars.length === 1` path and returns the auth URL without further verification.

**Advisory 1 — the App's installation permissions.** Not addressed, because it isn't
addressable from the workflow, and the review is right that it's the one thing the diff can't
prove. I tried to verify it from here and can't: `/repos/{owner}/{repo}/installation` needs an
App JWT and `/user/installations` needs a token authorized to a GitHub App, so both 401/403
with a normal user token.

It needs a human with repo admin. At **Settings → GitHub Apps → bestax-release-bot →
Permissions**, confirm all three of `Contents`, `Issues` and `Pull requests` are
*Read and write*.

The severity split is worth keeping in mind:

- Missing **Contents** write → safe. `@semantic-release/github`'s `verifyConditions` checks
  push permission up front, so it fails before `npm publish`.
- Missing **Issues** or **Pull requests** write → the failure lands in the `success` step,
  *after* npm publish and the tag push, reintroducing exactly the half-published state this PR
  exists to prevent. It would also cascade: the three release steps are sequential with no
  `if: always()`, so a failure in bulma-ui's step means create-bestax and bestax-migrate never
  run.

If you'd rather remove the dependency than verify it, setting `successComment: false` and
`releasedLabels: false` on `@semantic-release/github` drops the need for `issues` and
`pull-requests` write entirely — but that changes release behaviour and touches all three
`release.config.js` files, so it belongs in its own PR rather than this one.

allxsmith:
## Advisory 1 — resolved

Repo admin confirmed in **Settings → GitHub Apps → bestax-release-bot → Permissions** that
`Contents`, `Issues` and `Pull requests` are all *Read and write*.

That closes the one residual risk the deep review flagged as unverifiable from the diff: the
`success` step of `@semantic-release/github` (release comments and the `released` label) runs
*after* `npm publish`, so a missing `issues`/`pull-requests` grant would have failed the run
post-publish and left the half-published state this PR exists to prevent — and cascaded, since
the three release steps are sequential with no `if: always()`.

All three review findings are now addressed:

| # | Source | Finding | Resolution |
|---|--------|---------|------------|
| 1 | Copilot | `SECURITY.md` understated the App's usage | Fixed in `d4749ea` — names all three permissions, separates ruleset-bypassing writes from API calls |
| 2 | Deep review | `git remote set-url` step is a no-op for the push | Confirmed against `lib/get-config.js:73` and removed in `d4749ea`; real mechanism documented at the mint step |
| 3 | Deep review | App installation permissions unverifiable | Verified by repo admin (this comment) |

Also worth recording that the SECURITY.md wording is now accurate *because* of this
verification — it documents `contents`, `issues` and `pull-requests` write as the App's
permission set, which matches what is actually installed.

PR #451, opened 2026-08-01
fix(docs): flatten tab JSX before the LLM plugin reads it, not after

allxsmith:
# Pull Request

## Description

Fixes a **live content-loss bug on `main`**: tab labels are missing from the published LLM
artifacts right now. Found while rebasing #433, but it stands alone and should land first.

`docusaurus-plugin-llms` 0.5.0 added an unconditional strip of PascalCase JSX tags
(`src/content.ts`):

```js
// Remove MDX/JSX component tags (PascalCase element names such as <Tabs>,
// <TabItem>, <Admonition>), keeping their inner text content.
cleaned = cleaned.replace(/<\/?[A-Z][A-Za-z0-9.]*\b[^>]*\/?>/g, '');
```

Keeping *inner text* is not keeping *props*. Content carried in an attribute goes with the tag:

- **`<TabItem label="…">`** keeps its body but loses its label, so the `####` headings the
  flattener emits never existed.
- **`<PackageManagerTabs command="…" />`** is self-closing — its entire payload is the prop, so
  the command vanishes completely.

Both are silent. `flatten-llms-tabs.mjs` ran *after* the build, and `verifyArtifact` looks for
JSX that **survived**, never for content that **vanished** — so the build stayed green while
the machine-read copy of the docs quietly lost information. #447 bumped the plugin 0.4.0 →
0.5.1 and was green because no page used `PackageManagerTabs` yet; the `TabItem` half has been
broken since it merged.

Live evidence on `main` — `App root (global)` is a `TabItem` label in `skills/theming.mdx`:

| | source | `llms-full.txt` / per-page twin |
|---|---|---|
| before | 1 | **0** |
| after | 1 | **1** |

## The fix is ordering, not content

`scripts/prepare-llms-source.mjs` (chained ahead of `docusaurus build`) mirrors `docs/` into a
gitignored `.llms-src/` with the tabs already flattened. The plugin reads that mirror via

```js
docsDir: [{ path: '.llms-src', routeBasePath: 'docs' }]
```

a supported shape — `path` is the directory it reads, `routeBasePath` is what URLs are built
from — so published links are unchanged. By the time the plugin runs there is no PascalCase JSX
left, and its strip becomes a no-op instead of a lossy step. `transform()` itself is unchanged;
only *when* it runs moved.

Verification also moves earlier, so a failure now names the source file an author can fix
rather than a generated artifact path.

## Related Issue(s)

Refs #447
Refs #433
Refs #418

## Type of Change

- [x] Bug fix
- [x] Build tooling

## Alternatives considered

Pinning the plugin back to 0.4.0 would also work — 0.4.0 has no such strip, and its caret
ranges (`"minimatch": "^9.0.3"` vs 0.5.1's exact `"9.0.3"`) would let the `minimatch@9`/`yaml@2`
overrides from #447 be pruned. Rejected: staying on current dependencies is the house rule, and
upstream's direction — plain-text artifacts shouldn't contain JSX — is defensible. This change
works *with* that direction rather than against it.

## Checklist

- [x] I have added tests that prove my fix is effective

## Verification

`pnpm --filter @allxsmith/bestax-docs test` — **83 pass** (77 existing + 6 new).
`prepare-llms-source.test.mjs` targets the failure class that let this ship green:

- a self-closing `PackageManagerTabs` becomes a real pnpm fence
- a `TabItem` label survives as a heading, not just its body
- the mirror is rebuilt from scratch, so a deleted page stops appearing in `llms.txt`
- `prepare` refuses a source file the flattener cannot handle, naming it
- the **real 144-page corpus** leaves the plugin nothing to strip
- **the plugin's own regex is applied to the mirror and proven to delete nothing**

Also checked by hand:

- `pnpm run check:urls` passes; `llms.txt` links still resolve to `https://bestax.io/docs/…`
- 144 documents processed, unchanged
- **Control build**: reverting `docsDir` to its old string form makes the labels disappear
  again, confirming the mechanism.
- The one `.md` twin that lands outside `/docs/` (the migration index page, which sets an
  explicit `slug:`) does so **identically before and after** — pre-existing plugin behaviour,
  not caused by this change. Worth its own look sometime.

## Additional Context

I have not filed this upstream yet. It is arguably a real bug — "keeping their inner text
content" is the stated rationale, and prop-borne content isn't covered by it — and there are
currently zero open issues on that repo with a responsive maintainer. Happy to file a minimal
repro if you want; this PR is independent of the outcome either way.

## Summary by CodeRabbit

* **Documentation**
  * Improved documentation build processing for tabbed content while preserving commands, labels, code samples, and tab bodies.
  * Published documentation routes remain unchanged.
  * Added validation to catch unsupported inline formatting and content-processing issues before publication.
  * Generated documentation artifacts are automatically refreshed and excluded from formatting checks.

* **Tests**
  * Expanded coverage for tab rendering, content preservation, stale generated content cleanup, and full documentation validation.


allxsmith:
## Filed upstream

rachfop/docusaurus-plugin-llms#64 — "JSX tag stripping silently drops content carried in props (0.5.0+)".

It reports both cases with minimal repros: the `<TabItem label>` one (which is what is live on
`main` today) and the self-closing prop-carrying one. It offers three directions rather than a
demand — a `preserveComponents` option, a content hook that runs *before* cleaning, or emitting
`label`/`value` as a heading for the `<Tabs>` case the code comment already names — and offers
a PR for whichever they prefer. The workaround this PR implements is written up at the end, so
anyone else hitting it has a way out.

**This PR does not depend on the outcome.** Flattening before the plugin is correct regardless:
it works with upstream's stated direction (plain-text artifacts should not contain JSX) rather
than against it. If they add a hook, the mirror can be swapped for it later without changing
`transform()` or any authoring convention — only where it runs would move again.

allxsmith:
Superseded by a simpler approach — closing in favour of authoring the pnpm block as the component's **children** rather than a prop.

The plugin strips tags but keeps inner text, so a fenced block inside the component survives untouched. Verified by running the real `cleanMarkdownContent` from 0.5.1 over both shapes:

| authoring | plugin output |
|---|---|
| `<PackageManagerTabs command="add foo" />` | `### Install` → `### Next` — **content gone** |
| fenced block as children | `### Install` → ```bash pnpm add foo``` → `### Next` — **exactly right** |

That makes the artifact correct *by construction* rather than correct *by pipeline*, and it removes everything this PR added: the mirror, the `docsDir` indirection, the extra build step, and the two ignore entries. Strictly less machinery for the same result on the component we own.

**Known trade-off, accepted:** this does not help `<TabItem label="…">`, because `label` is Docusaurus's own component API and can't move to children. Those labels stay missing from the artifacts on `skills/theming.mdx` and `skills/custom-component.mdx` — `<Tabs>` collapses to `"body one\n\nbody two"` with nothing distinguishing the options. Tracked by the upstream issue and worth fixing separately.

Upstream rachfop/docusaurus-plugin-llms#64 stays open and still describes both cases.

PR #458, opened 2026-08-02
chore: pin @babel/parser to 7.x and tier the .github agent-edit rule

allxsmith:
Follow-up to #430, which proposed `@babel/parser` 8.0.4 for `bestax-migrate`. That bump is not takeable. This PR records why in the code, stops Dependabot reopening it, and fixes the contract rule that made the second part take three round-trips.

## 1. Why Babel 8 is blocked

Three of the options in `runner.ts`'s `PARSER_OPTIONS` are invalid under 8.0.4. Verified by parsing against the real `@babel/parser@8.0.4` with the exact config from `runner.ts`:

1. **`pipelineOperator: { proposal: 'minimal' }`** — Babel 8 accepts only `fsharp` and `hack`, and validates the options object *before* reading any source, so **every** `js/jsx/ts/tsx` file throws, not just legacy ones. `cli.ts:122-126` catches per file, so a run against any codebase would print `✖` for every file and migrate nothing. Only the Sass path (line-based, never touches Babel) would still work.
2. **`deprecatedImportAssert`** — removed with no replacement (it warns and no-ops), so `import data from './data.json' assert { type: 'json' }` fails with `Missing semicolon`.
3. **`declare module M {}`** — rejected in 8.0.4 (babel/babel#18120), which now requires a string name. Older TypeScript still uses the namespace form.

Also: `@babel/parser@8` declares `engines: node ^22.18.0 || >=24.11.0`, while this package declares `>=22` and its guard at `src/index.ts:15-21` only checks `major < 22` — Node 22.0–22.17, all of 23.x, and 24.0–24.10 fall outside.

CI on #430 surfaced only the typecheck symptom (`'deprecatedImportAssert' is not assignable to type 'PluginConfig'`), which is why the other two needed writing down.

### Changes

- **`bestax-migrate/src/runner.ts`** — the header comment explained `deprecatedImportAssert` only. It now names all three blockers, so the next reader (or bot) sees why the major is pinned instead of rediscovering it.
- **`bestax-migrate/src/sources/react-bulma-components/__tests__/transform.test.ts`** — the parser config had exactly one regression test (import-assert) and nothing covering the rest of the plugin list. Adds two covering shapes an old codebase actually contains: legacy decorators + class properties + private fields + numeric separators, and the `React.createClass` form.

To be accurate about what these tests do and don't buy: they do **not** detect Babel 8 specifically — both cases parse fine under 8 once the pipeline option is fixed. Blocker 1 is already caught loudly, since it fails every existing test. These close the general gap in plugin-list coverage.

## 2. `.github/dependabot.yml` — pin the major

Adds an `ignore` entry so the 8.x bump stops being reopened weekly. Scoped to `version-update:semver-major` only; minor and patch still flow through the `production-dependencies` group.

## 3. `.github/CLAUDE.md` — tier the agent-edit rule

The contract said *"propose a diff in a comment, do not apply one"* for all of `.github/**`. That blanket line was cheap to state, but it spent maintainer attention on a change with no security surface — this PR's one-entry `dependabot.yml` ignore stalled across three separate asks — which is attention not available for the allowlist case that actually needs it.

The change splits two enforcement layers the old line conflated:

- **Automated sessions** (fix loop, `ai-triage`, `ai-scan`) — the mechanical denial is unchanged and now stated as non-negotiable. Those sessions ingest untrusted issue and PR text; editing the doc does not relax `--disallowedTools`.
- **Interactive sessions working with a maintainer** — surface the rule and the security reason, confirm it has been weighed, then act on the answer.

It also bakes in the *reason*, which previously lived only in whichever conversation happened to derive it: **a change here can grant capability with no permissions diff for a reviewer to notice.** That is the sentence that has to travel to other contributors' sessions, because it is what lets a future session reason about a case this one didn't anticipate.

The allowlist row keeps a hard floor — *argue it every time, name the credential, never as a side effect of another task* — so tiering the rule cannot be read as licence to widen a confinement boundary quietly.

## Security review checklist

Per `.github/CLAUDE.md`, since this touches `.github/**`:

- **Does any allowlist grow?** No. No `--allowedTools` or `--disallowedTools` entry is added, removed, or reworded.
- **Does a model session gain the ability to execute, fetch, or reach the network?** No.
- **Does model or issue text reach a comment body?** No. No workflow logic changes.
- **Posting identity?** Unchanged — no workflow file is touched.
- **New repository variable?** No.
- **Action SHAs?** Unchanged; no `uses:` line is modified.
- **Verdict / budget paths?** Untouched. Fail-closed and fail-open behaviour is unaltered.
- **Do the security comments claim exactly what the mechanism delivers?** This is the one to review closely. The new text asserts the mechanical denial on automated sessions is unaffected by the doc edit — that is true (`--disallowedTools` and the fix loop's `.github/**` refusal are enforced outside this file), and it is worth confirming independently rather than taking from the diff.

Net effect on enforcement: **none.** Invariants I1 and I2 are untouched. The only behavioural change is to what an interactive session does when a maintainer has already decided.

## Verification

```
pnpm exec turbo run typecheck lint test --filter=bestax-migrate   # 187 tests pass
pnpm format:check                                                 # clean repo-wide
```

`dependabot.yml` was validated by parsing it and confirming the `labels` and `groups` keys survived the insertion. It is not covered by `format:check` (glob is `**/*.{ts,tsx,js,jsx,mjs,md,mdx}`) and carries a pre-existing double-quote style that predates this PR; I matched the file's existing style rather than reformatting untouched lines.

## Note for the reviewer

This PR does two unrelated things — the Babel pin and a governance change. Happy to split commit `8e83775` onto its own branch if you'd rather review the contract edit separately.

## Summary by CodeRabbit

* **Documentation**
  * Clarified parser compatibility requirements, including Babel 7 support and known Babel 8 limitations.
  * Documented support considerations for legacy import assertions and older TypeScript module declarations.
  * Improved contributor guidance for security reviews and dependency update decisions.

* **Tests**
  * Expanded migration coverage for modern JavaScript syntax, legacy decorators, private fields, and older React component patterns.
  * Added verification for import rewriting and conversion of the `loading` property to `isLoading`.

### DIFF
--- a/.github/CLAUDE.md
+++ b/.github/CLAUDE.md
@@ -21,11 +21,16 @@ waved through — not because agents are untrusted, but because the usual review
 
 Scale the pushback to the blast radius:
 
-| Change                                                                        | What to do                                                                                                                                                                                 |
-| ----------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| `--allowedTools` / `--disallowedTools` on a session holding a credential      | **Argue it properly, every time.** Name the credential in that job, explain why each entry cannot write, and get an explicit decision. Never as a side effect of another task. See rule 2. |
-| Workflow logic, triggers, `permissions:`, action SHAs, verdict/budget paths   | Name the invariant at stake (I1, I2, or the rule below), confirm it holds, then act.                                                                                                       |
-| Config with no security surface — `dependabot.yml` ignores, labels, schedules | Just make the change and say what it does. Do not gate it behind a debate.                                                                                                                 |
+| Change                                                                                       | What to do                                                                                                                                                                                 |
+| -------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `--allowedTools` / `--disallowedTools` on a session holding a credential                     | **Argue it properly, every time.** Name the credential in that job, explain why each entry cannot write, and get an explicit decision. Never as a side effect of another task. See rule 2. |
+| Workflow logic, triggers, `permissions:`, action SHAs, verdict/budget paths                  | Name the invariant at stake (I1, I2, or the rule below), confirm it holds, then act.                                                                                                       |
+| Config with no security surface — labels, schedules, `semver-major` `dependabot.yml` ignores | Just make the change and say what it does. Do not gate it behind a debate.                                                                                                                 |
+
+The `dependabot.yml` qualifier in that last row is load-bearing: a `semver-major` ignore only
+declines a breaking upgrade, but an ignore covering **patch or minor suppresses CVE fixes** for
+that dependency and belongs in the middle row, argued on its merits. Scope every ignore you add,
+and never widen an existing one's `update-types` as a drive-by.
 
 Origin: an interactive session hit the old blanket "propose a diff, do not apply one" line while
 adding a `dependabot.yml` ignore entry — a change with no security surface at all — and stalled
--- a/.github/CLAUDE.md
+++ b/.github/CLAUDE.md
@@ -21,16 +21,29 @@ waved through — not because agents are untrusted, but because the usual review
 
 Scale the pushback to the blast radius:
 
-| Change                                                                                       | What to do                                                                                                                                                                                 |
-| -------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| `--allowedTools` / `--disallowedTools` on a session holding a credential                     | **Argue it properly, every time.** Name the credential in that job, explain why each entry cannot write, and get an explicit decision. Never as a side effect of another task. See rule 2. |
-| Workflow logic, triggers, `permissions:`, action SHAs, verdict/budget paths                  | Name the invariant at stake (I1, I2, or the rule below), confirm it holds, then act.                                                                                                       |
-| Config with no security surface — labels, schedules, `semver-major` `dependabot.yml` ignores | Just make the change and say what it does. Do not gate it behind a debate.                                                                                                                 |
+| Change                                                                                                                      | What to do                                                                                                                                                                                 |
+| --------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `--allowedTools` / `--disallowedTools` on a session holding a credential                                                    | **Argue it properly, every time.** Name the credential in that job, explain why each entry cannot write, and get an explicit decision. Never as a side effect of another task. See rule 2. |
+| Workflow logic, triggers, `permissions:`, action SHAs, verdict/budget paths, anything a workflow reads as a gate or trigger | Name the invariant at stake (I1, I2, or the rule below), confirm it holds, then act.                                                                                                       |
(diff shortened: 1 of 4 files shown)


allxsmith:
@coderabbitai full review

allxsmith:
Both review findings addressed in f35ce7b.

**Claude deep review — `.github/CLAUDE.md:28` (advisory).** Correct, and it's the failure mode this file itself warns about. The row claimed `dependabot.yml` ignores have no security surface, which holds only for the `semver-major` form added here — an ignore covering patch or minor suppresses CVE fixes. Narrowed the row to `semver-major`, moved the patch/minor case explicitly into the middle tier, and added the reason plus a "never widen an existing one's `update-types` as a drive-by" line so the row can't license a future ignore that isn't safe.

**CodeRabbit — `transform.test.ts` (minor).** Also correct. The fixtures contained `#private`, optional chaining, nullish coalescing and a static class property, but nothing asserted any of them survived. That gap matters more than it looks: a plugin dropped from `PARSER_OPTIONS` would still *parse* a file whose syntax has since become standard, while recast quietly lost the node — the test would have stayed green. Now asserts every proposal shape round-trips.

Took the negative assertion too. `expect(output).toContain('isLoading')` passes just as well if the rename *duplicated* the prop rather than replacing it. Verified the pattern discriminates before committing:

```
passes on migrated output (want false): false   // <Button ... isLoading>
catches un-migrated prop  (want true):  true    // <Button ... loading>
catches leftover alongside(want true):  true    // <Button ... loading={x} isLoading>
catches self-closing      (want true):  true    // <Button loading />
```

**One correction to this PR's own description**, which the deep review caught: I framed the Dependabot `ignore` as what keeps Babel 8 away from the codemod. It isn't the primary defence — `"@babel/parser": "^7.26.0"` is a caret range and cannot resolve 8.x regardless. The `ignore` only stops the weekly PR churn. The protection was already in the manifest; section 2 above overstated its role.

Gate re-run: `pnpm format:check` clean repo-wide, 187 tests passing, typecheck and lint green.

allxsmith:
Fixed in a4cf6f9. The finding was correct.

`needs-security-review` is a refusal gate, so calling labels inert was wrong. I grepped the workflows rather than taking it on faith:

| Label | Workflows that read it |
|---|---|
| `needs-security-review` | `ai-scan`, `claude`, `claude-implement`, `claude-repro`, `bestaxbot-reply` |
| `ai-loop` | 7 workflows |
| `ai-loop-paused` | `claude-pr-loop`, `on-slop`, `close-stale-bestaxbot-prs` |
| `deep-review` | `claude-review`, `claude-pr-loop` |

All four gate real behaviour. Deleting or renaming one disables a control **with no workflow diff at all** — the exact hazard the paragraph above that table exists to name, so the row was undercutting its own section. Same for schedules: a `schedule:` on `ai-scan` or the stale sweep decides when a security control runs.

This is the same over-generalization as the earlier `dependabot.yml` advisory, left behind in the same row — I narrowed the ignore and didn't re-read the two words next to it. The real dividing line isn't which file the change lives in, it's **cosmetic vs. load-bearing**, so the row now says that outright:

- Third row is now "genuinely inert config — a cosmetic label, a docs-build cadence, a `semver-major` ignore."
- Middle row absorbs "anything a workflow reads as a gate or trigger."
- Added the check that settles it without judgement: **grep the workflows for what you're changing; if anything reads it, it's middle row.**

`pnpm format:check` clean repo-wide.

allxsmith:
@coderabbitai full review

Issue #490, opened 2026-08-07
bulma-ui: narrow component color unions to CSS-backed values (next major)


allxsmith:
Follow-up to #367, which deprecated color values with no shipped CSS (TSDoc notes + dev-mode console warnings) while keeping every type unchanged. At the next major, make the types tell the truth:

1. **Narrow the per-component unions.** `Progress.color`, `Notification.color` (and `NotificationOptions.color`), and the root `Hero.color` become the CSS-backed set: `'primary' | 'link' | 'info' | 'success' | 'warning' | 'danger' | 'black' | 'white' | 'light' | 'dark'` (plus `'text'`, see item 3). The Hero root also drops `'inherit' | 'current'` — `Hero.Head`/`Hero.Body`/`Hero.Foot` keep theirs, since their `color` routes to the CSS-backed `has-text-*` helpers. Stop typing these props off `(typeof validColors)[number]`; use per-component unions or a shared `ComponentModifierColor` type (precedent: `TagColor` in `elements/Tag.tsx`, `TableColor` in `elements/Td.tsx`).
2. **Remove the `color` prop from `Pagination` and `Tabs`** (deprecated since #367; Bulma has never shipped `.pagination`/`.tabs` color CSS). Remove their `is-${color}` emission too, and update their rows in `skills/bestax-theming/references/themeable-components.md` — once the emission is gone, the `skills-sync` conformance check stops requiring the rows to carry color claims.
3. **Consider adding `'text'` (and evaluate `'white-bis'`/`'white-ter'`) to `validColors`** in `bulma-ui/src/helpers/bulmaClassHelpers.ts` — extending the constant is allowed. `.progress.is-text`/`.notification.is-text`/`.hero.is-text` CSS ships today but is unreachable through the current unions. Verify the `has-text-*`/`has-background-*` helper behavior for each value before widening the helper-prop surface.
4. **bestax-migrate:** add `valueTodo` maps for the removed values on the `color` entries of `Hero`, `Notification`, and `Progress` in `bestax-migrate/src/sources/react-bulma-components/mapping.ts` (the `Button` entry shows the pattern), e.g. `grey: 'no is-grey CSS; use bgColor="grey-light" or a supported modifier'`. Pagination and Tabs have no `color` mapping entries, so nothing to do there. The `kitchen-sink` e2e typecheck will catch any miss.
[...]

PR #491, opened 2026-08-07
fix(bulma-ui): deprecate CSS-less color values, warn in dev, fix has-text fall-through

### LINKED ISSUE #367: [Bug] bulma-ui: color props typecheck values with no shipped CSS; Box color falls through to has-text-*
Part of #363 (skill-loop findings). Two related type-vs-CSS truth gaps, both observed causing agent failures across the 10-run eval:

1. **Color unions are typed wider than the CSS that ships.** `Progress`, `Notification`, and `Hero` accept the full 17-value `validColors` union, but the shipped CSS has no `is-grey*` / `is-*-bis` / `is-*-ter` rules for any component. `<Progress color="grey">` typechecks, emits `is-grey`, and renders default-styled. This produced two drafted-then-caught wobbles in run i07 and the eval's only **shipped** defect in run i10 (the Fable comparison bar silently unstyled — `runs/i10` scorecard on the archive branch). tsc caught every *invented* prop in the eval; it cannot catch these. Narrowing per-component unions to CSS-backed values converts the whole class into compile errors.
2. **`Box`'s `color` prop falls through `useBulmaClasses` to `has-text-<color>`** — it renders text color, never a box variant (there are no `.box.is-*` rules). Verified by live render during the eval (i07 pass). Either destructure and drop the prop, or document the actual behavior — the current shape reads like a filled-variant API.

Evidence: [report.md §remaining gaps](https://github.com/allxsmith/bestax/blob/chore/skill-improvement-loop/experiment/skill-loop/report.md) (library-level items 1 and 3). The skills now document both behaviors (PR #365), but the type-level fix is what makes the failure impossible.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

allxsmith:
Closes #367 (the type-level narrowing itself is deferred to #490 — see below).

Bulma 1.0.4 iterates its 11-color `$colors` map for component modifiers, so several of our color props accept values that emit a class no shipped CSS rule matches. This PR makes every such case visible without breaking anyone, in three moves:

## 1. Dead `is-<color>` values now warn in dev and are documented as deprecated

`Progress`, `Notification` (both the component prop and `NotificationOptions` for the programmatic API), and the `Hero` root accept the 17-value `validColors` union, but only `primary, link, info, success, warning, danger, black, white, light, dark` have shipped CSS for `.progress`/`.notification`/`.hero`. The seven grey/bis/ter values (plus `inherit`/`current` on the Hero root) render unstyled — the exact failure that produced the skill-loop eval's only shipped defect (run i10's silently unstyled comparison bar).

- Each prop's TSDoc now names the CSS-backed values and the dead ones, so the generated docs tables, IDE hover, and anything reading the `.d.ts` see the truth inline.
- A new internal warn-once helper (`bulma-ui/src/helpers/colorDeprecations.ts`, not exported from the package root) logs one development-mode console warning per component+value. Production builds are silent.
- Types are unchanged — removal happens in the next major (#490).

## 2. `Pagination` and `Tabs`: the whole `color` prop is deprecated

Both emit `is-${color}` on their root, and Bulma ships no `.pagination.is-*` or `.tabs.is-*` color CSS at all — every value has always been a silent no-op (the prop is consumed before `useBulmaClasses`, so there is no `has-text-*` fallback either). The prop now carries an `@deprecated` TSDoc tag (rendered as a **Deprecated.** note in the docs tables) and warns once in dev. Emission is kept for now; removal lands with the follow-up major.

## 3. The `has-text-*` fall-through on nine components is now deliberate

`Box`, `Block`, `Content`, `Delete`, `IconText`, `Image`, `Buttons`, `Card`, and `Container` declare a 6-value `color` that was never destructured — it rode the props spread into `useBulmaClasses`, rendered `has-text-<color>`, and (because the spread came last) silently **overrode** `textColor` when both were set. Now:

- `color` is destructured and passed as `textColor ?? color`: identical output when one prop is set, and `textColor` wins when both are (the bug fix).
- TSDoc on each `color` prop states the real behavior: a text-color alias, not a filled variant.

## Ripples

- **Stories:** `Progress`/`Notification` color argTypes trimmed to the 10 live values and every argType given a `description`; both files removed from the stories-conformance `LEGACY_EXEMPT` list. The Pagination `Colors` story (showcasing the dead prop) is deleted; the Tabs `color` argType is removed.
- **Docs:** regenerated props tables pick up the new TSDoc; hand fixes to `valid-values.md` (the "also component `color` props" claim), the color guide (`<Box color="white">` → `textColor`, `<Notification color="white">` → `textColor`, native `<span color>` → `<Span textColor>`), and the Pagination/Tabs overview prose.
- **Skills:** `bestax-theming/references/themeable-components.md` warning block and the Notification/Hero/Progress/Box/Pagination/Tabs rows updated to the deprecated + dev-warn reality; `bestax-layout-scaffold` Hero row annotated.
- **Tests:** new helper suite for the warn-once/dev-gating logic; per-component cases assert the class still renders, the warning fires exactly once, live values stay silent, production stays silent, `notification.show()` is covered, `Hero.Body` (a text helper) does not warn, and the nine fall-through components render `has-text-*` with `textColor` taking precedence.

Non-breaking throughout: no type changed, no prop removed, no emission changed. The only behavior change is the `textColor`-precedence bug fix in item 3. Release-wise this is a `fix(bulma-ui)` patch.

## Summary by CodeRabbit

- **New Features**
  - `color` now consistently works as a text-color alias across supported components, with `textColor` taking precedence.
- **Bug Fixes**
  - Added development warnings for deprecated or unsupported color values and color props without visual CSS effects.
- **Documentation**
  - Clarified color behavior, supported values, deprecations, and recommended `textColor`/`bgColor` alternatives.
  - Updated Storybook controls and examples to reflect current styling behavior.
- **Tests**
  - Added coverage for color aliases, precedence rules, warning behavior, and production warning suppression.


allxsmith:
Addressed all review findings in 117c0c0:

- **`isDev()` fail-open in browser bundles** (Copilot, CodeRabbit, deep review): the helper now reads a bare `process.env.NODE_ENV` inside `try/catch`, so bundlers replace it statically (dev serve warns, production builds compile the check away) and a runtime with no `process` at all (raw CDN ESM) lands in the `catch` and stays silent — fail closed, warnings can never fire in production. The missing-`process` test now asserts silence, and an explicit `NODE_ENV=development` case was added.
- **`color.md` scoped its color claims** (CodeRabbit): the Text Color section now teaches `textColor` (tables renamed), explains where `color` means the filled `is-<color>` variant instead, and the example uses `textColor` on `Title`/`SubTitle` (which never had a `color` prop) and `Button`. The Background section teaches `bgColor`; the `Box`/`Card` examples use `bgColor`, and the `Notification` example uses its `color` variant since `Notification` exposes no background prop at all.
- **Hero row in the layout skill** (CodeRabbit): now lists the exact CSS-backed union and the full accepted-but-deprecated set including `inherit`/`current`.
- **`black-bis`/`black-ter` aren't greys** (CodeRabbit): the three inventory rows name them explicitly, the Hero row and the warning block also call out `inherit`/`current` on the Hero root.

PR #512, opened 2026-08-13
docs: settle the bestax-migrate bundling call, create-bestax ships it

allxsmith:
## What

Aligns the last docs with the call made in #385: the `bestax-migrate` skill **is** part of the create-bestax bundle. Four files, prose only:

- `docs/docs/skills/intro.md`: the Migrate bullet loses its "Not bundled by `create-bestax`" sentence.
- `docs/docs/skills/migrate.mdx`: the "Unlike the other skills, this one is **not** bundled" paragraph becomes the bundled framing (preinstalled, idle until legacy imports show up).
- `bestax-migrate/CLAUDE.md`: the "contested, see #385" note becomes the settled policy, with the keep-in-agreement list (`sync-skills.mjs`, the `CLAUDE_MD` roster, the docs skills pages).
- `create-bestax/CLAUDE.md`: the per-skill bundling rule now cites #385 as settled instead of open.

## Why

The code has shipped the skill since #345 (`sync-skills.mjs` allowlist, the scaffolded `CLAUDE_MD` roster, `setupSkills()`), while two docs pages still said it was deliberately not bundled. #385 asked for a direction decision; the decision is recorded on the issue: one uniform bundle, no per-skill carve-out to keep in sync, and the skill costs a fresh scaffold nothing.

## Not in this PR

- No behavior changes: `sync-skills.mjs` and `setupSkills()` are untouched.
- `pnpm gen` produced no diff (the MCP index and skill catalog are unchanged), so there is no regeneration commit.

## Testing

- `pnpm all` green locally. The first cold run tripped the known pre-existing `bestax-mcp` sync-skills race (`ENOTEMPTY` on `data/skills/`); the constituent tasks pass sequentially, as documented in #508.
- `git grep -ni "not bundled"` over `*.md`/`*.mdx`: no skill-related hits remain.

Closes #385

## Summary by CodeRabbit

* **Documentation**
  * Clarified that the migration skill is bundled with new `create-bestax` projects.
  * Documented that migration guidance remains inactive unless legacy imports are detected.
  * Updated release and synchronization guidance to keep related tooling and documentation aligned.

PR #513, opened 2026-08-13
chore: keep local agent worktrees out of prettier's sweep

allxsmith:
## What

Adds `.claude/worktrees/` to the root `.prettierignore`.

## Why

Agent sessions leave git worktrees (full checkouts of other branches) under `.claude/worktrees/`, and the root `format:check` glob walks into them. A worktree's copy of `bestax-migrate/fixtures/` no longer matches that root-anchored exemption, so `pnpm all` fails locally on files this repo already declines to format, on branches whose own PR CI format-checks them from a clean tree. Surfaced while running the gate for #512.

Deliberately narrow: `.claude/commands/` is tracked, hand-written markdown and stays formatted. CI behavior is unchanged (checkouts there never contain local worktrees); this only makes the local gate truthful.

## Testing

- With three stale agent worktrees present, `pnpm run format:check` went from 9 failures (all under `.claude/worktrees/`) to clean.
- The tracked `.claude/commands/*.md` files are still checked (verified they match the glob and are not excluded).

## Summary by CodeRabbit

* **Chores**
  * Excluded local agent worktree files from automated formatting checks.
  * Documented the formatting exclusion for improved development workflow consistency.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -26,3 +26,11 @@ bestax-migrate/.e2e-tmp/
 # read off these files. Same argument as the codemod fixtures above. The
 # harness's own scripts (eval/agent-loop/bin, lib) are ours and stay formatted.
 eval/agent-loop/runs-*/*/app-src/
+
+# Local agent worktrees: full checkouts of other branches that agent sessions
+# leave under .claude/. Each branch is format-checked by its own PR's CI; seen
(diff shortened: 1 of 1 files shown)


allxsmith:
@coderabbitai review

PR #531, opened 2026-08-18
test: hold pack-manifest and check-conformance in agreement

### LINKED ISSUE #435: [Refactor] No test holds pack-manifest.mjs and check-conformance.mjs in agreement
## Affected Package(s)

- [x] Other: repo-root `scripts/` + `bestax-migrate/scripts/`

---

## Area(s) of Code

- `scripts/check-conformance.mjs` — `checkPublishableManifests()`, `UNRESOLVABLE_AT_PACK`
- `bestax-migrate/scripts/pack-manifest.mjs` — `resolveSpecifier()` and its two `process.exit(1)` branches

---

## Motivation

These two files encode the same rule twice and must not drift.

`pack-manifest.mjs` decides which pnpm specifier shapes it will resolve at pack time. `checkPublishableManifests()` decides which shapes it will let past CI on the strength of the pack hooks being wired. Any shape the script **refuses** but the check **excuses** is a silent CI-green / release-red inversion — the check reports all clear and the release is what breaks. That is the exact opposite of why the check exists.

That inversion happened twice during review of #417, both times in code that looked obviously correct:

1. **`catalog:`** — flagged by CodeRabbit, fixed in `4127ead`. Reproduced live at head before fixing:
   ```
   catalog: in devDeps + hooks correctly wired
     check-conformance --only=publishable-manifests  ->  exit 0   (passes)
     pack-manifest.mjs prepack                       ->  exit 1   (hard fails)
   ```
2. **pnpm's alias form `workspace:<name>@<range>`** — found while re-reading `resolveSpecifier` after the first fix, fixed in `de6a900`. It was unwrapped to a bare `@allxsmith/bestax-bulma@^5`, which no package manager can install, and exited 0 — so it would have shipped.

Nothing prevents a third. The deep review on #417 traced all seven documented pnpm rewrite shapes and found the two files now agree shape-for-shape, but explicitly named the absence of a test as the top remaining risk:

> The riskiest part is no longer a shape drift but the absence of a test freezing that agreement — that is why a third instance keeps being "assumable."

Worth noting the whole custom script surface is untested: 1,638 lines across six scripts, zero tests, and three of them run during `npm publish`.

---

## Proposed Changes

A table-driven test over the shape matrix, asserting the invariant directly: **the script refuses a shape if and only if the check flags it.**

Matrix (all currently handled):

| specifier | script | check |
|---|---|---|
| `workspace:*` | resolve → exact version | excused with hooks |
| `workspace:^` / `workspace:~` | resolve → prefixed range | excused with hooks |
| `workspace:` (bare) | resolve → exact version | excused with hooks |
| `workspace:^5.0.0` (explicit range) | unwrap | excused with hooks |
| `workspace:<name>@<range>` (alias) | **exit 1** | **flag** |
| `catalog:` / `catalog:<name>` | **exit 1** | **flag** |

The test should drive both real implementations rather than reimplementing their logic — the bug class is precisely "two implementations disagree," so a third copy in the test defeats the purpose. `pack-manifest.mjs` needs a seam for this; it currently reads `package.json` from a path derived from `import.meta.url`.

Include the negative cases too. `workspace:^5.0.0` must **not** trip the alias predicate (`rest.includes('/') || rest.lastIndexOf('@') > 0`) — a false positive there breaks the actual fix from #412.

---

## Risks or Drawbacks

`pack-manifest.mjs` currently has no injection point for its package root, so it needs a small refactor to be testable. That is a real change to a script that runs during publish, so it should land with the test in the same PR and be verified with a real `npm pack` before merge.

Do this **before** the resolver replacement — the test is what makes that swap safe to attempt.

---

## Additional Context

Follow-up to #417 (fixes #412). Advisory 1 of 3 from the second deep review on that PR.

allxsmith:
Closes #435.

Two files encode the same rule about which pnpm specifier shapes survive `npm publish`:

- `bestax-migrate/scripts/pack-manifest.mjs` decides what it rewrites at pack time.
- `checkPublishableManifests()` in `scripts/check-conformance.mjs` decides what it lets past CI, on the strength of those pack hooks being wired.

**A shape the script REFUSES but the check EXCUSES is a green CI with a red release** — the check reports all clear and the release is what breaks, which is the exact inversion the check exists to prevent.

That inversion shipped twice during review of #417, both times in code that read as obviously correct: `catalog:` (fixed in `4127ead`) and pnpm's alias form `workspace:<name>@<range>` (fixed in `de6a900`, found unwrapping to a bare `@scope/pkg@^5` that no package manager can install — #412 wearing a different hat). Nothing prevented a third, and the deep review on #417 named the absence of a test as the top remaining risk.

## The test drives both real implementations

A third copy of the rule in a test would be one more thing to drift, which is the bug class rather than a fix for it. `scripts/pack-manifest.test.mjs` imports the actual resolver and the actual conformance predicate and asserts the invariant as a **biconditional** — the script refuses a shape if and only if the check flags it — across all twelve documented shapes:

| specifier | expected |
| --- | --- |
| `workspace:*`, `workspace:` (bare) | resolve → exact version |
| `workspace:^`, `workspace:~` | resolve → prefixed range |
| `workspace:^5.0.0`, `workspace:~5.0.0`, `workspace:5.0.0`, `workspace:>=5 <6` | unwrap |
| `workspace:<name>@<range>` (alias, scoped and unscoped) | **refused by both** |
| `catalog:`, `catalog:<name>` | **refused by both** |

Including the negative cases the issue called out: `workspace:^5.0.0` and friends must **not** trip the alias predicate, or the fix for #412 starts rejecting the very shape it was written to support. Both arms of that predicate are exercised separately, so neither can be dropped.

## The seams

Both files needed one, and both use the pattern already established in this repo (`check-security-txt-expiry.mjs`, `gen-mcp-index.mjs`, `verify-attestation.mjs`).

**`pack-manifest.mjs`** derived its package root from `import.meta.url` and called `process.exit` inside the decision logic, so the decision could not be called without packing something. Now:

- The per-specifier decision is one exported function, `rewriteSpecifier`, taking its version lookup as an argument and throwing `UnsupportedSpecifierError` rather than exiting.
- The `catalog:` refusal moved out of the prepack loop into that same function, so **"refused or resolved" lives in one place instead of two** — which is a small reduction in the very duplication this issue is about.
- `main()` is exported and only runs when the file is executed directly.

**`check-conformance.mjs`** exports `UNRESOLVABLE_AT_PACK` and `unresolvableAtPack`, and guards `main()` — importing a module should not run a repo-wide conformance sweep as a side effect.

## Verified by mutation, not by a green run

A test that passes proves nothing unless it fails when the invariant breaks. Each inversion that actually shipped was reintroduced and the suite re-run:

| mutation | result |
| --- | --- |
| check excuses `catalog:` (`4127ead` reverted) | 2 failures |
| check excuses the alias form (`de6a900` reverted) | 3 failures |
| script false-positives on `^` ranges (breaks #412's fix) | 3 failures |
| script stops refusing `catalog:` (the opposite drift direction) | 3 failures |
| bare `workspace:` stops resolving | 1 failure |
| **a refusal returns 0 instead of 1** | 3 failures |
| postpack stops restoring the manifest | 1 failure |

One mutation is **not** caught, and it is an equivalent mutation rather than a gap: widening `lastIndexOf('@') > 0` to `includes('@')` cannot change any real shape's verdict, because an explicit range holds no `@` for the widened predicate to see. Recorded here rather than quietly omitted.

## The publish path, exercised for real

The issue flags this explicitly: `pack-manifest.mjs` runs during `npm publish`, so a refactor of it has to be proven, not reasoned about.

`npm pack` end to end:

```
packed manifest devDependency : ^5.11.1
workspace:/catalog: in tarball : none
repo manifest after postpack   : restored byte for byte
stale .pack-backup             : none
```

**The exit codes matter as much as the rewrite**, because npm reads them to abort a publish — if a refusal returned 0, npm would publish the broken tarball this whole guard exists to prevent, the bug inverted. Those paths are frozen too:

```
refused catalog: specifier  -> exit 1, manifest untouched, no backup written
refused alias form          -> exit 1, manifest untouched, no backup written
unresolvable workspace dep  -> exit 1, manifest untouched, no backup written
prepack with stale backup   -> exit 1 (refuses to clobber)
postpack with no backup     -> exit 0 (not an error)
unknown / missing mode      -> exit 1
```

The "no backup written" assertions are load-bearing in their own right: a stale `.pack-backup` makes the *next* prepack refuse to run, so a half-done refusal would wedge the following release too.

## Verification

```
233 script tests pass (62 new), all 5 conformance checks pass, lint and format clean,
full `pnpm all` green
```

Deep review's advisory 2 is addressed in `cf95eb7`: the biconditional now also runs over ~29 unlisted specifiers — plausible future pnpm shapes, adjacent protocols neither side owns, degenerate strings — asserting only that the two sides *agree*, not what the verdict should be. Verified the same way as everything else: adding `|| rest === 'latest'` to one side only turns the suite red, where before it passed.

## Note on scope

The issue asks for this **before** the resolver replacement it also contemplates — the test is what makes that swap safe to attempt. Nothing here changes which shapes are accepted or refused; the behaviour is identical, and the mutation table is the evidence for that.

**A correction to a figure quoted from the issue.** #435 describes the custom script surface as "1,638 lines across six scripts, zero tests, and three of them run during `npm publish`". I repeated that here before checking it, and all three numbers are now out of date:

- The surface is **~30,000 lines across 92 script files** — `scripts/`, all four packages, `docs/` and `eval/` — not 1,638 across six. The issue's figure was presumably repo-root `scripts/` when it was written.
- **Four** scripts run during `npm publish`, not three: `pack-manifest.mjs` and `bulma-ui/scripts/pack-pointer-files.mjs` (both `prepack` + `postpack`), plus `sync-skills.mjs` in `create-bestax` and in `bestax-mcp` (`prepack`).
- "Zero tests" no longer holds — there are nine script test files, several added recently (`check-pointer-urls`, `verify-attestation`, `check-security-txt-expiry`, `auto-close-duplicates`, `bypass-annotations`, `gen-mcp-index`, `props-extract`, `package-manager-translate`), and this PR adds the tenth.

The motivation is unchanged and the direction of the number does not matter to it: `pack-manifest.mjs` is publish-critical and had no test until now. But the figure should not be repeated as-is.

### DIFF
--- a/bestax-migrate/scripts/pack-manifest.mjs
+++ b/bestax-migrate/scripts/pack-manifest.mjs
@@ -29,14 +29,25 @@
  * risk npm not running `postpack` and leaving the repo manifest rewritten for
  * @semantic-release/git to commit — the exact failure the backup below exists
  * to prevent.
+ *
+ * The per-specifier decision is exported and takes its version lookup as an
+ * argument, and `main` only runs when this file is executed directly. That
+ * seam exists for one reason (#435): this script and
+ * scripts/check-conformance.mjs encode the same rule about which pnpm shapes
+ * are resolvable, and a shape this script REFUSES but the check EXCUSES is a
+ * green CI with a red release. That inversion happened twice during review of
+ * #417 — once for `catalog:`, once for the alias form — so a test now drives
+ * both real implementations and asserts they agree shape for shape. It can
+ * only do that if the decision is callable without packing anything.
  */
 import fs from 'node:fs';
 import path from 'node:path';
-import { fileURLToPath } from 'node:url';
+import process from 'node:process';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 
-const pkgRoot = path.dirname(path.dirname(fileURLToPath(import.meta.url)));
-const manifest = path.join(pkgRoot, 'package.json');
-const backup = path.join(pkgRoot, 'package.json.pack-backup');
+const defaultPkgRoot = path.dirname(
+  path.dirname(fileURLToPath(import.meta.url))
+);
 
 const DEP_SECTIONS = [
   'dependencies',
@@ -45,23 +56,41 @@ const DEP_SECTIONS = [
   'optionalDependencies',
 ];
 
-/** Version of a workspace package, read through the linked node_modules entry. */
-function resolveWorkspaceVersion(name) {
-  const linked = path.join(pkgRoot, 'node_modules', name, 'package.json');
-  if (!fs.existsSync(linked)) {
-    console.error(
-      `pack-manifest: cannot resolve the workspace dependency "${name}" — ` +
-        `${path.relative(pkgRoot, linked)} does not exist.\n` +
-        'Run `pnpm install` from the repo root before packing.'
-    );
-    process.exit(1);
+/**
+ * Raised for a specifier this script will not resolve. Thrown rather than
+ * exited so the decision can be exercised by a test; `main` turns it back into
+ * the same message-and-exit-1 the CLI has always produced.
+ */
+export class UnsupportedSpecifierError extends Error {
+  constructor(message) {
+    super(message);
+    this.name = 'UnsupportedSpecifierError';
   }
-  const { version } = JSON.parse(fs.readFileSync(linked, 'utf8'));
-  if (!version) {
-    console.error(`pack-manifest: "${name}" has no version in its manifest`);
-    process.exit(1);
-  }
-  return version;
+}
+
+/**
+ * Version of a workspace package, read through the linked node_modules entry.
+ * Curried over the package root so a test can supply its own lookup instead of
+ * needing a real pnpm-linked tree.
+ */
+export function makeWorkspaceVersionResolver(pkgRoot = defaultPkgRoot) {
+  return name => {
+    const linked = path.join(pkgRoot, 'node_modules', name, 'package.json');
+    if (!fs.existsSync(linked)) {
+      throw new UnsupportedSpecifierError(
+        `pack-manifest: cannot resolve the workspace dependency "${name}" — ` +
+          `${path.relative(pkgRoot, linked)} does not exist.\n` +
+          'Run `pnpm install` from the repo root before packing.'
+      );
+    }
+    const { version } = JSON.parse(fs.readFileSync(linked, 'utf8'));
+    if (!version) {
+      throw new UnsupportedSpecifierError(
+        `pack-manifest: "${name}" has no version in its manifest`
+      );
+    }
+    return version;
+  };
 }
 
 /**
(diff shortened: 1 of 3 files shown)


allxsmith:
All three advisories acknowledged; one is now closed in `cf95eb7`.

**Advisory 2 — the enumerated-shape gap.** This was the useful one and it is fixed. You are right that agreement held for the twelve listed shapes *by assertion* and for everything else only because the two predicates are currently textually identical — so a one-sided edit touching an unlisted shape would slip through, which is this file's own bug class landing just outside where it was looking.

The biconditional now also runs over ~29 unlisted specifiers: plausible future pnpm shapes (`workspace:latest`, `workspace:1.x`, `workspace:^1.0.0-beta.1`, `workspace:@scope/name` with no range), adjacent protocols neither side owns (`npm:`, `file:`, `link:`, `git+https:`, `jsr:`), and degenerate strings (empty, whitespace, wrong case, 200 chars).

It deliberately asserts only that the two sides **agree**, not what the verdict should be. Deciding the right answer for a hypothetical future pnpm syntax is not this test's job; noticing that the two files stopped answering it the same way is.

Verified the same way as the rest of the table — adding `|| rest === 'latest'` to one predicate only, a shape absent from the twelve, turns the suite red where it previously passed.

**Advisory 1 — the uncaught mutation.** Agreed, and thank you for checking it rather than taking the PR body's word for it. Equivalent, not a gap: an explicit range holds no `@` for the widened predicate to see.

**Advisory 3 — you could not execute the suite.** Worth recording what that means for the rest of your review: the mutation table and the `npm pack` results were produced by running them, not by tracing, so where your static analysis and those results agree you have independent confirmation. CI's `node --test "scripts/*.test.mjs"` is wired via the root `test` script and is green on this branch.

Also corrected in the PR body while here: I had quoted the issue's "1,638 lines across six scripts, zero tests, three run during npm publish" without checking it. All three figures are stale — the surface is ~30,000 lines across 92 script files, **four** scripts run during publish (`pack-manifest.mjs` and `pack-pointer-files.mjs` for prepack+postpack, plus `sync-skills.mjs` in create-bestax and bestax-mcp), and nine script test files already exist. The motivation is unchanged, but the figure should not be repeated as-is.

PR #556, opened 2026-08-24
ci: make the AGENTS.md roster scope fence-aware, closing the class

allxsmith:
## Description

The #548 deep review named the `bulma-ui/AGENTS.md` roster scope as **the one remaining extractor** in the hardened class that was not fence-aware: a fenced example quoting `Agent skills (…)` before the real parenthetical steals the match and scopes the check to the quote. No current trigger exists (the phrase occurs once in the file), which is why it was advisory-grade — closed here the same way every sibling was, with the decoy as a fixture.

## Review-round accounting (verified against main, not assumed)

- The #546 Copilot finding (protocol rule advising the plain-range spelling the sibling rule forbids) is **already fixed on main** — the consumer branch skips sibling names citing that review. My initial probe double-reported only because it omitted the third argument the walk passes.
- The #545 CodeRabbit findings (quote parity, trim'd compare) landed via the pre-merge reconciliation commits.
- This was the only surviving finding from the full deep-review + Copilot + CodeRabbit round across #544/#545/#546/#548.

Refs #548

## Checklist
- [x] Fixture proves the decoy no longer steals the anchor
- [x] All 17 conformance checks and full script suite green

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of agent skill rosters by ignoring matching text inside fenced examples.
  * Prevented quoted or illustrative content from being mistaken for the active roster.

* **Tests**
  * Added coverage confirming fenced examples do not affect roster validation results.

### DIFF
--- a/scripts/check-conformance.mjs
+++ b/scripts/check-conformance.mjs
@@ -2236,8 +2236,25 @@ export const SKILL_ROSTERS = [
   {
     file: 'bulma-ui/AGENTS.md',
     // This one ships in the npm tarball, so a stale roster here is
-    // consumer-facing rather than internal.
-    scope: text => text.match(/Agent skills \(([^)]*)\)/)?.[1] ?? null,
+    // consumer-facing rather than internal. Fence-aware like every other
+    // scope — the #548 deep review named this as the one straggler in the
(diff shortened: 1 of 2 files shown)


allxsmith:
@coderabbitai full review

PR #558, opened 2026-08-25
fix(bestax-migrate): only convert Button remove to Delete on a true literal

### LINKED ISSUE #553: [Bug] bestax-migrate: `<Button remove={false}>` and `remove={expr}` silently become `<Delete>`
## Affected Package(s)

---

## Description

Same root cause as the `Heading` issue, different component. The `button` special (`bestax-migrate/src/sources/react-bulma-components/specials.ts:207-214`) keys on attribute **presence**:

```ts
/** `<Button remove />` is Bulma's delete cross → bestax `<Delete />`. */
button(ctx, _path, element) {
  const attr = findAttr(element, 'remove');
  if (!attr) return {};
  removeAttr(element, attr);
  ctx.dirty = true;
  return { target: 'Delete' };
},
```

So an ordinary button written `<Button remove={false}>Save</Button>` — or, worse, a conditional one written `<Button remove={isRemove}>Save</Button>` — is rewritten into a Bulma delete cross. No TODO, no report entry.

---

## Reproduction Steps

2. Create `probe/b.jsx`:

```jsx
import React from 'react';
import { Button } from 'react-bulma-components';

export function P({ isRemove }) {
  return (
    <div>
      <Button remove />
      <Button remove={false}>Save</Button>
      <Button remove={isRemove}>Save</Button>
    </div>
  );
}
```

3. `node bestax-migrate/dist/index.js react-bulma-components probe/b.jsx --no-deps`

---

## Expected Behavior

| Input | Expected |
| --- | --- |
| `<Button remove />` / `remove={true}` | `<Delete />` |
| `<Button remove={false}>Save</Button>` | `<Button>Save</Button>` (prop dropped, nothing else changes) |
| `<Button remove={expr}>Save</Button>` | `TODO(bestax-migrate)` explaining the branch has to be split by hand; do not pick a target silently |

---

## Actual Behavior

Reported as `1 file(s) scanned, 1 transformed, **0 TODO(s) left**`:

```jsx
import { Delete } from "@allxsmith/bestax-bulma";

<Delete />              {/* ✅ */}
<Delete>Save</Delete>   {/* ❌ was remove={false} — a normal button became a delete cross */}
<Delete>Save</Delete>   {/* ❌ was remove={isRemove} — silent best-guess on a dynamic value */}
```

Note the second and third also carry children into `<Delete>`, which Bulma's delete element does not render meaningfully — so the button's label disappears from the UI.

---

## Environment

- Input library: react-bulma-components 4.x
- Node: 22.22.2, OS: linux

---

## Additional Context

Companion to the `Heading` prop-presence issue — both are the same defect shape and are the only two specials that skip `literalValueOf`. `columns` (`specials.ts:176-204`), `panel-tab` (`specials.ts:629`), `breadcrumb-item` (`specials.ts:665`) and `alignTarget` (`specials.ts:44-74`) all handle literal `false` and dynamic expressions correctly, so the fix can follow those.

A shared "boolean prop selects a target" helper would close both issues at once; if they're taken separately, land the helper with whichever goes first.

---

## Definition of Done

- [ ] Fix + regression tests (failing before, passing after) — `__testfixtures__` pair for the literal cases, the dynamic form in `fixtures/kitchen-sink/src/leftovers.tsx` asserted via TODO rules

bestaxbot:
## Summary

The `button` special in `bestax-migrate/src/sources/react-bulma-components/specials.ts` keyed on **attribute presence** rather than its value: any `<Button remove={...}>` — including `remove={false}` or a dynamic `remove={isRemove}` — was unconditionally rewritten into `<Delete />`, silently dropping the button's children with no TODO or report entry.

This resolves the attribute's literal value with the existing `literalValueOf` helper (the same pattern already used by `columns`, `panel-tab`, `breadcrumb-item`, and `alignTarget`):

| Input | Now produces |
| --- | --- |
| `<Button remove />` / `remove={true}` | `<Delete />` (unchanged) |
| `<Button remove={false}>Save</Button>` | `<Button>Save</Button>` — prop dropped, nothing else changes |
| `<Button remove={expr}>Save</Button>` | left as-is with a `TODO(bestax-migrate)` — the branch has to be split by hand, no silent guess |

## Test plan

- [x] New fixture pair `__testfixtures__/button-remove.{input,output}.tsx` covering all three cases
- [x] Unit tests in `transform.test.ts` asserting the `false` case drops the prop and the dynamic case emits a `prop:remove` TODO without becoming `<Delete>`
- [x] Added the dynamic case to `fixtures/kitchen-sink/src/leftovers.tsx` and asserted the `prop:remove` TODO rule in `e2e/kitchen-sink.test.ts`
- [x] `pnpm --filter bestax-migrate run test:coverage` (227 tests passing, coverage thresholds hold)
- [x] `pnpm run format:check`

Fixes #553

### DIFF
--- a/bestax-migrate/src/sources/react-bulma-components/specials.ts
+++ b/bestax-migrate/src/sources/react-bulma-components/specials.ts
@@ -204,13 +204,28 @@ const SPECIALS: Record<string, SpecialHandler> = {
     return { handledProps: ['multiline'] };
   },
 
-  /** `<Button remove />` is Bulma's delete cross → bestax `<Delete />`. */
-  button(ctx, _path, element) {
+  /**
+   * `<Button remove />` (or a literal `remove={true}`) is Bulma's delete
+   * cross → bestax `<Delete />`. A literal `false` just drops the prop; a
+   * dynamic value can render as either one depending on runtime state, so
+   * it's a TODO instead of a silent guess.
+   */
+  button(ctx, path, element) {
     const attr = findAttr(element, 'remove');
     if (!attr) return {};
-    removeAttr(element, attr);
-    ctx.dirty = true;
-    return { target: 'Delete' };
+    const literal = literalValueOf(attr);
+    if (literal.kind === 'boolean') {
+      removeAttr(element, attr);
+      ctx.dirty = true;
+      return literal.value ? { target: 'Delete' } : {};
+    }
+    addTodo(
+      ctx,
+      path,
+      'prop:remove',
+      '`remove` has a dynamic value; this can render as either a Button or a Delete cross — split the branch by hand'
+    );
+    return {};
   },
 
   /** Heading → Title / SubTitle / plain `.heading` paragraph. */
--- a/bestax-migrate/src/sources/react-bulma-components/specials.ts
+++ b/bestax-migrate/src/sources/react-bulma-components/specials.ts
@@ -205,16 +205,19 @@ const SPECIALS: Record<string, SpecialHandler> = {
   },
 
   /**
-   * `<Button remove />` (or a literal `remove={true}`) is Bulma's delete
-   * cross → bestax `<Delete />`. A literal `false` just drops the prop; a
-   * dynamic value can render as either one depending on runtime state, so
+   * `<Button remove />` (or any truthy literal) is Bulma's delete
+   * cross → bestax `<Delete />`. A falsy literal just drops the prop; only a
+   * genuine expression can render as either one depending on runtime state, so
    * it's a TODO instead of a silent guess.
    */
   button(ctx, path, element) {
     const attr = findAttr(element, 'remove');
     if (!attr) return {};
     const literal = literalValueOf(attr);
-    if (literal.kind === 'boolean') {
+    if (literal.kind !== 'expression') {
+      // A statically-known value always renders the same at runtime: truthy
+      // (`remove`, `remove={true}`, `remove="true"`) → `<Delete />`; falsy
+      // (`remove={false}`, `remove=""`, `remove={0}`) → just drop the prop.
       removeAttr(element, attr);
       ctx.dirty = true;
       return literal.value ? { target: 'Delete' } : {};
--- a/bestax-mcp/data/skills.json
+++ b/bestax-mcp/data/skills.json
@@ -88,7 +88,7 @@
         {
           "id": "component-map",
           "file": "references/component-map.md",
-          "bytes": 11894
+          "bytes": 13188
         },
         {
           "id": "css-migration",
--- a/bestax-migrate/e2e/kitchen-sink.test.ts
+++ b/bestax-migrate/e2e/kitchen-sink.test.ts
@@ -174,6 +174,7 @@ describe('kitchen-sink e2e', () => {
     expect(rules).toContain('prop:colorVariant');
     expect(rules).toContain('prop:value');
     expect(rules).toContain('prop:delta');
+    expect(rules).toContain('prop:remove');
     expect(todos.length).toBeGreaterThanOrEqual(10);
     const migrated = fs.readFileSync(
       path.join(tmpDir, 'src', 'leftovers.tsx'),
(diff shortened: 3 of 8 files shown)


bestax-migrate/src/sources/react-bulma-components/specials.ts
```diff
@@ -204,13 +204,28 @@ const SPECIALS: Record<string, SpecialHandler> = {
...
+   * it's a TODO instead of a silent guess.
+   */
+  button(ctx, path, element) {
     const attr = findAttr(element, 'remove');
     if (!attr) return {};
-    removeAttr(element, attr);
-    ctx.dirty = true;
-    return { target: 'Delete' };
+    const literal = literalValueOf(attr);
+    if (literal.kind === 'boolean') {
```
allxsmith: This gate mislabels static literals as dynamic: `literalValueOf` returns kind `'string'` for `remove="true"`, so a value that always renders the delete cross at RBC runtime gets the '`remove` has a dynamic value' TODO — factually wrong — and the attr survives onto the bestax `<Button>`, which has no `remove` mapping, so it leaks to the DOM. On main this converted to `Delete` correctly, so it's a small regression, not just a miss (uncommon input, mostly untyped JS codebases, but real). Statically-known strings/numbers should take the literal path: truthy → Delete, falsy (`remove=""`, `remove={0}`) → drop; only genuine expressions get the dynamic TODO.

allxsmith (changes requested):
One correctness blocker (inline) and one doc-sync blocker.

**Doc sync**: `skills/bestax-migrate/references/component-map.md` line 14 still says Button `remove` is 'Auto; `<Button remove/>` → `<Delete/>`' — dynamic values are now Flagged, and sibling PR 559 updated this exact table for Heading. `bestax-migrate/CLAUDE.md` requires the skill references to move in the same PR.

Non-blocking, for a follow-up rather than this PR: this is now one of three inline copies of the literal-true/false/dynamic ladder (with 559's two) that `props.ts`'s `booleanToProp` branch already centralizes, and the copies are already drifting — this one keeps the attr on the dynamic path while 559's remove it. A shared `booleanTarget` helper would keep the convention in one place. Same presence-vs-value class is also still live in the untouched `field` handler (`multiline={false}` / `multiline={expr}` silently become `grouped="multiline"`) — worth its own issue.

bestax-migrate/src/sources/react-bulma-components/specials.ts
bestaxbot: Good catch — fixed in 62084bf. You're right: `literalValueOf` returns `'string'`/`'number'` for `remove="true"` and `remove={0}`, and the handler only took the literal path for `'boolean'`, so a statically-known value fell through to the dynamic TODO and left the unmapped `remove` attr on the emitted `<Button>`.

The gate is now `literal.kind !== 'expression'`: any statically-known value resolves by truthiness — truthy (`remove`, `remove={true}`, `remove="true"`) → `<Delete/>`, falsy (`remove={false}`, `remove=""`, `remove={0}`) → drop the prop. Only a genuine expression still gets the `prop:remove` TODO. Added a unit test asserting the string/number path (no TODO, no leaked attr) and extended the fixture pair with the `"true"`/`""`/`{0}` cases.

allxsmith:
@coderabbitai review

bestax-migrate/src/sources/react-bulma-components/specials.ts
allxsmith: Verified 62084bf: the gate is now kind !== 'expression' with truthiness resolution, literalValueOf carries value for string and number literals (checked — remove={1} converts, remove={0} drops), the fixture covers remove="true", and the component-map row states by-value semantics. All 228 tests pass locally. Resolving.

allxsmith (approved):
Verified: static literals resolve by truthiness (regression from main closed), literalValueOf carries values for all literal kinds, fixture covers the reported input, and the component-map row documents by-value semantics. 228 tests pass locally.

PR #559, opened 2026-08-25
fix(bestax-migrate): pick Heading target by prop value, not presence

### LINKED ISSUE #552: [Bug] bestax-migrate: Heading special keys on prop presence, not value — `subtitle={false}` and dynamic values silently produce the wrong component
## Affected Package(s)

---

## Description

The `heading` special (`bestax-migrate/src/sources/react-bulma-components/specials.ts:217-253`) decides its target by asking whether the `subtitle` / `heading` **attribute exists**, never what its value is:

```ts
const subtitleAttr = findAttr(element, 'subtitle');
if (subtitleAttr) {
  removeAttr(element, subtitleAttr);
  ctx.dirty = true;
  return { target: 'SubTitle' };
}
return { target: 'Title' };
```

So `subtitle={false}` and `subtitle={someExpression}` both map to `SubTitle`, and `heading={false}` / `heading={expr}` both become a plain `<p className="heading">`. No TODO is emitted, the prop is deleted, and the run reports success — this is a silent, wrong rewrite of a dynamic value, which the package's own hard rule forbids ("Anything unsafe gets a `// TODO(bestax-migrate): <hint>` comment … never a silent skip, never a best-guess rewrite of dynamic values").

RBC v4 semantics, from the published package (`react-bulma-components@4.1.0`, `esm/components/heading/heading.js`):

```js
title: !subtitle && !heading,
subtitle: subtitle,
```

`subtitle={false}` is a **Title**. The codemod turns it into a SubTitle.

---

## Reproduction Steps

2. Create `probe/h.jsx`:

```jsx
import React from 'react';
import { Heading } from 'react-bulma-components';

export function Probe({ isSub, useHeadingStyle }) {
  return (
    <div>
      <Heading subtitle>a</Heading>
      <Heading subtitle={true}>b</Heading>
      <Heading subtitle={false}>c</Heading>
      <Heading subtitle={isSub}>d</Heading>
      <Heading heading={false}>e</Heading>
      <Heading heading={useHeadingStyle}>f</Heading>
      <Heading size={3}>g</Heading>
    </div>
  );
}
```

3. `node bestax-migrate/dist/index.js react-bulma-components probe --no-deps`

---

## Expected Behavior

| Input | Expected |
| --- | --- |
| `<Heading subtitle>` / `subtitle={true}` | `<SubTitle>` |
| `<Heading subtitle={false}>` | `<Title>` |
| `<Heading heading={false}>` | `<Title>` |
| `<Heading subtitle={expr}>` / `heading={expr}` | a `TODO(bestax-migrate)` + a conservative target, prop preserved or explained — never a silent pick |

`alignTarget` (`specials.ts:44-74`) is the model already in the codebase: string literal → mapped target, non-literal → `addTodo(…, 'dynamic value; pick between X / Y by hand')`. `columns` (`specials.ts:176-204`), `panel-tab` (`specials.ts:629`) and `breadcrumb-item` (`specials.ts:665`) all inspect `literalValueOf` correctly. `heading` is the outlier (see also the companion `button` / `remove` issue).

---

## Actual Behavior

Reported as `1 file(s) scanned, 1 transformed, **0 TODO(s) left**`, output:

```jsx
import { SubTitle, Title } from "@allxsmith/bestax-bulma";

<SubTitle>a</SubTitle>              {/* ✅ */}
<SubTitle>b</SubTitle>              {/* ✅ */}
<SubTitle>c</SubTitle>              {/* ❌ was subtitle={false} → should be Title */}
<SubTitle>d</SubTitle>              {/* ❌ was subtitle={isSub} → should be a TODO */}
<p className="heading">e</p>        {/* ❌ was heading={false} → should be Title */}
<p className="heading">f</p>        {/* ❌ was heading={useHeadingStyle} → should be a TODO */}
<Title size="3">g</Title>           {/* ✅ */}
```

This is visible in a real migrated app: a page heading written `<Heading size={2} subtitle={false}>Legacy Dashboard</Heading>` renders as `.subtitle` after migration, so the page's `h1` is visually demoted with no warning.

---

## Environment

- Input library: react-bulma-components 4.1.0 (semantics read from the published tarball as text; never installed)
- Node: 22.22.2, OS: linux

---

## Additional Context

Found while validating a migration of a legacy RBC v4 + React 17 app. The migration is otherwise sound — the migrated app bundles and renders correctly against real Bulma v1 CSS; these prop-presence cases are the exception.

Suggested shape of the fix: give `heading` the same literal-inspection treatment as `alignTarget`, i.e. resolve `literalValueOf(attr)`, branch on `kind === 'boolean'`, and `addTodo` with rule `prop:subtitle` / `prop:heading` for the expression case. A shared helper for "boolean prop that selects a target" would cover this issue and the `button` / `remove` one together.

---

## Definition of Done

- [ ] Fix + regression tests (failing before, passing after) — `__testfixtures__` pair for the literal cases, plus the dynamic forms in `fixtures/kitchen-sink/src/leftovers.tsx` asserted via TODO rules
- [ ] `skills/bestax-migrate/references/` updated in the same PR if the mapping guidance changes

bestaxbot:
## Summary

The `heading` special (`bestax-migrate/src/sources/react-bulma-components/specials.ts`) picked its rewrite target by checking whether the `subtitle` / `heading` **attribute was present**, never what its literal value was. That meant:

- `<Heading subtitle={false}>` silently became `<SubTitle>` instead of `<Title>`
- `<Heading heading={false}>` silently became a plain `<p className="heading">` instead of `<Title>`
- `<Heading subtitle={expr}>` / `<Heading heading={expr}>` (dynamic values) were also silently mapped without any `TODO(bestax-migrate)`, which the package's own hard rule forbids.

RBC v4's actual semantics (`esm/components/heading/heading.js`) are `title: !subtitle && !heading`, `subtitle: subtitle` — the *value* picks the target.

### Fix

`heading`'s special handler now inspects `literalValueOf(attr)` (the same pattern already used by `alignTarget`, `columns`, `panel-tab`, and `breadcrumb-item`):

- literal `true` → same structural rewrite as before (`SubTitle` / plain `<p className="heading">`)
- literal `false` → treated as if the prop were absent (falls through to `Title` unless `subtitle` says otherwise)
- dynamic (non-literal) → leaves a `prop:subtitle` / `prop:heading` `TODO(bestax-migrate)` comment and picks a conservative fallback (`Title`/`SubTitle` — never the structural plain-element rewrite, since that can't be undone by hand as easily)

### Tests

- New fixture pair `__testfixtures__/heading-literals.{input,output}.tsx` covering the literal `subtitle={true|false}` / `heading={false}` cases — confirmed this fails against the pre-fix code (`b`/`c` cases mis-transform) and passes after the fix.
- New unit tests in `transform.test.ts` asserting the `prop:subtitle` / `prop:heading` TODO rules fire for dynamic values and that the fallback target is `Title`.
- `fixtures/kitchen-sink/src/leftovers.tsx` gained two `Heading` dynamic-value cases; `e2e/kitchen-sink.test.ts` now asserts both TODO rules appear.
- Updated `skills/bestax-migrate/references/component-map.md`'s `Heading` row to describe value-based (not presence-based) selection.

Fixes #552

## Test plan

- [x] `pnpm --filter bestax-migrate run test:coverage` (227 tests passed; 96.53% stmts / 87.35% branches, above the 95%/78% threshold)
- [x] `pnpm run format:check`
- [x] Confirmed the new fixture test fails against the pre-fix `specials.ts` and passes after the fix (regression-test gate)

🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

- **New Features**
  - Improved migration of Bulma heading components based on `heading` and `subtitle` prop values.
  - Static values now produce appropriate title, subtitle, or plain heading output.
  - Dynamic or unsupported values retain TODO annotations for review.

- **Documentation**
  - Updated migration guidance and component mappings to describe heading conversion behavior.

- **Tests**
  - Added coverage for static, dynamic, truthy, and falsy heading properties.

### DIFF
--- a/bestax-migrate/src/sources/react-bulma-components/specials.ts
+++ b/bestax-migrate/src/sources/react-bulma-components/specials.ts
@@ -282,13 +282,16 @@ const SPECIALS: Record<string, SpecialHandler> = {
 
     // A truthy literal `heading` collapses to a plain `.heading` paragraph —
     // but only when `subtitle` isn't a dynamic value the collapse would drop.
+    // A truthy *literal* `subtitle` is safe to collapse: RBC applies its class
+    // independently, so it rides along on the paragraph as `heading subtitle`
+    // (dropping it would be the same silent loss the dynamic guard prevents).
     if (headingTruthy && !subtitleDynamic) {
       removeAttr(element, headingAttr);
       const className = mergeClassName(
         ctx,
         path,
         element,
-        'heading',
+        subtitleTruthy ? 'heading subtitle' : 'heading',
         'Heading'
       );
       const rest = stripModifierProps(
--- a/bestax-migrate/src/sources/react-bulma-components/specials.ts
+++ b/bestax-migrate/src/sources/react-bulma-components/specials.ts
@@ -216,11 +216,16 @@ const SPECIALS: Record<string, SpecialHandler> = {
   /**
    * Heading → Title / SubTitle / plain `.heading` paragraph, matching RBC's
    * `title: !subtitle && !heading` / `subtitle: subtitle` — the *value* of
-   * each prop picks the target, not merely whether it was passed. A literal
-   * `false` is equivalent to the prop being absent; a dynamic value can't be
-   * resolved at codemod time, so it's left in place with a TODO and a
-   * conservative (Title/SubTitle, never the structural plain-element rewrite)
-   * fallback so the branch can be split by hand. The subtitle class is applied
+   * each prop picks the target, not merely whether it was passed. Any
+   * statically-known literal resolves to a fixed truthiness at codemod time
+   * (a bare prop and `heading={true}`/`heading="heading"` are all truthy;
+   * `heading={false}`/`heading=""`/`heading={0}` are falsy and equivalent to
+   * the prop being absent), exactly as RBC evaluates the prop at runtime —
+   * this mirrors the Button `remove` handler so the two don't contradict each
+   * other on string/number literals. Only a genuine expression can't be
+   * resolved: it's left in place with a TODO and a conservative
+   * (Title/SubTitle, never the structural plain-element rewrite) fallback so
+   * the branch can be split by hand. The subtitle class is applied
    * independently of heading in RBC, so a dynamic `subtitle` blocks the
    * structural `heading` collapse — collapsing would silently drop it.
    */
@@ -229,8 +234,14 @@ const SPECIALS: Record<string, SpecialHandler> = {
     const subtitleAttr = findAttr(element, 'subtitle');
     const headingLit = headingAttr ? literalValueOf(headingAttr) : undefined;
     const subtitleLit = subtitleAttr ? literalValueOf(subtitleAttr) : undefined;
-    const headingDynamic = !!headingLit && headingLit.kind !== 'boolean';
-    const subtitleDynamic = !!subtitleLit && subtitleLit.kind !== 'boolean';
+    const headingDynamic = headingLit?.kind === 'expression';
+    const subtitleDynamic = subtitleLit?.kind === 'expression';
+    // A resolvable literal (boolean/string/number) collapses to its runtime
+    // truthiness; an expression stays dynamic.
+    const headingTruthy =
+      !!headingLit && headingLit.kind !== 'expression' && !!headingLit.value;
+    const subtitleTruthy =
+      !!subtitleLit && subtitleLit.kind !== 'expression' && !!subtitleLit.value;
 
     // Dynamic values can't be resolved to a target: leave the prop on the
     // element (so its expression survives for hand-splitting) plus a TODO.
@@ -251,13 +262,9 @@ const SPECIALS: Record<string, SpecialHandler> = {
       );
     }
 
-    // Literal-true `heading` collapses to a plain `.heading` paragraph — but
-    // only when `subtitle` isn't a dynamic value the collapse would drop.
-    if (
-      headingLit?.kind === 'boolean' &&
-      headingLit.value === true &&
-      !subtitleDynamic
-    ) {
+    // A truthy literal `heading` collapses to a plain `.heading` paragraph —
+    // but only when `subtitle` isn't a dynamic value the collapse would drop.
+    if (headingTruthy && !subtitleDynamic) {
       removeAttr(element, headingAttr);
       const className = mergeClassName(
         ctx,
@@ -286,18 +293,18 @@ const SPECIALS: Record<string, SpecialHandler> = {
       return { replaced: true };
     }
 
-    // A resolvable literal `heading` (`false`, or `true` blocked by a dynamic
-    // subtitle) behaves as if the prop were absent — drop it. A dynamic
-    // `heading` stays put (its TODO is already filed above).
-    if (headingLit?.kind === 'boolean') {
+    // A resolvable literal `heading` that didn't collapse (falsy, or truthy
+    // but blocked by a dynamic subtitle) behaves as if the prop were absent —
+    // drop it. A dynamic `heading` stays put (its TODO is already filed above).
+    if (headingLit && !headingDynamic) {
       removeAttr(element, headingAttr);
       ctx.dirty = true;
     }
 
-    if (subtitleLit?.kind === 'boolean') {
+    if (subtitleLit && !subtitleDynamic) {
       removeAttr(element, subtitleAttr);
       ctx.dirty = true;
-      return { target: subtitleLit.value ? 'SubTitle' : 'Title' };
+      return { target: subtitleTruthy ? 'SubTitle' : 'Title' };
     }
     // Dynamic subtitle (attr kept) or no subtitle → conservative Title.
     return { target: 'Title' };
--- a/bestax-migrate/src/sources/react-bulma-components/specials.ts
+++ b/bestax-migrate/src/sources/react-bulma-components/specials.ts
@@ -213,42 +213,77 @@ const SPECIALS: Record<string, SpecialHandler> = {
     return { target: 'Delete' };
   },
 
-  /** Heading → Title / SubTitle / plain `.heading` paragraph. */
+  /**
+   * Heading → Title / SubTitle / plain `.heading` paragraph, matching RBC's
+   * `title: !subtitle && !heading` / `subtitle: subtitle` — the *value* of
+   * each prop picks the target, not merely whether it was passed. A literal
+   * `false` is equivalent to the prop being absent; a dynamic value can't be
+   * resolved at codemod time, so it's left as a TODO with a conservative
+   * (Title/SubTitle, never the structural plain-element rewrite) fallback.
+   */
   heading(ctx, path, element) {
     const headingAttr = findAttr(element, 'heading');
     if (headingAttr) {
-      removeAttr(element, headingAttr);
-      const className = mergeClassName(
-        ctx,
-        path,
-        element,
-        'heading',
-        'Heading'
-      );
-      const rest = stripModifierProps(
-        ctx,
-        path,
-        attributesOf(element).filter(
-          a => !['size', 'weight', 'spaced', 'subtitle'].includes(a.name.name)
-        ),
-        'Heading'
-      );
-      const replacement = plainElement(
-        ctx.j,
-        'p',
-        className,
-        rest,
-        element.children ?? []
-      );
-      path.replace(replacement);
-      ctx.dirty = true;
-      return { replaced: true };
+      const literal = literalValueOf(headingAttr);
+      if (literal.kind === 'boolean' && literal.value === true) {
+        removeAttr(element, headingAttr);
+        const className = mergeClassName(
+          ctx,
+          path,
+          element,
+          'heading',
+          'Heading'
+        );
+        const rest = stripModifierProps(
+          ctx,
+          path,
+          attributesOf(element).filter(
+            a => !['size', 'weight', 'spaced', 'subtitle'].includes(a.name.name)
+          ),
+          'Heading'
+        );
+        const replacement = plainElement(
+          ctx.j,
+          'p',
+          className,
+          rest,
+          element.children ?? []
+        );
+        path.replace(replacement);
+        ctx.dirty = true;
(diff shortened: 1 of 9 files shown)


bestax-migrate/src/sources/react-bulma-components/specials.ts
```diff
@@ -213,42 +213,77 @@ const SPECIALS: Record<string, SpecialHandler> = {
...
+        ctx.dirty = true;
+        return { target: literal.value ? 'SubTitle' : 'Title' };
+      }
+      addTodo(
+        ctx,
+        path,
+        'prop:subtitle',
+        '`subtitle` has a dynamic value; pick between Title / SubTitle by hand'
+      );
       removeAttr(element, subtitleAttr);
```
allxsmith, line 326: The dynamic branches call `removeAttr`, which permanently deletes the user's expression while the TODO asks them to use it: `<Heading subtitle={choice === 'a'}>x</Heading>` becomes `<Title>x</Title>` plus a comment, and `choice === 'a'` now exists only in git history — with several Headings in one file the comments can't even be matched back to their elements, and runtime styling silently degrades to always-Title. Sibling PR 558's button handler deliberately KEEPS `remove={expr}` on the element in the identical situation; keep the attr here too so the branch can actually be split by hand.

allxsmith (changes requested):
Two correctness blockers — one inline, one on the review thread (`<Heading heading subtitle={isSub}>` returns before the `prop:subtitle` check) — plus a doc-sync blocker.

**Doc sync**: `docs/docs/guides/getting-started/migration/react-bulma-components.md` line 46 still states 'Heading subtitle → SubTitle' unconditionally — it's now Title for `subtitle={false}` and a TODO for dynamic values. The skill table was updated but the migration guide moves in the same PR too (`bestax-migrate/CLAUDE.md`).

Also note: `heading="heading"` (a statically truthy string literal) takes the dynamic branch, same gate shape as 558's `remove="true"` — whatever fix lands there should land here.

bestax-migrate/src/sources/react-bulma-components/specials.ts
```diff
@@ -213,42 +213,77 @@ const SPECIALS: Record<string, SpecialHandler> = {
...
-        'p',
-        className,
-        rest,
-        element.children ?? []
-      );
-      path.replace(replacement);
-      ctx.dirty = true;
-      return { replaced: true };
+      const literal = literalValueOf(headingAttr);
+      if (literal.kind === 'boolean' && literal.value === true) {
```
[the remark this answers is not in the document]
allxsmith: Verified — with `heading` literal-true, the attribute filter drops `subtitle` and the branch returns `{ replaced: true }` before the new `prop:subtitle` check runs, so `<Heading heading subtitle={isSub}>` collapses to `<p className="heading">` with no TODO. RBC applies the subtitle class independently of heading (both truthy renders `className="heading subtitle"`), so the drop isn't behavior-preserving either. It's a pre-existing hole on main, but this PR's premise is that dynamic values get flagged — the subtitle check needs to run before the structural return.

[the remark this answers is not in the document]
bestaxbot: Fixed in 8a076c0. The handler now resolves both props up front and only takes the structural collapse when `heading` is literal-true **and** `subtitle` is not dynamic. So `<Heading heading subtitle={isSub}>` no longer returns `{ replaced: true }` before the subtitle check — instead it leaves the `subtitle={isSub}` expression in place, files a `prop:subtitle` TODO, and falls back to `<Title>` (non-structural) rather than collapsing to `<p className="heading">`. Added a unit test that asserts the output has no `className="heading"`, keeps `subtitle={isSub}`, and emits the TODO.

bestax-migrate/src/sources/react-bulma-components/specials.ts
bestaxbot: Fixed in 8a076c0. The dynamic branches no longer call `removeAttr` — a non-literal `subtitle` or `heading` stays on the element (e.g. `<Title subtitle={choice === 'a'}>`), so the expression survives for hand-splitting and isn't stranded in git history. This mirrors #558's button handler keeping `remove={expr}`. Only resolvable literals are dropped (`false`, or a literal-true `heading` when it actually collapses). Updated the two dynamic unit tests to assert the expression is preserved.

allxsmith:
@coderabbitai review

bestax-migrate/src/sources/react-bulma-components/specials.ts
allxsmith: 8a076c0 fixes both halves of this thread (dynamic props preserved on the element; the dynamic subtitle now blocks the heading collapse) — verified, 228 tests pass. One residual before this converges: the gate here is still kind !== 'boolean', so heading="heading" and subtitle="yes" — statically truthy string literals — are mislabeled dynamic and get the false 'dynamic value' TODO, where main resolved them correctly. Sibling PR 558 (62084bf) just switched the identical ladder to kind !== 'expression' with truthiness resolution; please align heading/subtitle the same way so the two handlers don't contradict each other on string literals.

bestaxbot: Fixed in f063086 — aligned the heading/subtitle ladder with #558's button handler.

The gate was `kind === 'boolean'` for the literal path, so `literalValueOf` returning `'string'`/`'number'` for `heading="heading"` or `subtitle="yes"` fell through to the dynamic `prop:*` TODO — a regression against main, which resolved those. Now it keys on `kind !== 'expression'` and collapses every statically-known literal to its runtime truthiness:

- truthy `heading` (`heading`, `heading={true}`, `heading="heading"`) → plain `.heading` paragraph, same as before for the bare form
- truthy `subtitle` (`subtitle`, `subtitle={true}`, `subtitle="yes"`) → `SubTitle`
- falsy literal (`{false}`, `""`, `{0}`) → treated as absent, drops the prop
- only a genuine expression still stays on the element with the `prop:subtitle`/`prop:heading` TODO

Both handlers now agree on string/number literals. Added two unit tests asserting `heading="heading"`/`subtitle="yes"` resolve with no dynamic TODO (`className="heading"` / `<SubTitle>` respectively). Gates green: bestax-migrate lint / typecheck / test:coverage (230 passed, 96.53% stmts / 87.55% branches, above 95%/78%), format:check, check:conformance, gen:mcp:check, gen:catalog:check.

The docs and skill component-map already read 'literal → Auto, dynamic value → flagged', which this only makes more accurate, so no doc change was needed.

allxsmith (approved):
Verified through both rounds: dynamic props are preserved on the element with their TODOs, a dynamic subtitle blocks the structural heading collapse, and f063086 aligns the literal gate with 558's Button convention — every statically-known literal resolves by truthiness, mirroring RBC's runtime evaluation, so the two handlers agree. Tests pass locally across the suite.

allxsmith (approved):
Re-approved after the main merge-back: conflicts were unions (both PRs' tests kept, component-map table carries both the new Button and Heading rows), skills.json restamped via pnpm gen:mcp. 239 tests, lint, typecheck, format, skills-sync all green locally.

allxsmith (commented):
No defects on the current head — the truthiness convention is consistent through the whole handler now, and the merge unions check out. One design question before this merges, since this PR owns two of the three copies:

`button` (merged via #558), `heading`, and `subtitle` now each inline the same literal-resolution ladder — resolve every non-expression literal by truthiness, keep the attr + TODO for a genuine expression. `props.ts`'s `booleanToProp` branch is a fourth sibling of the same shape. The copies already drifted once (that's how the `kind !== 'boolean'` / `kind !== 'expression'` split happened), and `panel-tab`/`breadcrumb-item` carry the older `active` ladders too. Is it worth extracting a shared helper (`resolveBooleanish(attr)` returning truthy/falsy/expression) in this PR while both handlers are open — or is that a follow-up issue once this lands? Either answer works; I want the decision recorded rather than a fourth drift.

bestax-migrate/src/sources/react-bulma-components/specials.ts
```diff
@@ -231,10 +231,58 @@ const SPECIALS: Record<string, SpecialHandler> = {
...
+        ctx,
+        path,
+        'prop:heading',
+        '`heading` has a dynamic value; when truthy it renders a plain `<p className="heading">` instead of Title/SubTitle — resolve by hand'
+      );
+    }
+
+    // A truthy literal `heading` collapses to a plain `.heading` paragraph —
+    // but only when `subtitle` isn't a dynamic value the collapse would drop.
+    if (headingTruthy && !subtitleDynamic) {
```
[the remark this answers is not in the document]
bestaxbot: Fixed in 85c4e91. When both props are truthy literals the element still collapses to the plain paragraph, but the base class is now `subtitleTruthy ? 'heading subtitle' : 'heading'`, so `<Heading heading subtitle>x</Heading>` becomes `<p className="heading subtitle">x</p>` instead of dropping the class. This matches RBC applying the subtitle class independently of heading, and it makes the literal path consistent with the guard that already blocks the collapse for a *dynamic* subtitle. Added a unit test asserting `className="heading subtitle"` with no TODO, plus a both-truthy case in the heading-literals fixture pair. Gates green: bestax-migrate lint / typecheck / test:coverage (240 passed, 96.66% stmts / 87.98% branches) and format:check.

allxsmith:
Word given — file the tracking issue with your site inventory, targeting one atomic PR that introduces resolveBooleanish in jsx-utils.ts and rewires all five sites under one test, including the booleanToProp alignment. Add a sixth site to the inventory while you're there: the field handler resolves multiline by presence — findAttr only, value never inspected — so multiline={false} and multiline={expr} both silently become grouped="multiline" (specials.ts, ~line 436). Same presence-vs-value class this PR family fixed; it belongs in the same sweep.

allxsmith (approved):
Verified 85c4e91: the literal-both collapse now carries the subtitle class (className="heading subtitle"), matching RBC's independent application of the class — fixture and unit tested, 240 tests green locally. The resolveBooleanish decision is recorded: follow-up issue for one atomic PR across all six sites (including booleanToProp and field's multiline). Approving.

PR #560, opened 2026-08-25
fix(bestax-migrate): stop mislabeling RBC's own stylesheet as a third-party extension

### LINKED ISSUE #555: [Bug] bestax-migrate: react-bulma-components' own stylesheet imports in SCSS are mislabeled as a third-party `bulma-components` extension and left in place
## Affected Package(s)

---

## Description

`EXTENSION_IMPORT` (`bestax-migrate/src/sources/react-bulma-components/styles.ts:73-74`) detects 0.9-era add-on packages:

```ts
const EXTENSION_IMPORT =
  /^\s*@import\s+['"][^'"]*\bbulma-([\w-]+?)(?:\/|\.|['"])/;
```

`\bbulma-` matches **inside** `react-bulma-components` (the `-` before `bulma` is a word boundary), so the source library's own stylesheet is captured as an extension named `bulma-components`, left in place, and annotated with advice to keep it:

```scss
// TODO(bestax-migrate): bulma-components is a Bulma 0.9-era extension — check its Bulma v1 compatibility; the bestax extras already style Radio/Checkbox and the advanced form controls — see …
@import 'react-bulma-components/src/index.sass';
```

Two problems compound:

1. **The advice is wrong.** This isn't a third-party extension, it's the library being migrated away from. The report entry reads `third-party Bulma extension \`bulma-components\` left in place; verify it against Bulma v1 or replace it with the bestax extras`.
2. **The result doesn't build.** `deps.ts` removes `react-bulma-components` from `package.json` in the same run, so after the user runs their install the stylesheet imports a package that is no longer there:

```
Error: Can't find stylesheet to import.
  ╷
5 │ @import 'react-bulma-components/src/index.sass';
```

The JS side already handles the equivalent case — `import 'react-bulma-components/dist/react-bulma-components.min.css'` is rewritten to the bestax bundle (covered by `src/sources/react-bulma-components/__tests__/transform.test.ts:349`). Only the Sass path misses it.

---

## Reproduction Steps

2. Create `probe/a.scss`, `probe/b.scss`, `probe/c.scss` with, respectively:

```scss
@import 'react-bulma-components/src/index.sass';
@import '~react-bulma-components/src/index.sass';
@import 'react-bulma-components/dist/react-bulma-components.min.css';
```

3. `node bestax-migrate/dist/index.js react-bulma-components probe --no-deps`

All three are reported as `third-party Bulma extension \`bulma-components\``. A genuine extension (`@import 'bulma-checkradio/dist/css/bulma-checkradio.sass';`) is correctly reported as `bulma-checkradio`, so the regex is right for its intended input — it just over-matches.

---

## Expected Behavior

Treat the source library's own stylesheet like the JS side does — replace it per `--css` mode (`bestax` → the bestax bundle / extras, `bulma` → `bulma/sass`, `keep` → leave with an accurate TODO) — and in every case do not describe it as a third-party extension to be verified against Bulma v1. Excluding the source library's package name from `EXTENSION_IMPORT` (e.g. requiring the match to start at the specifier's package boundary) is the minimum fix; rewriting is the useful one.

---

## Actual Behavior

Mislabeled TODO + report entry as above, import left untouched, and a Sass build that fails once the user installs, since the package was just removed from the manifest.

---

## Environment

- Node: 22.22.2, OS: linux

---

## Additional Context

Scope check, from the published tarballs (read as text, never installed):

- **react-bulma-components 4.1.0 ships no stylesheets at all** — no `.css`, `.sass` or `.scss` files in the package. So a pure v4 app can't have these imports and is unaffected.
- **react-bulma-components 3.4.0 ships both** `dist/react-bulma-components.min.css` and `src/index.sass`, and its README documents exactly these two lines: `import 'react-bulma-components/dist/react-bulma-components.min.css';` for basic setup and `@import "~react-bulma-components/src/index.sass"` for the "Advanced setup" path.

So this hits apps that came up on v3 and later moved to v4 while keeping their stylesheet wiring — which is a large part of the "old project on an unmaintained library" population the codemod targets. That is also why the JS-side already handles the v3 CSS path defensively; the Sass side should match.

Found while migrating a legacy RBC v4 + React 17 app end to end; the rest of that migration bundles and renders correctly.

---

## Definition of Done

- [ ] Fix + regression tests covering bare, `~`-prefixed and `dist/*.min.css` forms in SCSS, plus a genuine `bulma-*` extension still detected
- [ ] Migrated stylesheet compiles with `sass` (see the `with (…)` comma issue — same test hook if it lands first)
- [ ] `skills/bestax-migrate/references/` + the docs migration guide updated in the same PR if the stylesheet guidance changes

bestaxbot:
## Summary

`EXTENSION_IMPORT` in `bestax-migrate/src/sources/react-bulma-components/styles.ts` matched `\bbulma-` anywhere in the specifier — including inside `react-bulma-components`, since the hyphen before "bulma" counts as a word boundary. That mislabeled the source library's own stylesheet import (`react-bulma-components/src/index.sass`, `~`-prefixed, or the bundled v3 CSS) as a third-party `bulma-components` extension and left it untouched. Since `deps.ts` removes `react-bulma-components` from `package.json` in the same run, the leftover import fails to resolve once the user installs (`Error: Can't find stylesheet to import.`).

**Fix:**

- `EXTENSION_IMPORT` now requires the `bulma-*` package name to start at a specifier segment boundary (right after the quote/`~`, or after a `/`), so it no longer matches a name that merely *contains* `bulma-` mid-word.
- A new `RBC_STYLE_IMPORT` recognizer detects the library's own stylesheet (bare/`~`-prefixed `src/index.sass`, or the bundled `dist/react-bulma-components(.min).css`) and replaces it per `--css` mode instead of ever calling it a third-party extension:
  - `bestax` (default): `@use '@allxsmith/bestax-bulma/scss/bestax';`
  - `bulma`: plain `@use 'bulma/sass';` (no extras, matching existing `bulma`-mode behavior elsewhere in this file)
  - `keep`: still replaces it with `@use 'bulma/sass';` plus an explanatory TODO — the import can never resolve post-migration, so `keep` mode fixes it the same way the existing JS-side CSS pass already treats RBC's dead v3 bundled CSS under `--css keep`, rather than leaving a guaranteed-broken import in place.
  - If the file already has its own `bulma/…` root import elsewhere, the now-redundant RBC stylesheet line is dropped instead of importing Bulma twice.
- Updated `skills/bestax-migrate/references/css-migration.md` and the docs migration guide with the corrected guidance, and regenerated `bestax-mcp/data` (byte count changed).

## Regression test

Added a failing→passing reproduction: confirmed the new tests fail against the pre-fix code (mislabeled as `bulma-components`, left in place) for exactly the reported reason, then pass against the fix. New coverage in `styles.test.ts` includes bare/`~`-prefixed/`dist/*.min.css` forms across all three `--css` modes, the redundant-import-drop case, and a check that a genuine `bulma-*` extension is still correctly detected alongside the RBC stylesheet.

Fixes #555

## Test plan

- [x] `pnpm --filter bestax-migrate run test:coverage` (234/234 passing, coverage 96.5%/87.4% — above the 95%/78% threshold)
- [x] `pnpm run format:check`
- [x] Confirmed the new tests fail against the pre-fix code for the reported reason, then pass against the fix

## Summary by CodeRabbit

- **New Features**
  - Improved migration support for `react-bulma-components` stylesheet imports, including v3 entry points and deep partials.
  - Preserved Sass variable overrides and prevented duplicate Bulma imports.
  - Added clear TODO guidance for unsupported deep component imports.

- **Bug Fixes**
  - Reduced false positives when detecting third-party Bulma extensions.
  - Ensured dependency-update settings are honored during stylesheet transformations.

- **Documentation**
  - Expanded migration guidance for stylesheet handling, supported imports, overrides, and TODO reporting.

### DIFF
--- a/bestax-migrate/src/sources/react-bulma-components/styles.ts
+++ b/bestax-migrate/src/sources/react-bulma-components/styles.ts
@@ -91,6 +91,19 @@ const RBC_STYLE_IMPORT =
   /^(\s*)@import\s+(['"])((?:\.\.?\/)+node_modules\/|~)?react-bulma-components\/[^'"]*\2\s*;?\s*$/;
 
 /**
+ * The subset of RBC stylesheet imports that are the library's own *root*: the
+ * bare package specifier, the documented Sass entry point (`src/index`), and
+ * the bundled v3 CSS (`dist/react-bulma-components(.min).css`). These pull in
+ * the whole library, which `bulma/sass` supersedes — so when the file already
+ * has a Bulma root, dropping one of these loses nothing. A deep RBC partial
+ * (`src/components/navbar.sass`) is NOT a root: `bulma/sass` doesn't carry a
+ * given partial's styles, so removing one silently would drop CSS. We flag
+ * those instead (the package's "never a silent skip" rule).
+ */
+const RBC_ROOT_STYLESHEET =
+  /^\s*@import\s+(['"])(?:(?:\.\.?\/)+node_modules\/|~)?react-bulma-components(?:\/src\/index(?:\.s[ac]ss)?|\/dist\/react-bulma-components(?:\.min)?\.css)?\1\s*;?\s*$/;
+
+/**
  * A Bulma module root already pulled in via `@use` — the file's own
  * `@use 'bulma/sass'` (any prefix, configured or not) or the bestax bundle
  * (`scss/bestax`, which itself loads `bulma/sass with (…)`). When one is
@@ -405,21 +418,48 @@ export const transformStyles: StylesTransform = (
               ? "replaced react-bulma-components's stylesheet import with bulma/sass; the package is removed from dependencies, so the old import would not resolve"
               : "replaced react-bulma-components's stylesheet import with bulma/sass; it is a Bulma 0.9 asset the migrated v1 components can't use (--no-deps left the package in place)"
           );
+        } else {
+          // bestax/bulma modes rewrite the file's root without a TODO — still
+          // report it so the migration summary reflects the restructured root.
+          report(
+            collector,
+            filePath,
+            i + 1,
+            'sass',
+            "replaced react-bulma-components's stylesheet import with a Bulma v1 @use root"
+          );
         }
-      } else if (cssMode === 'bestax' && !extrasAdded) {
-        // A Bulma root already exists in this file — a `bulma/…` @import we
-        // convert, or the file's own `@use` root. Keep it and add only the
-        // extras: never a second root, and never the configured bundle that
-        // would reconfigure an already-loaded module.
-        out.push(
-          prefix
-            ? `${indent}@use '${prefix}@allxsmith/bestax-bulma/src/scss/extras';`
-            : `${indent}${EXTRAS_USE}`
-        );
-        extrasAdded = true;
+      } else {
+        // A Bulma root already covers this file (a `bulma/…` @import we
+        // convert, the file's own `@use` root, or an earlier RBC line we
+        // rewrote). In bestax mode keep that root and add only the extras —
+        // never a second root, never the configured bundle that would
+        // reconfigure an already-loaded module.
+        if (cssMode === 'bestax' && !extrasAdded) {
+          out.push(
+            prefix
+              ? `${indent}@use '${prefix}@allxsmith/bestax-bulma/src/scss/extras';`
+              : `${indent}${EXTRAS_USE}`
+          );
+          extrasAdded = true;
+        }
+        // Dropping the RBC line is only safe for its root/index stylesheet,
+        // which `bulma/sass` supersedes. A deep RBC partial can carry styles
+        // `bulma/sass` doesn't, so removing it silently would lose CSS — flag
+        // it with a TODO + report entry instead (never a silent skip).
+        if (!RBC_ROOT_STYLESHEET.test(line)) {
+          out.push(
+            `${indent}// ${TODO}: dropped a react-bulma-components stylesheet partial that has no Bulma v1 root equivalent; port any styles it carried beyond Bulma's own by hand — see ${GUIDE}`
+          );
+          report(
+            collector,
+            filePath,
+            i + 1,
+            'sass',
+            'react-bulma-components stylesheet partial dropped; port any styles it carried beyond Bulma’s own by hand'
+          );
+        }
       }
-      // Otherwise (bulma/keep with a root already present) the line is
-      // redundant — drop it rather than importing Bulma twice.
       changed = true;
       continue;
     }
--- a/bestax-migrate/src/sources/react-bulma-components/styles.ts
+++ b/bestax-migrate/src/sources/react-bulma-components/styles.ts
@@ -68,10 +68,23 @@ const OTHER_BULMA_IMPORT = /^\s*@import\s+['"][^'"]*bulma[^'"]*['"]/;
 /**
  * Third-party Bulma extension packages (`bulma-checkradio`, `bulma-switch`,
  * …) — 0.9-era add-ons whose v1 compatibility varies; several are covered
- * by the bestax extras.
+ * by the bestax extras. The `bulma-` package name must start at a specifier
+ * segment boundary (right after the quote/`~`, or after a `/`) so this does
+ * not also match a name that merely *contains* `bulma-`, like
+ * `react-bulma-components` (see EXTENSION_IMPORT's own regression test).
  */
 const EXTENSION_IMPORT =
-  /^\s*@import\s+['"][^'"]*\bbulma-([\w-]+?)(?:\/|\.|['"])/;
+  /^\s*@import\s+['"](?:~|[^'"]*\/)?bulma-([\w-]+?)(?:\/|\.|['"])/;
+
+/**
+ * The source library's own stylesheet (RBC v3's two documented setup
+ * paths — the bundled CSS and the "advanced" Sass entry point, bare or
+ * `~`-prefixed). This is not a third-party extension: it is the library
+ * being migrated away from, and `deps.ts` removes its package.json entry
+ * in the same run, so leaving the import in place breaks the Sass build.
+ */
+const RBC_STYLE_IMPORT =
+  /^(\s*)@import\s+(['"])(~)?react-bulma-components\/(?:src\/index\.(?:sass|scss)|dist\/react-bulma-components(?:\.min)?\.css)\2\s*;?\s*$/;
 
 /** Any `$name: value;` declaration (with or without `!default`). */
 const VAR_DECL = /^\s*\$([\w-]+)\s*:\s*(.+?)\s*(!default)?\s*;\s*$/;
@@ -121,6 +134,7 @@ export const transformStyles: StylesTransform = (
   const out: string[] = [];
   let changed = false;
   let extrasAdded = source.includes('@allxsmith/bestax-bulma/scss');
+  let rbcStyleEmitted = false;
 
   // Pass 1: find safe leading variable overrides and the first root import.
   const rootImportIndex = lines.findIndex(line => ROOT_IMPORT.test(line));
@@ -234,6 +248,43 @@ export const transformStyles: StylesTransform = (
         'sass',
         `Bulma 0.9 partial path \`bulma/sass/${importPath}\` has no direct v1 equivalent`
       );
+      changed = true;
+      continue;
+    }
+
+    const rbcStyle = line.match(RBC_STYLE_IMPORT);
+    if (rbcStyle && !line.includes(TODO)) {
+      const indent = rbcStyle[1];
+      if (rootImportIndex === -1 && !rbcStyleEmitted) {
+        // The library's own stylesheet never resolves post-migration —
+        // deps.ts removes react-bulma-components from package.json in the
+        // same run — so every mode replaces it with a real Bulma entry
+        // point rather than leaving a guaranteed-broken import; this
+        // mirrors how the JS-side CSS pass treats RBC's dead v3 bundled
+        // CSS even under `--css keep`.
+        if (cssMode === 'bestax') {
+          out.push(`${indent}@use '@allxsmith/bestax-bulma/scss/bestax';`);
+          extrasAdded = true; // the bundle already includes the extras
+        } else {
+          out.push(`${indent}@use 'bulma/sass';`);
+        }
+        if (cssMode === 'keep') {
+          out.push(
+            `${indent}// ${TODO}: replaced react-bulma-components's own stylesheet with @use 'bulma/sass' — its package.json entry is removed, so the old import no longer resolves; install bulma@^1 — see ${GUIDE}`
+          );
+          report(
+            collector,
+            filePath,
+            i + 1,
+            'sass',
+            "replaced react-bulma-components's stylesheet import with bulma/sass; the package is removed from dependencies, so the old import would not resolve"
+          );
+        }
+        rbcStyleEmitted = true;
+      }
+      // Otherwise a real `bulma/…` root import already brings in Bulma (and,
+      // in bestax mode, the extras) elsewhere in this file — drop the now-
+      // redundant line entirely rather than importing Bulma twice.
       changed = true;
       continue;
     }
--- a/bestax-migrate/src/sources/react-bulma-components/styles.ts
+++ b/bestax-migrate/src/sources/react-bulma-components/styles.ts
@@ -77,14 +77,28 @@ const EXTENSION_IMPORT =
   /^\s*@import\s+['"](?:~|[^'"]*\/)?bulma-([\w-]+?)(?:\/|\.|['"])/;
 
 /**
- * The source library's own stylesheet (RBC v3's two documented setup
- * paths — the bundled CSS and the "advanced" Sass entry point, bare or
- * `~`-prefixed). This is not a third-party extension: it is the library
- * being migrated away from, and `deps.ts` removes its package.json entry
- * in the same run, so leaving the import in place breaks the Sass build.
+ * The source library's own stylesheet. This is not a third-party extension:
+ * it is the library being migrated away from, and `deps.ts` removes its
+ * package.json entry in the same run, so leaving the import in place breaks
+ * the Sass build. Any `react-bulma-components/…` specifier is dead the same
+ * way — the documented v3 setup paths (the bundled CSS, the "advanced" Sass
+ * entry point), but also deep partials and extensionless forms — so, like
+ * `transform.ts`, we match on the package prefix rather than enumerating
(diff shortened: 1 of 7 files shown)


bestax-migrate/src/sources/react-bulma-components/styles.ts
```diff
@@ -238,6 +252,43 @@ export const transformStyles: StylesTransform = (
       continue;
     }
 
+    const rbcStyle = line.match(RBC_STYLE_IMPORT);
+    if (rbcStyle && !line.includes(TODO)) {
+      const indent = rbcStyle[1];
+      if (rootImportIndex === -1 && !rbcStyleEmitted) {
```
allxsmith: This branch silently drops leading `$var` overrides: pass-1 folding is gated on `rootImportIndex !== -1`, which the RBC specifier can never set, so there's no `with (...)` clause, no TODO, and no report line. RBC v3's documented theming pattern — `$primary: #ff6b35;` above `@import '~react-bulma-components/src/index.sass';` — migrates to a bare `@use` that compiles and loses the entire theme, because `@use` ignores pre-declared vars. On main this input at least got the unrecognized-import TODO, so the PR turns a flagged case into a silent wrong one. Fold the leading vars like the ROOT_IMPORT path does, or at minimum emit the TODO.

bestax-migrate/src/sources/react-bulma-components/styles.ts
```diff
@@ -68,10 +68,23 @@ const OTHER_BULMA_IMPORT = /^\s*@import\s+['"][^'"]*bulma[^'"]*['"]/;
...
+
+/**
+ * The source library's own stylesheet (RBC v3's two documented setup
+ * paths — the bundled CSS and the "advanced" Sass entry point, bare or
+ * `~`-prefixed). This is not a third-party extension: it is the library
+ * being migrated away from, and `deps.ts` removes its package.json entry
+ * in the same run, so leaving the import in place breaks the Sass build.
+ */
+const RBC_STYLE_IMPORT =
+  /^(\s*)@import\s+(['"])(~)?react-bulma-components\/(?:src\/index\.(?:sass|scss)|dist\/react-bulma-components(?:\.min)?\.css)\2\s*;?\s*$/;
```
allxsmith: RBC_STYLE_IMPORT misses specifier shapes this file's own siblings handle: the relative node_modules prefix ROOT_IMPORT explicitly supports (`@import '../../node_modules/react-bulma-components/src/index.sass';`), extensionless `react-bulma-components/src/index`, and deep partials like `src/components/navbar.sass`. All fall through to the generic convert-by-hand TODO while `deps.ts` deletes the package in the same run — the guaranteed-broken build this branch's own comment says every mode must avoid. `transform.ts` already treats any `react-bulma-components/` prefix as dead; the same prefix rule here would close the class.

allxsmith (changes requested):
Four confirmed issues: the two inline (silent theme loss; missed specifier shapes) and two on the review threads (second Bulma entry point colliding with existing `@use` roots; the keep-mode TODO asserting a package removal that `--no-deps` skips). The first three are compile-or-correctness level; the last is wording. This one needs another round before merge.

bestax-migrate/src/sources/react-bulma-components/styles.ts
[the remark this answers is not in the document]
allxsmith: Verified, and it reproduces in two layouts. Same-file: `@use 'bulma/sass';` plus the RBC import yields a duplicate `sass` namespace in bulma/keep mode; in bestax mode the bundle is hard-configured (`bestax.scss` is `@use 'bulma/sass' with ($primary: ...)`) so Sass fails with 'already loaded, so it can't be configured'. Cross-file: a theme file rewritten to a configured `@use 'bulma/sass' with (...)` and a legacy file rewritten to the bestax bundle fail once both reach one compilation graph. Existing Bulma `@use` roots need to be detected and preserved, with bestax mode adding only the extras.

bestax-migrate/src/sources/react-bulma-components/styles.ts
```diff
@@ -238,6 +252,43 @@ export const transformStyles: StylesTransform = (
...
+        // CSS even under `--css keep`.
+        if (cssMode === 'bestax') {
+          out.push(`${indent}@use '@allxsmith/bestax-bulma/scss/bestax';`);
+          extrasAdded = true; // the bundle already includes the extras
+        } else {
+          out.push(`${indent}@use 'bulma/sass';`);
+        }
+        if (cssMode === 'keep') {
+          out.push(
+            `${indent}// ${TODO}: replaced react-bulma-components's own stylesheet with @use 'bulma/sass' — its package.json entry is removed, so the old import no longer resolves; install bulma@^1 — see ${GUIDE}`
```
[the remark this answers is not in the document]
allxsmith: Verified — `migrateFiles` hands the transform only `cssMode`, so under `--no-deps` the TODO and report assert a package removal that didn't happen. The replacement itself is defensible; condition the wording (the PR's own skill doc already carries the '(unless --no-deps)' nuance) or thread the deps setting through to the transform and the report line.

bestax-migrate/src/sources/react-bulma-components/styles.ts
bestaxbot: Fixed. RBC_STYLE_IMPORT now matches any react-bulma-components/... specifier — bare, ~-prefixed, or the relative-node_modules Parcel form — instead of enumerating only src/index.sass and dist/*.css, so deep partials (src/components/navbar.sass) and extensionless forms (src/index) are caught too. That mirrors transform.ts is react-bulma-components/ prefix rule, so every RBC stylesheet gets rewritten into a real Bulma root rather than falling through to the generic convert-by-hand TODO while deps.ts deletes the package. New it.each coverage exercises the relative-path, extensionless, and deep-partial shapes, plus a prefix-preservation case (../../node_modules/bulma/sass + the raw-path extras).

bestax-migrate/src/sources/react-bulma-components/styles.ts
bestaxbot: Fixed. The RBC stylesheet now participates in the same leading-var fold the @import bulma root does: pass 1 folds the vars above whichever line becomes the file is Bulma root (a bulma @import if present, else the RBC import), and computed values still emit the unsafe-var TODO. So $primary: #ff6b35; above @import react-bulma-components/src/index.sass now migrates to @use bulma/sass with ($primary: #ff6b35) + extras, preserving the theme — no more silent loss. Note this required dropping the hard-configured scss/bestax bundle from the RBC path (it cannot take user vars); see my reply on the existing-@use-root thread for the full rationale. Covered by a new folds-leading-$var test.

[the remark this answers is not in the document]
bestaxbot: Fixed. Two changes close this:

1. The RBC path no longer emits the hard-configured scss/bestax bundle. It now emits @use bulma/sass (+ @use .../scss/extras in bestax mode) — the identical shape the @import bulma root path already produces. This removes the bundle-vs-configured-root divergence you flagged cross-file: an RBC-derived file and a bulma/sass-derived file now converge instead of one loading a pre-configured bundle that collides with the other.

2. Same-file existing @use roots are detected (USE_BULMA_ROOT: the file own @use bulma/sass, any prefix, or a bestax bundle). When one is present, foldTargetIndex is -1 so no second root/with() is emitted; bestax mode adds only the extras (load-safe, extras only read utilities), and bulma/keep drop the redundant RBC line. New tests cover the mixed @use bulma/sass + RBC import case in both bestax (one root, extras added once) and bulma (one root, line dropped) modes.

The residual cross-file case where two files each configure bulma/sass with (...) is the inherent Bulma-v1/Sass-modules single-configuration constraint that already applies to the @import root path — not something a per-file line transform can resolve — but the bundle-specific collision you identified is gone.

bestax-migrate/src/sources/react-bulma-components/styles.ts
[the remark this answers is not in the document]
bestaxbot: Fixed by threading the deps setting through to the transform. TransformOptions gained a deps field, cli.ts passes options.deps into transformStyles, and the keep-mode branch now branches on options.deps !== false. With deps on it keeps the accurate its package.json entry is removed... wording; under --no-deps it instead says the RBC stylesheet targets Bulma 0.9 not the v1 your components now use, --no-deps kept the dependency, so install bulma@^1 alongside it — no longer asserting a removal that did not happen. The report() message mirrors the same split, and the skill doc note is worded to match. New test runs keep mode with { deps: false } and asserts the output/report do not claim removal and do mention --no-deps.

allxsmith:
@coderabbitai review

bestax-migrate/src/sources/react-bulma-components/styles.ts
allxsmith: Verified 036cf79: the RBC root is now a fold target — leading overrides land in with (...), unsafe ones get the TODO plus a report line, and the emitted root converges with the @import path's shape. 242 tests pass locally. Resolving.

bestax-migrate/src/sources/react-bulma-components/styles.ts
allxsmith: Verified 036cf79: matching is prefix-based now (bare, ~, and relative node_modules forms, deep partials, extensionless), USE_BULMA_ROOT prevents a second root with bestax mode adding only the extras, and the keep-mode TODO/report branch on the threaded deps flag. Resolving.

allxsmith (approved):
Verified all four findings closed: leading var overrides fold into with (...) with unsafe ones flagged, existing @use roots are detected so no second root is ever emitted (bestax mode adds only the extras), specifier matching is prefix-based across every documented form, and the keep-mode messaging branches on the threaded deps flag. 242 tests pass locally.

allxsmith (approved):
Re-approved after the main merge-back. One seam mattered: emitBulmaRoot predates main's formatFoldedValue, so the merge wires the parenthesization into the RBC path and adds a regression test pinning that a comma-valued override folded through the RBC root emits as a parenthesized list. 252 tests, lint, typecheck, format, skills-sync all green locally.

bestax-migrate/src/sources/react-bulma-components/styles.ts
```diff
@@ -320,6 +372,58 @@ export const transformStyles: StylesTransform = (
...
+        // would reconfigure an already-loaded module.
+        out.push(
+          prefix
+            ? `${indent}@use '${prefix}@allxsmith/bestax-bulma/src/scss/extras';`
+            : `${indent}${EXTRAS_USE}`
+        );
+        extrasAdded = true;
+      }
+      // Otherwise (bulma/keep with a root already present) the line is
+      // redundant — drop it rather than importing Bulma twice.
```
allxsmith: This drop is only "redundant" for the index/root import — for a deep partial it silently loses CSS, and I can reproduce it in both modes:

```scss
/* bulma mode */
@import 'bulma/bulma.sass';
@import 'react-bulma-components/src/components/navbar.sass';
```
migrates to just `@use 'bulma/sass';` — the navbar partial vanishes with **no TODO and no report entry**. Same for a second RBC import in bestax mode (`index.sass` + `tooltip.sass` → root + extras, tooltip partial gone silently). A deep RBC partial's styles aren't guaranteed to be covered by `bulma/sass`, and the package's hard rule is "anything unsafe gets a TODO + a report entry — never a silent skip" (bestax-migrate/CLAUDE.md).

The root/index import being dropped when a root exists is fine as-is. Non-root RBC imports need the TODO + report treatment when removed — something like "removed a react-bulma-components partial; port any styles it carried beyond Bulma's own by hand."

Secondary, non-blocking: the root rewrite itself reports only in keep mode — bestax/bulma modes restructure the file's root with no report line, so the migration summary undercounts what happened. Worth a report entry in all three modes.

allxsmith (changes requested):
One hard-rule violation found while re-reviewing the merged head; inline below with an executed repro. Secondary note folded into the same comment.

bestax-migrate/src/sources/react-bulma-components/styles.ts
bestaxbot: Fixed — good catch, this was a real hard-rule violation.

**The silent-partial drop (blocking).** The old branch dropped *any* RBC line once a Bulma root was present, calling it "redundant." That's only true for the library's own root/index stylesheet, which `bulma/sass` supersedes; a deep partial (`src/components/navbar.sass`) can carry styles `bulma/sass` doesn't. I added `RBC_ROOT_STYLESHEET` to classify the root/index forms (bare specifier, `src/index`, bundled `dist/*.css`), and now a *non-root* RBC import removed beside an existing root gets a `// TODO(bestax-migrate)` + report entry ("port any styles it carried beyond Bulma's own by hand") instead of vanishing.

I also had to split the drop branch so the extras-once and partial-flag concerns are independent — otherwise your bestax repro (`index.sass` + `tooltip.sass` → root + extras) consumed the second RBC line in the extras branch and dropped the tooltip partial silently. Now the extras are added once *and* the partial is flagged. Both repros are covered by new tests: bulma-mode `bulma/bulma.sass` + `navbar.sass`, bestax-mode `index.sass` + `tooltip.sass`, and the existing-`@use`-root + `navbar.sass` case — each asserts one root, extras-once where applicable, and a 'partial dropped' report. A companion test pins that the plain root/index drop stays silent (no false TODO).

**Secondary (reporting completeness).** The RBC root rewrite now emits a report entry in bestax/bulma modes too, not just keep — so the summary reflects the restructured root in all three. I narrowed the 'not mislabeled' test assertion to forbid the *extension* wording rather than the package name, since a report that names `react-bulma-components` while rewriting its root is legitimate (the mislabel we actually guard against is the `bulma-components` third-party-extension label).

Updated `css-migration.md` and the docs migration guide, regenerated the MCP index byte count. Green locally: 256 tests via turbo, lint, typecheck, coverage, format, and full conformance.

allxsmith: Verified 7db47db against the exact repros from this thread: the navbar partial in bulma mode and the tooltip partial in bestax mode both now leave a site TODO plus a report entry, and the RBC root rewrite reports in all three css modes (the secondary note). 256 tests pass locally. Resolving.

allxsmith (approved):
Verified 7db47db: dropped RBC deep partials now carry a site TODO and a report entry in every mode (reproduced both failure cases from the review — both fixed), and the root rewrite reports in all three css modes. 256 tests, lint, typecheck, skills-sync green locally.

allxsmith (approved):
Re-approved after the post-559 merge-back: clean auto-merge (the two skills.json restamps touch different lines), gen:mcp reproduces the merged manifest byte-for-byte, 257 tests green locally.

Issue #487, opened 2026-08-07
[Security] harden-runner silently downgrades egress-policy: block to audit — no AI job has ever enforced egress


allxsmith:
## Problem

> **Superseded 2026-08-27.** harden-runner v2.21.0 fixed the root cause upstream, so **step 2
> below (seed the cache entry) is obsolete** — do not build it. #577 implements the revised fix
> and closes this issue. See the closing comment for details.

`ai-scan.yml` and `claude-repro.yml` both declare `egress-policy: block`. Neither one is
enforcing it. harden-runner silently downgrades them to `audit` at startup, and several
security comments describe a control that is not running.

**Correction (applied 2026-08-27; proposed 2026-08-16 and re-raised, never made until now).**
This paragraph originally claimed "**no job in this repository has ever blocked an outbound
connection**". That is false, and this issue's own comments falsified it. The cause is GitHub
making the Actions cache read-only for **untrusted triggers**, so anything on `schedule`,
`workflow_dispatch` or `release` is outside that class and arms normally —
`security-txt-expiry.yml` reported `EgressPolicy:block` on run 31952287389, and
`supply-chain.yml`'s `consumer-sbom` did the same on run 32088629752 and actually refused a
connection.

The accurate statement has three categories, not one:

| State | Workflows |
| --- | --- |
| Declares `block`, **downgraded** in practice | `ai-scan`, `claude-repro` |
| Declares `audit` **by declaration**, not by downgrade | `ai-triage` |
| **No harden-runner at all** — nothing to downgrade | `claude`, `claude-implement`, `claude-pr-loop`, `claude-review` |

Steps 2 and 3 below only ever addressed the first group: the pre-step that writes `agent.json`
never runs for the third, so neither step does anything for those four. That gap is now
tracked in #578.

Found while collecting the endpoint list for #456's audit-to-block flip.

## Evidence

harden-runner says so itself. From the `author` job of run
[31137122500](https://github.com/allxsmith/bestax/actions/runs/31137122500), which is configured
`egress-policy: block`:

```
##[warning]Cache reservation failed: cache write denied: token has no writable scopes
Failed to save: Unable to reserve cache with key harden-runner-cacheKey, another job may be creating this cache.
cache version: v2
Unable to fetch cacheURL TypeError: Invalid URL
Switching egress-policy to audit mode
```

and its effective config, logged in the post-step:

```
EgressPolicy:audit
```

Same three lines in the `ai-scan` run
[...]

Issue #455, opened 2026-08-02
[Security] ai-scan's Claude session runs with a write-scoped GITHUB_TOKEN


allxsmith:
## Problem

`.github/workflows/ai-scan.yml` runs its whole pipeline in one `scan` job, and that job needs
write permissions:

```yaml
permissions:
  issues: write        # apply needs-security-review on issues; budget marker
  pull-requests: write # apply needs-security-review on PRs
```

The Claude session step sits inside that job, so **the model session holds a write-scoped
`GITHUB_TOKEN`**. It is read-only in practice only because the Bash allowlist admits nothing
that writes (`gh view` / `gh diff`).

That is one control where it reads like two. Widening `allowedTools` later — a plausible,
innocuous-looking change — silently hands an injected session write access to issues and pull
requests, with no permission diff for a reviewer to notice.

The header comment previously asserted the session held "a read GITHUB_TOKEN", which was simply
untrue; that has been corrected to describe the actual arrangement, but the arrangement itself
is what wants fixing.

## Fix

Split the job so the session runs with `contents: read` only. The write steps bracket it, so it
is a three-way split:

- **gate** — `issues: write`, charges the daily budget marker on the tracking issue; outputs
  whether to run
- **scan** — `contents: read`, the Claude session, outputs a verdict
- **label** — `issues: write`, `pull-requests: write`, applies `needs-security-review`
  fail-closed on a non-clean verdict

This is the pattern the repo already uses twice: `claude-repro.yml` (`author` holds the model
token, `publish` holds the write token, neither holds both) and `test-deploy.yml` since #442
(`build` runs repo code with no secrets, `test-deploy` holds the Cloudflare credentials and runs
no repo code).

## Why not in #361

Raised as an advisory in the deep review on that PR. It is a defense-in-depth improvement rather
than a live hole — the session genuinely cannot write today — and restructuring a 260-line
security workflow at merge time, without its own review pass, trades more risk than it removes.

## Acceptance

- [ ] The Claude session step runs in a job with no `write` permission of any kind
- [ ] The fail-closed labeler behaviour is unchanged (crashed / unparsable / `is_error` session
      still flags)
- [ ] The budget accounting on the tracking issue is unchanged
- [ ] The header comment's description of the token scope matches the new structure

PR #570, opened 2026-08-26
Extract the workflow security parsers into tested scripts

### LINKED ISSUE #454: [Refactor] Extract the shell security parsers from workflow YAML so they can be tested
## Problem

Two pieces of load-bearing security logic live as shell embedded in workflow YAML, which means
they cannot be executed outside a workflow run and therefore have no regression guard:

- **The fail-closed verdict parser** — `.github/workflows/ai-scan.yml`. Decides whether a scan
  verdict is clean or flags the item with `needs-security-review`. It must fail *closed*: a
  crashed, empty, unparsable or inconclusive session has to flag rather than pass.
- **The publish sanitizer** — `.github/workflows/claude-repro.yml`. Neutralizes `@mentions`,
  machine markers (`<!-- ... -->`, `Duplicate of #N`, the `*-RESULT` sentinels) and fence
  delimiters in an attacker-influenced draft before it is posted to a public comment.

Both were verified by hand at review time — the verdict parser was traced across seven input
shapes (clean / flagged / `is_error`+clean / missing record / trailing-clean / empty / missing
file) and failed closed in all of them. That is a point-in-time check, not a guard. The next
edit to either `sed` pipeline or `jq` filter has nothing stopping it from quietly widening.

The sanitizer already has a concrete cross-file coupling worth protecting: it defangs
`Duplicate of #` precisely because `github-actions[bot]` is an author that
`scripts/auto-close-duplicates.mjs` trusts. Two patterns in two files that must agree.

## Why this is not just "write some tests"

They are not callable. Testing them as-is would mean grepping the step body out of the YAML and
`eval`-ing it, which is brittle in exactly the way that makes people delete tests.

The real fix is to extract each into a file the workflow calls — `scripts/*.sh` or `.mjs` — and
test that file directly. That is a refactor of security-critical automation, so it wants its own
PR and its own review pass rather than riding along with unrelated work.

## Scope

- [ ] Extract the `ai-scan.yml` verdict parser to a callable script; test it against the seven
      input shapes above, asserting it fails closed on every degenerate one.
- [ ] Extract the `claude-repro.yml` sanitizer to a callable script; test mentions, each machine
      marker, fence delimiters, and the 12000-byte truncation.
- [ ] Keep the workflow steps thin wrappers so the YAML stays readable.

## Already done

The third item from the original finding is covered: `scripts/auto-close-duplicates.mjs` now has
tests (`scripts/auto-close-duplicates.test.mjs`), and root `pnpm test` runs
`node --test "scripts/*.test.mjs"` so root-level scripts are no longer invisible to CI. That
runner is the hook these extracted scripts would plug into.

## Context

Raised as an advisory in the deep review on #361. Not blocking that PR — nothing there is known
to be wrong, and `AI_TRIAGE_AUTOCLOSE` is currently `dry-run` — but the gap is real and outlives
the PR.

allxsmith:
Closes #454.

Extracts the two inline shell security parsers that `.github/CLAUDE.md` rule 9 tracks under #454 into `scripts/*.mjs` with `node --test` siblings, leaving the workflow steps as thin wrappers. No behavior change is intended, and the parity work below is what backs that claim.

## What moved where

**`scripts/parse-scan-verdict.mjs`** replaces the jq/grep/sed block in `ai-scan.yml`'s fail-closed labeler. CLI contract: always exits 0, prints exactly `clean` or `flagged <category>`; every internal problem prints `flagged other` and never echoes input-derived text to stdout or stderr (the execution file contains echoed attacker text and is promised to stay off the public log). The YAML wrapper keeps its own `flagged`/`other` default, upgrades only on a positively parsed verdict (`OUT=$(node …) || OUT=''`), and admits a category into the step summary only through a closed case enum — so a failed invocation (checkout failed under `always()`, script missing) still flags. Parsing parity details worth reviewing: the stream fallback is all-or-nothing (skipping unparsable lines could drop a final `is_error: true` record and promote an earlier clean one — a fail-open parse), `is_error` must be strictly `false` (gating the category branch too, as the shell's `$OK` did), and last-non-empty-line filters on length rather than trim (jq `select(length > 0)` parity — a trailing whitespace-only line still flags).

**`scripts/sanitize-repro-draft.mjs`** replaces the five-stage `sed | head -c 12000` pipeline in `claude-repro.yml`'s publish job. It round-trips stdin as latin1 bytes, so invalid UTF-8 passes through byte-identically and truncation is exactly `head -c` (applied after entity expansion, as before). The wrapper runs it with **no fallback** under `set -euo pipefail`: a missing or crashing sanitizer fails the step and nothing unsanitized can ever reach the comment — the opposite failure shape from the parser wrapper, and each is deliberate (invariant I2 vs. rule 4).

The test sibling imports `MARKER` and `DUPLICATE_RE` from `scripts/auto-close-duplicates.mjs` and asserts sanitized output can never satisfy that consumer — the cross-file coupling the old YAML comment described in prose is now an executed assertion.

## Parity verification (done before commit)

- Verdict parser: the original jq/grep/sed fragment (run verbatim with jq 1.7.1) and the new CLI agree on **27/27** fixtures covering the seven hand-traced shapes plus degenerates (stream form, bad stream line, mixed non-object values, missing file, CRLF, unknown category, absent `is_error`). The full YAML wrapper glue was also simulated end-to-end: 28/28.
- Sanitizer: the original pipeline (GNU sed in an `ubuntu:24.04` container) and the new CLI are byte-identical on **12/13** fixtures including invalid UTF-8, a multibyte char straddling byte 12000, and entity expansion pushing content past the cap. The 13th is an accepted divergence: JS `^`/m also treats a bare `\r` as a line boundary, so a CR-delimited sentinel gets defanged where sed left it **live** — strictly more sanitized, pinned by a test with a do-not-"fix" note. A second defang-more divergence was added after CodeQL (js/bad-tag-filter) flagged the extracted regex: the sanitizer now also breaks the alternate HTML comment close `--!>`, which the old sed left alone (same gap, invisible to analysis inside YAML). Both are documented in the script header and pinned by tests; all other inputs were re-verified byte-identical against GNU sed.

## Workflow-change checklist (.github/CLAUDE.md)

- **Allowlists**: none grow. `ai-scan.yml`'s `--allowedTools` is untouched.
- **Model sessions**: no session gains execution, fetch, or network. The new checkout + setup-node in the publish job hold no model token; the drafting job is untouched.
- **Sanitizer between model text and comments**: still deterministic, now tested; the publish path fails closed on sanitizer failure.
- **Posting identity**: unchanged — `GITHUB_TOKEN` (github-actions[bot]) everywhere; still no PAT in either workflow.
- **Repository variables**: none added.
- **Action SHAs**: the new `checkout` and `setup-node` steps use the repo-wide pins (post-change audit: exactly one SHA per action across `.github/workflows/`).
- **Verdict path fails closed** at both layers (script contract + wrapper default), now regression-guarded.
- **Permissions**: `publish` gains `contents: read` for the checkout. Its trigger is `issues: labeled`, so the ref is trusted default-branch content — no attacker-controlled ref exists for the event. Two deliberate non-additions, per the no-drive-by posture: `publish` still has no harden-runner (rule 10 binds new jobs; this one is existing), and `ai-scan`'s harden-runner allowlist was not widened for setup-node's potential nodejs.org fetch — egress is not enforced anywhere today (#487) and that list is already documented as incomplete there.

Out of scope, on purpose: the #455 labeler job split, and the three smaller inline exec-file sentinel checks (`ai-triage.yml`, `claude-review.yml`, `claude-repro.yml`'s author job) — rule 9 now names them for extraction whenever one next needs an edit, with `parse-scan-verdict.mjs`'s exported helpers as the reuse point.

Verified locally: `pnpm test` (turbo suites + 426 script tests), `pnpm lint`, `pnpm format:check` all green.

## Summary by CodeRabbit

- **New Features**
  - Added robust parsing for AI scan results in JSON and newline-delimited formats.
  - Added draft sanitization that removes automation-sensitive markers and limits content to 12,000 bytes.
  - Improved automated scan verdict handling with supported clean and flagged outcomes.

- **Bug Fixes**
  - Invalid, missing, incomplete, or errored scan results now fail closed as flagged.
  - Sanitization failures or empty sanitized drafts now prevent publication.

- **Tests**
  - Added comprehensive coverage for scan parsing, draft sanitization, malformed input, and byte-limit edge cases.

### DIFF
--- a/.github/CLAUDE.md
+++ b/.github/CLAUDE.md
@@ -191,8 +191,14 @@ writing the trigger string at all.
 Shell embedded in a workflow step cannot be unit-tested without extracting it first. Put
 non-trivial parsing in `scripts/*.mjs` with a `node --test` sibling — root `pnpm test` runs
 `node --test "scripts/*.test.mjs"` (the glob is quoted so Node expands it, not the shell).
-Two parsers are still inline and tracked in #454: the publish sanitizer and the scan-verdict
-parser.
+The two #454 parsers are extracted: the scan-verdict parser
+(`scripts/parse-scan-verdict.mjs`, called by `ai-scan.yml`) and the publish sanitizer
+(`scripts/sanitize-repro-draft.mjs`, called by `claude-repro.yml`) — their test siblings pin
+the fail-closed matrix and the byte behavior of the shell they replaced, so edit script and
+tests together. Smaller instances of the same shape remain inline (the exec-file sentinel
+checks in `ai-triage.yml`, `claude-review.yml`, and `claude-repro.yml`'s author job); when
+one of those next needs an edit, extract it and reuse `parse-scan-verdict.mjs`'s exported
+helpers rather than growing the YAML.
 
 ### 10. `harden-runner` on new jobs starts at `block`, and `block` does not currently mean block
 
--- a/.github/CLAUDE.md
+++ b/.github/CLAUDE.md
@@ -200,6 +200,21 @@ checks in `ai-triage.yml`, `claude-review.yml`, and `claude-repro.yml`'s author 
 one of those next needs an edit, extract it and reuse `parse-scan-verdict.mjs`'s exported
 helpers rather than growing the YAML.
 
+**Extraction has a bootstrap gap: a new script flags its own introducing PR.** A workflow runs
+the YAML from the PR merge ref, but the jobs here deliberately check out the default branch
+(never PR head code). A script the PR adds therefore does not exist in the workspace its own
+run reads: `node` exits `MODULE_NOT_FOUND`, the caller takes its fallback, and a fail-closed
+verdict path labels the item on infrastructure grounds without the verdict ever being read.
+Origin: #570 applied `needs-security-review` to itself this way, which then gated
+`claude-repro`, `claude-fix`, `@claude` and `@bestaxbot` on that PR.
+
+Do not close this by checking out the PR head — running PR-authored code in a job holding a
+credential is the thing that pin prevents, and trading it away for a cosmetic label is a far
+worse deal than the flag. Treat the gap as expected cost instead: say in the PR description
+that the flag is self-inflicted, and clear the label by hand once the script is on the default
+branch. The same applies to moving or renaming a script a workflow already calls — the rename
+lands in the YAML a run before it lands in the tree that run checks out.
+
 ### 10. `harden-runner` on new jobs starts at `block`, and `block` does not currently mean block
 
 New jobs ship with `egress-policy: block`. An **existing** live job may be introduced at `audit`
--- a/.github/workflows/ai-scan.yml
+++ b/.github/workflows/ai-scan.yml
@@ -195,13 +195,6 @@ jobs:
           ref: ${{ github.event.repository.default_branch }}
           persist-credentials: false
 
-      # No pnpm, no cache: the verdict parser below is dependency-free
-      # (node: builtins only), so plain Node 24 is all it needs.
-      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7
-        if: steps.gate.outputs.run == 'true'
-        with:
-          node-version: '24'
-
       - name: Run Claude (security scan)
         id: claude
         if: steps.gate.outputs.run == 'true'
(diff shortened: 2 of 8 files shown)


allxsmith:
Note on the `needs-security-review` label: it is the extraction demonstrating its own fail-closed path, not a model verdict.

This PR's scan run took the workflow file from the merge ref — so the **new** wrapper ran — but `ai-scan` deliberately checks out the **default branch**, which does not contain `scripts/parse-scan-verdict.mjs` yet. The `node` invocation failed (`MODULE_NOT_FOUND`, visible in run 32915532640), the wrapper kept its `flagged`/`other` default, and the label was applied — exactly the behavior the wrapper's comment promises for a missing script.

This is a one-time cross-over quirk unique to this PR: every other PR's scan runs its own merge-ref workflow (the old inline jq), and after merge the workflow and the script travel together on `main`. The label will need a maintainer to remove it as usual.

allxsmith:
@coderabbitai review

PR #577, opened 2026-08-27
ci: enforce egress-block on the AI workflows instead of declaring it

### LINKED ISSUE #487: [Security] harden-runner silently downgrades egress-policy: block to audit — no AI job has ever enforced egress

allxsmith:
Closes #487.

> **Body rewritten 2026-08-28** to describe the change as it now stands, and updated again after
> review round 4. Four review rounds materially altered it, and the original text understated the
> allowlist changes — which rule 2 makes a security problem, not a tidiness one. Review history is
> in the comments.

## What this is

`ai-scan` and `claude-repro` have declared `egress-policy: block` and enforced nothing since
GitHub made the Actions cache read-only for untrusted triggers on 2026-06-26. harden-runner
v2.20.0 armed block mode by reading a cache entry, and on the resulting miss its catch block set
`egress_policy = "audit"` — a security control failing open, announced with one `core.info` line.

**The issue's written fix plan is obsolete and this PR does not follow it.** #487 step 2 asks for
a workflow seeding `harden-runner-cacheKey` from a trusted trigger, re-seeded after every pin bump
and before each 7-day eviction. None of that is needed:
[harden-runner v2.21.0](https://github.com/step-security/harden-runner/releases/tag/v2.21.0)
removes the downgrade path outright. Upstream
[#675](https://github.com/step-security/harden-runner/issues/675) is closed as fixed, and the
maintainer declined the `fail-on-policy-downgrade` input the issue asked for, "since there is no
downgrade left to fail on". A pin bump supersedes all of step 2.

## Ordering hazard — please do not merge #510 before this

Dependabot #510 bumps harden-runner 2.20.0 → 2.21.0 grouped with `pnpm/action-setup` and
`codeql-action/upload-sarif`. **Merging it alone arms block on jobs whose allowlists are
incomplete**, and `ai-scan` / `claude-repro` would fail at the Claude CLI download. The bump must
land with the allowlist work, which is why it is here. Once this merges, dependabot will rebase
#510 and drop harden-runner from the group.

## Allowlist changes (rule 2 — justified per entry)

**Three entries added** to `ai-scan`, `claude-repro` and `ai-triage`:

| Entry | Why | Can it write to this repo? |
| --- | --- | --- |
| `claude.ai:443` | `claude-code-action` curls the CLI installer from here. Reached on every measured run; listed on none. | No — serves the installer script. |
| `downloads.claude.ai:443` | The installer pulls the CLI binary from here. | No — static distribution. |
| `release-assets.githubusercontent.com:443` | The action's first composite step is `oven-sh/setup-bun`, which fetches bun from a GitHub **release asset**. Every other block job in this repo already lists this host. | No — GitHub's release-asset CDN. |

The third was found in review, not by measurement, and the reason is worth keeping: the
verification run showed `Install Bun` succeeding in 2.1s off the Actions cache (`Cache hit … Using
a cached version of Bun`), so the download path was never exercised. These jobs can only **read**
that cache (#487's root cause), so an eviction or a bun bump turns the restore into a download the
allowlist must permit — otherwise `ai-scan` fails on every incoming item with the assertion step
still green.

**One entry removed: `statsig.anthropic.com`.** This is a correctness fix, not tidying. The domain
has gone **NXDOMAIN**, and harden-runner's agent aborts and reverts its firewall when an
allow-listed host will not resolve — leaving `block` in its config and nothing enforcing. That is
what happened on this PR's first run. A dead allowlist entry is not inert; it silently disables
the whole policy. #487 had already measured the host as never contacted. Every other allow-listed
host in the repository resolves (checked).

**`http-intake.logs.us5.datadoghq.com` is denied, not allow-listed.** It is CLI telemetry, not a
functional dependency, so it stays off every list *and* is silenced at the source with
`CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC`. Denying and silencing is tighter than allow-listing;
the switch also covers feature-flag evaluation, which is what `statsig.anthropic.com` was for.

**Scope change from the original plan:** that setting now applies to **all nine**
`claude-code-action` steps, not just the three block-mode ones. Leaving it off the audit-mode jobs
would have put the Datadog host into the very endpoint report #578 depends on, inviting whoever
implements #578 to allow-list a host we just decided to deny. #578 step 4 is updated to match.

## Fail-closed assertion

Every block-mode job now carries, after harden-runner:

```yaml
- name: Assert egress policy is enforced
  run: |
    set -uo pipefail
    [ -e /home/agent/agent.json ] || { echo "::error::harden-runner installed no agent"; exit 1; }
    jq -e '.egress_policy == "block"' /home/agent/agent.json || exit 1
    for _ in $(seq 1 30); do
      grep -q '^Initialized' /home/agent/agent.status 2>/dev/null && exit 0
      sleep 1
    done
    exit 1
```

That is rule 10's canonical form, trimmed of the error strings the jobs carry; copy the full
version out of any block job rather than this snippet. Three shapes in it are load-bearing and
were each added in review:

- **The `-e` check first**, because harden-runner has deliberate install-nothing-exit-0 paths (a
  StepSecurity outage, the `skip-harden-runner` property, a container runner). It still fails —
  a job holding a credential must not run unprotected — but it says why, instead of a bare
  `jq: could not open file`.
- **`^Initialized`, not `-qx`.** `writeStatus` appends without a trailing newline, so a second
  status concatenates onto the same line and an exact-line match stops matching.
- **The status check polls.** The pre-step waits only for the file to *exist* and gives up after
  ~9s, while the agent resolves every allow-listed host before writing its status.

Both files are needed. `agent.json` holds the policy the pre-step *decided* and catches a
downgrade (#487). `agent.status` is written only once the firewall rules are installed, and
catches the other failure: `agent.json` is written *before* the agent starts, so an agent that
never comes up leaves `block` in the config with no firewall.

**Scope, stated honestly:** this proves enforcement was armed **at that step**. It does not prove
the policy stays armed — the agent writes `Initialized` then serves, and a later runtime error
makes it revert with both files unchanged. Rule 10 documents that, and documents why a
`systemctl is-active` check is *not* the fix (it covers a sub-second window while binding every
block job to an internal unit name and the TLS-path binary selection).

**Step order matters in `ai-scan`, in both directions.** The assertion sits *after* the gate: the
fail-closed labeler is guarded by `always() && steps.gate.outputs.run == 'true'`, so an abort
before the gate would leave items both unscanned and unflagged — a fail-OPEN verdict path (rule 4)
introduced by a fail-CLOSED check.

It also sits *before* the budget charge, which is why the gate now only **reads** the counter and
a separate later step writes it. Found in review: with read and charge in one step, every run that
failed the assertion spent a scan slot without scanning. `AI_SCAN_DAILY_LIMIT` such failures in
one UTC day — and a StepSecurity outage hits every incoming item at once — put the gate back to
`run=false`, which skips that same labeler and leaves later items unscanned *and* unflagged. The
same fail-open, one step removed. The charge still precedes the session, so a crashed scan counts;
splitting the steps does not make the counter atomic and does not claim to, since two concurrent
runs raced identically when read and write shared a step.

Kept inline rather than factored into a composite action — but not for the reason an earlier
revision of this body gave. Rule 1's greps recurse over `.github/`, so a pin under
`.github/actions/` still matches; the claim that it would "match zero lines" was wrong, and rule 1
now says so. The real reason is that reading the job is the whole review signal here:
`harden-runner`'s policy and allowlist are argued per job against the credential in *that* job
(rule 2), and a shared wrapper would collapse those separate allowlists into one — a widening with
no `permissions:` diff.

## `ai-triage` audit → block

Introduced at `audit` because the action's runtime egress was undocumented, and that caution was
justified — the audit run produced the measured list. Both preconditions now hold, so it flips.
This is the job holding `AI_LOOP_PAT`. Note what it does not buy: `api.github.com` is necessarily
allow-listed, so egress-block cannot stop a write issued through the PAT — the tool allowlist
still carries that.

## The six jobs that had no harden-runner

`claude`, `claude-implement`, `claude-pr-loop` (`fix` and `verify`), `claude-review` and
`bestaxbot-reply` had none, so the trigger class was never why they did not enforce — there was
nothing to downgrade. They now get one at **`audit`** under rule 10's "existing live job" clause:
they check out branches, install dependencies and run sessions with a model token, and none of
that egress has ever been measured, so a guessed allowlist would break them on the first miss.
Audit produces the list a block flip needs. Follow-up: **#578**.

`bestaxbot-reply` was added in review — it runs repo code holding both the PAT and the OAuth
token, and my first inventory missed it.

## Contract updates (`.github/CLAUDE.md`)

- **I1** scoped to the pipeline built around it. The unqualified claim was already false:
  `claude-implement`, `claude-pr-loop`, `claude-review`, `claude` and `bestaxbot-reply` all run
  repo code beside the model token. Named, with the weaker controls that hold them.
- **Rule 10** retitled and rewritten: mandates the assertion, states what it does and does not
  prove, carries the unresolvable-host hazard, and tabulates enforcement **per job** — including
  that only `claude-repro`'s `author` is hardened, while `publish` (which runs the sanitizer over
  attacker text holding `issues: write`) has no egress policy at all. It also states the placement
  rule in full: after any gate an `always()` fail-closed step depends on, **and** before any step
  that spends a metered budget — a fail-closed check must not be able to consume the resource that
  keeps the fail-closed path reachable.
- **Rule 1** records that the greps are the whole verification, and that the pin does not fully
  determine which agent binary runs (`isTLSEnabled` is a 3s probe that fails open to a
  source-unavailable binary). It also gains a **repo-wide** pin check — it only ever documented a
  per-action grep while requiring one SHA per action across the repo, so an action left stale by
  an earlier bump was invisible to it. And the keep-it-in-the-workflow requirement stands on its
  real reason now, not the wrong one described above.
- **Review checklist** gains an item for the assertion.

## Verification

| Job | Verified enforcing | How |
| --- | --- | --- |
| `ai-scan` | ✅ | live `pull_request: opened`, real Claude session |
| `ai-triage` | ✅ policy only | same run; gate skipped the session |
| `consumer-sbom` ×4 | ✅ | dispatch |
| `security-txt-expiry` | ✅ | dispatch (dry-run) |
| `claude-repro` | ❌ | `issues` runs the default branch's file — post-merge only |
| `deploy-worker` | ❌ | deliberately not dispatched; deploys to production |
| `sign-sbom` | ❌ | `release`-only |

Decisive run after the statsig fix
([33137148739](https://github.com/allxsmith/bestax/actions/runs/33137148739)): `Initialized` ×4,
zero `Reverted changes`, zero resolution failures, `EgressPolicy:block`, assertion green, session
completed.

Local, **against the current head** (these counts moved when `main` merged — `ai-scan` and
`ai-triage` were split into 3 and 4 jobs there, adding five harden-runner sites):

- **18** harden-runner uses, all on one SHA — both the per-action and the repo-wide pin check
- **12** `block` jobs, and all 12 carry the assertion; **6** `audit` jobs, the intended ones
- **9** `claude-code-action` sites, all setting the telemetry variable via step `env:`
- all 23 workflows parse; every allow-listed host resolves; `pnpm format:check` clean

Each of those is read off the parsed YAML rather than counted by eye, which is how two earlier
versions of this section got the number wrong.

> ### ⚠️ The live evidence is stale, and this is the main thing left before merge
>
> Run 33137148739 exercised a **single-job `ai-scan` that no longer exists**. It predates the
> `main` merge entirely. The seven `ai-scan` / `ai-triage` jobs that now carry assertions have
> never run in their current shape, and neither has the fail-closed path this PR reworked — the
> `enforcement=failed` output, `label`'s widened guard, or the after-write assertions on `label`
> and `cleanup`.
>
> Opening any issue or PR against this branch fires both workflows and would produce that
> evidence cheaply. The failing paths are the ones that matter and none of them are exercised by
> a green run.

## Revert path

One line per job (`block` → `audit`, or the pin back to `bf7454d`), but the assertion step must
come out with it or it will fail every reverted job.

## Summary by CodeRabbit

* **Security Enhancements**
  * Expanded network-egress monitoring and blocking across automated workflows.
  * Workflows now verify security controls are installed, configured for blocking, and initialized before continuing.
  * Added clearer failure reporting when enforcement is unavailable or fails to initialize.
  * Expanded permitted endpoints to support release asset downloads.
  * Separated automated scanning and triage stages to limit credential access.

* **Privacy**
  * Disabled nonessential Claude Code telemetry across applicable workflows.

* **Documentation**
  * Updated security guidance, workflow limitations, validation behavior, and enforcement expectations.


allxsmith:
## Verification: block is enforced on an untrusted trigger

Opening this PR ran the modified `ai-scan.yml` on `pull_request: opened`, which is the live
test the description promised — `pull_request` events use the workflow file from the merge
ref, so this is the new config on a genuinely untrusted trigger.

**`ai-scan` — [run 33116609593](https://github.com/allxsmith/bestax/actions/runs/33116609593), all steps green.**

Effective config from the pre-step:

```
"allowed_endpoints":"api.anthropic.com:443 statsig.anthropic.com:443 claude.ai:443
downloads.claude.ai:443 api.github.com:443 github.com:443
objects.githubusercontent.com:443 registry.npmjs.org:443",
"denied_endpoints":"","egress_policy":"block"
```

The four things worth checking:

1. **Assertion passed** — `Assert egress policy is enforced` ran `jq -e ... /home/agent/agent.json` and printed `true`.
2. **Post-step reports `EgressPolicy:block`** — twice (pre and post).
3. **`Switching egress-policy to audit mode` appears zero times.** This is the line #487 is about; on v2.20.0 it was in every one of these runs.
4. **No blocked connections.** The only grep hit for "denied" in the whole log is the `"denied_endpoints":""` field in the config echo.

Most importantly **`Run Claude (security scan)` succeeded**, so the session installed and ran
the CLI under an enforcing block policy. That exercises exactly the failure an unsequenced
#510 merge would have caused, and confirms `claude.ai` + `downloads.claude.ai` are the
complete distribution set. Nothing tried to reach the Datadog intake host, so denying it and
setting `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC` behaved as intended.

Incidental datapoint that does not change the fix: the cache save *succeeded* on this run
(`Sent 258 of 258`, then `Adding cacheHost: productionresultssa13.blob.core.windows.net:443`),
so a same-repo `pull_request` is evidently not in the read-only class that broke `issues`
runs. Irrelevant to correctness under v2.21.0 — the read-first path no longer changes the
policy either way — but worth recording, because it means a same-repo PR was never the
strongest reproduction of #487.

**`ai-triage` — [run 33116609580](https://github.com/allxsmith/bestax/actions/runs/33116609580).**
The audit → block flip arms: assertion passed, `EgressPolicy:block` in both steps. Being
precise about what this run does *not* show — the gate skipped the session on this PR, so the
Claude step did not execute under block here. The allowlist it would use is byte-identical to
`ai-scan`'s and `ai-scan` exercised it with a real session, so the risk is small, but it is
inference rather than observation until the first real triage run after merge.

Still outstanding, as described in the PR body: `claude-repro` and `ai-triage`'s `issues` path
run the default branch's file and only take effect after merge; `sign-sbom` cannot be
exercised until a release.

allxsmith:
## Verification, part 2: trusted triggers

Both dispatchable block-mode workflows exercised from this branch. All green, all reporting
`EgressPolicy:block`, zero downgrade lines.

**`security-txt-expiry`** — [run 33117381591](https://github.com/allxsmith/bestax/actions/runs/33117381591),
dispatched with `dry-run=true` so it logged without touching the renewal issue. `Assert egress
policy is enforced` passed; the job completed normally through `Open or update the renewal
issue`.

**`supply-chain`** — [run 33117428299](https://github.com/allxsmith/bestax/actions/runs/33117428299),
conclusion `success`. The assertion ran **four times** and passed on every leg, which is the
`consumer-sbom` matrix case worth confirming:

| Job | Assertion |
| --- | --- |
| `Consumer SBOM (@allxsmith/bestax-bulma)` | pass |
| `Consumer SBOM (create-bestax)` | pass |
| `Consumer SBOM (bestax-migrate)` | pass |
| `Consumer SBOM (bestax-mcp)` | pass |

`Sign the SBOMs` skipped, as expected — it is `release`-only and still cannot be exercised.

### Coverage summary

| Job | Assertion verified | How |
| --- | --- | --- |
| `ai-scan` | ✅ | live `pull_request: opened` on this PR, with a real Claude session |
| `ai-triage` | ✅ policy only | same run; gate skipped the session |
| `consumer-sbom` ×4 | ✅ | dispatch |
| `security-txt-expiry` | ✅ | dispatch (dry-run) |
| `claude-repro` | ❌ | `issues` runs the default branch's file — post-merge only |
| `deploy-worker` | ❌ | deliberately not dispatched; it deploys to production |
| `sign-sbom` | ❌ | `release`-only |

Five of seven block jobs confirmed enforcing before merge. The three unverified paths are
unverifiable pre-merge rather than skipped, and `deploy-worker` is the one I would watch first
after merge — it is on `push` to `main`, so it runs immediately.

CI is green (`Build and Test`, both React majors, CodeQL including `Analyze (actions)`).

allxsmith:
## Correction: my first verification comment was a false positive

Copilot's review finding was right, and it invalidates the "block is enforced" claim I posted
[above](https://github.com/allxsmith/bestax/pull/577#issuecomment-5445260712). Retracting that
comment's conclusion and replacing it with the real result.

### What was wrong

The assertion checked `agent.json`, which records the policy harden-runner's pre-step
**decided**. That file is written *before* the agent starts. From `src/setup.ts` at the pinned
SHA, if the agent never comes up the pre-step polls for `agent.status` 30 times, then:

```js
if (counter > 30) { console.log("timed out"); /* print log */ break; }
```

…breaks out and falls through to `process.exit(0)`. Success. So a job can hold `block` in
`agent.json` with no firewall running, and my assertion passed over exactly that.

**It was not hypothetical — it happened on this PR's own `ai-scan` run.** Re-reading
[run 33116609593](https://github.com/allxsmith/bestax/actions/runs/33116609593) past the config
echo I had quoted:

```
Error resolving allowed domain ... unable to resolve domain statsig.anthropic.com., status 3
unable to resolve domain: statsig.anthropic.com. err: ... status 3
Reverted changes
```

with `timed out` twice and no `Initialized`. The agent **reverted its firewall and gave up**.
Egress was not enforced in the run I presented as proof that it was. I checked the log for
blocked connections and downgrade lines, found none, and read that as success — but silence
there meant the firewall was never up, not that nothing was refused.

### Root cause, and the general rule it produces

`statsig.anthropic.com` has gone **NXDOMAIN**, and harden-runner's agent aborts and reverts when
an allow-listed host will not resolve. **A dead allowlist entry is not inert — it silently
disables the entire policy.** #487 had already measured that this host was never contacted; it
was sitting in three allowlists doing nothing but switching the firewall off. I resolved every
allow-listed host across all workflows: `statsig.anthropic.com` is the only dead one.

There is an unpleasant symmetry worth naming, since it is the whole subject of #487: a PR fixing
a control that silently was not running shipped a control that silently was not running, and for
the same reason — checking declared configuration instead of effective state.

### Fixes (3d3b39f)

- **Drop `statsig.anthropic.com`** from `ai-scan`, `ai-triage`, `claude-repro`.
- **Assert the agent is live**, not just configured:
  ```yaml
  run: |
    jq -e '.egress_policy == "block"' /home/agent/agent.json
    grep -qx Initialized /home/agent/agent.status
  ```
  `agent.status` is written only after the rules are installed, so a reverted firewall now fails
  the job. Both lines are needed: the first catches a downgrade, the second catches a revert.
- **`bestaxbot-reply.yml` added at `audit`.** Copilot was right that my rule 10 inventory omitted
  it while it checks out code and runs a session holding both the PAT and the OAuth token — same
  class as the five I had listed. Also added to #578.
- **I1 scoped honestly.** Copilot flagged that my new inventory contradicted it. I1's unqualified
  form was already false — `claude-implement`, `claude-pr-loop`, `claude-review`, `claude` and
  `bestaxbot-reply` all run repo code beside the model token. It is a property `claude-repro` is
  built to preserve, and now says so, naming the weaker controls that hold the others.
- Rule 10 now carries the unresolvable-host hazard as an explicit operational rule.

### Re-verified properly

Opened a throwaway PR off this branch (#579, closed) purely to fire `ai-scan` on
`pull_request: opened` again with the fixed workflow —
[run 33137148739](https://github.com/allxsmith/bestax/actions/runs/33137148739):

| Signal | Before | After |
| --- | --- | --- |
| `Initialized` | 0 | **4** |
| `timed out` | 2 | 0 |
| `Reverted changes` | 1 | 0 |
| `unable to resolve domain` | 2 | 0 |
| `Switching egress-policy` | 0 | 0 |
| Effective policy | `block` (not enforcing) | **`block` (enforcing)** |

Assertion passed, and `Run Claude (security scan)` completed — so the session installs and runs
the CLI under a genuinely live firewall.

The dispatch runs I cited earlier are unaffected: `security-txt-expiry` and the four
`consumer-sbom` legs never had the statsig entry, and their logs do show `Initialized`. Those
verifications stand.

Thanks to Copilot for the catch — the stale `#574` references in its review were already fixed
in 29e3fb3 (the real follow-up is #578).

allxsmith:
Both CodeRabbit findings fixed in 3b7fe83.

**Scope of the downgrade claim.** The `ai-scan` and `claude-repro` headers said `block` degraded "on every run in this repo". That is the same repo-wide overstatement #487's own comments falsified and that I corrected in the issue body earlier in this PR — then reproduced in a workflow comment. It only ever hit the issue- and PR-triggered jobs; `schedule`, `workflow_dispatch` and `release` were never in the untrusted-trigger class and did enforce. Both headers now say so explicitly, and I grepped for the phrasing to be sure there are no others left.

**MD018.** A line in `.github/CLAUDE.md` began with a bare `#487`, which markdownlint reads as an invalid ATX heading. Now "issue #487".

That is three separate review findings on this PR that were all variants of one thing — a claim stated more broadly than the mechanism supports. Fitting, given the subject, and the reason rule 10 now carries the enforcement inventory as a table rather than prose.

allxsmith:
## Review round 2 — three more overstatements, all fixed in b186cac

A self-review pass found three further problems, all in the same family as the earlier ones: a
claim stated more broadly than the mechanism supports. I verified each against source before
changing anything.

### 1. The assertion proves less than the comments claimed (the substantive one)

The comments said "either one alone is a false pass", framed as if the pair were exhaustive.
**A third failure mode passes both lines.** From `step-security/agent` at v0.16.2, `agent.go`:

```go
writeStatus("Initialized")          // :307

for {
  select {
  case <-ctx.Done(): return nil
  case e := <-errc:                 // :313
    WriteLog(...)
    RevertChanges(...)              // :315
```

The agent writes `Initialized` and *then* enters its serve loop. Any runtime error delivered on
`errc` afterwards tears the firewall down — with `agent.status` still reading `Initialized` and
`agent.json` still reading `block`. Both assertion lines pass; nothing is enforcing. Separately,
`refreshDNSEntries` re-resolves every 30s and on failure only logs `failed to insert new
ipaddress in firewall` (`:357`), so an allow-listed host whose IPs rotate can quietly stop being
reachable mid-session.

No sudo and no hostile job code required — just a transient error during the 30-60 minute
session the assertion is meant to bound.

So the honest statement, now in all seven copies and rule 10: **this proves enforcement was armed
at that step; it does not prove the policy stays armed.** There is no step-level check for the
latter — the StepSecurity run report is where `Reverted changes` would show up. I would rather
the contract say that than imply a guarantee the mechanism cannot give, which is the whole
lesson of #487.

### 2. The revert was mis-attributed in all eight copies

The comments said *"the pre-step logs `timed out`, reverts its changes and still exits 0"*. The
pre-step never reverts anything — `grep -c RevertChanges src/setup.ts` is **0**. It logs, prints
`agent.log`, breaks, exits. Every revert call site is in the agent. The substance was right (block
stays in `agent.json`, no firewall, job continues) but someone debugging a red assertion would
have gone looking for a revert in harden-runner's output, where it never appears.

### 3. The pin does not fully determine which agent binary runs

Before installing, the pre-step calls `isTLSEnabled(owner)` — a 3-second probe that **fails
open**: on any error or timeout it logs `Unable to check TLS_STATUS. Defaulting to TLS enabled.`
and returns true, selecting `agent-ebpf` rather than `agent`. Only the latter is source-available
(`step-security/agent-ebpf` contains a README and nothing else), so the paths and status strings
this PR asserts can only be verified against one of the two.

Today this org's endpoint answers 403 (`TLS_NOT_ENABLED` in the run logs), so the verifiable
binary is what runs. But a StepSecurity outage would swap it with no diff and no pin change, and
if its status vocabulary differs, all seven block jobs fail at once — `deploy-worker` and
`sign-sbom` included. Recorded in rule 1 beside the path-rename risk, with the debugging hint.

### Not changed

The review also noted that rule 10's third category omits `on-slop`, `auto-label-claude-prs`,
`close-stale-bestaxbot-prs` and `stale`. None carries `claude-code-action`, so that is an
accuracy gap in an inventory rather than a hidden model-token job — worth tidying, but not worth
another round here. Noted for #578.

allxsmith:
## Review round 3 — two of these would have bitten in production (22b84af)

A cross-file pass found eight more. Two are behavioural and would have caused real damage on
merge; the rest are accuracy. All verified before changing anything.

### 1. Missing `release-assets.githubusercontent.com` — would have taken `ai-scan` dark

`claude-code-action`'s first composite step is `oven-sh/setup-bun`, which fetches bun from a
GitHub **release asset**. The three Claude allowlists listed only `objects.githubusercontent.com`
(the legacy redirect host). Every other block job in this repo lists both.

**My verification run did not catch this, and the reason matters.** From
[run 33137148739](https://github.com/allxsmith/bestax/actions/runs/33137148739):

```
Cache hit for: bun-fR4r1tsFeXfPQkusQwkKD2kGnsE=
Using a cached version of Bun: 1.3.14+0d9b296af
```

`Install Bun` succeeded in 2.1s off the Actions cache and never resolved the release host at all.
So the green run proved the cached path works, not that the allowlist is complete — and these are
exactly the jobs that can only **read** the Actions cache (#487's root cause), so they cannot
repopulate it. A cache eviction, or a bun version bump in a `claude-code-action` update, turns the
restore into a download the allowlist refuses, and `ai-scan` fails on every incoming issue and PR.

That failure would have been near-undiagnosable from the logs, because the assertion step passes.

### 2. The assertion sat before the gate — a fail-open verdict path

`ai-scan`'s fail-closed labeler is guarded by `always() && steps.gate.outputs.run == 'true'`. My
assertion was inserted **before** the gate, so a failing assertion aborted the job before `run`
was ever set — skipping the labeler and leaving the item **both unscanned and unflagged**.

That is a fail-CLOSED check introducing a fail-OPEN path on the verdict, which rule 4 forbids in
so many words, and CLAUDE.md's own note applies: downstream, an unscanned item is
indistinguishable from a clean one, so `claude-repro`, `claude-fix`, `@claude` and `@bestaxbot`
would all have treated those items as clean. Before this PR essentially nothing could fail ahead
of the gate — precisely because harden-runner exits 0 even when it reverts, which is why the
assertion exists.

Moved after the gate. A failed assertion now still applies `needs-security-review`, and the
Claude session — the thing egress-block actually protects — still comes after both.

### 3-8. Accuracy

- **`claude-repro`'s I1 header** called egress-block a "third leg" backstopping the `Write` +
  `publish` chain. `publish` has no harden-runner and no allowlist, so on the one exfil path the
  comment names, the new control contributes **nothing**. Rewritten to say what it does cover.
- **Rule 10's inventory listed `claude-repro` unqualified** though only `author` is hardened.
  `publish` — which runs the sanitizer over attacker-influenced text holding `issues: write` — has
  no egress policy at all. Now per job, and the table's scope is stated rather than implied.
- **`supply-chain`'s "three sibling jobs" counts were stale** (the file has five jobs, not four);
  I rewrote the paragraphs above them and left the numbers.
- **`attach-sbom` gates on `!cancelled()` and only requires `sbom`**, so a failed assertion in
  `sign-sbom` or `consumer-sbom` does not stop a release. Rule 10's "fails the job instead of
  degrading in silence" is true of the job, not the pipeline. Recorded in the comment rather than
  changed — nothing here may fail a release, and that was deliberate.
- **Telemetry setting extended to the six audit-mode sessions.** I had deliberately scoped it to
  the block jobs; that was wrong for a reason I had not considered. The *purpose* of audit on those
  jobs is to measure the endpoint list #578 will use — with telemetry on, that report contains the
  Datadog intake host, and whoever implements #578 would allow-list the host we just decided to
  deny. All nine `claude-code-action` steps now carry it.

Re-verified: 13 pins on one SHA, all workflows parse, and every allow-listed host in the
repository resolves.

allxsmith:
deep-review: This PR changes the egress-enforcement posture of every AI workflow, and three prior
review rounds each found the same class of defect: a security comment or contract claim that
asserts more coverage than the mechanism delivers. Two rounds found defects in the fixes from the
previous round. Please weight that history.

Highest-value areas, roughly in order:

1. **Does the assertion actually prove what the comments now claim?** It is
   `jq -e '.egress_policy == "block"' /home/agent/agent.json` plus
   `grep -qx Initialized /home/agent/agent.status`. The comments claim only that enforcement was
   armed *at that step*. Verify that narrowed claim is itself true, and that no remaining comment
   or `.github/CLAUDE.md` passage overstates it.

2. **Allowlist completeness under real enforcement.** These lists are now load-bearing for the
   first time. `release-assets.githubusercontent.com` was missed because a cached bun restore
   masked the download path. Look for other hosts reachable on a cache-miss or first-run path that
   a green run would not exercise — `claude-code-action`'s composite steps are the place to look.

3. **Step ordering and fail-open/fail-closed interactions.** `ai-scan`'s assertion was moved after
   the gate because placing it earlier skipped the fail-closed labeler and let items go unscanned
   and unflagged. Check the equivalent paths in `claude-repro` (does `cleanup` still unwedge the
   label if `author` dies at the assertion?) and `ai-triage` (its label-removal step is inside the
   job, not a separate `always()` job).

4. **Rule 10's per-job inventory in `.github/CLAUDE.md`.** It has been wrong twice — once omitting
   `bestaxbot-reply`, once implying `claude-repro` was hardened when only `author` is. Verify it
   against the actual files, including the stated scope boundary.

5. **The six jobs newly at `audit`.** They carry no `allowed-endpoints` by design. Confirm that is
   inert rather than harmful, and that nothing in the diff implies they enforce.

Please verify findings against the current head — the branch moved several times during review.

allxsmith:
## Copilot round addressed (a8e7b85) + deep review could not run

### Copilot findings

| Finding | Status |
| --- | --- |
| `claude-repro`'s I1 paragraph is grammatically incomplete and obscures which exfil path is scoped | Fixed — rewritten |
| PR description omits `release-assets.githubusercontent.com` from its rule-2 table, and still says only two entries were added | Fixed — **body rewritten in full** |
| PR description still says `statsig.anthropic.com` is left in place | Fixed — same rewrite |
| PR description's counts are stale (five audit jobs / 12 pins vs six / 13) | Fixed — same rewrite |
| Telemetry setting on the audit jobs contradicts #578 step 4 | Fixed — **#578 step 4 corrected**, since the code is right and the issue was stale |
| Assert a live `agent.service` as well | **Declined, on the record** |

The description had drifted far enough that patching it would have left it misleading, so it is
rewritten to describe the change as it stands, with the review history left in the comments. Rule
2 makes that a security matter rather than housekeeping — the allowlist table was understating
what the diff actually adds.

**On the liveness check**, now proposed twice: it would cover only the fraction of a second
between the agent initializing and the assertion step running. The revert that matters happens
minutes later, mid-session, where no step-level check reaches. In exchange it hard-codes an
internal unit name *and* binds every block job to the TLS-path binary selection — so a
StepSecurity outage would red all seven at once. Declined in rule 10 with that reasoning, so a
third round does not have to rediscover it.

### The deep review could not run on this PR

Label applied, workflow ran, **job went green, and no review was posted**:

```
##[warning] Skipping action due to workflow validation: Workflow validation failed.
The workflow file must exist and have identical content to the version on the
repository's default branch.
```

`claude-code-action` refuses to run when the PR modifies the workflow reviewing it — a guard
against a PR rewriting its own reviewer to leak the token. This PR modifies `claude-review.yml`,
so it disqualifies itself. Not a misconfiguration, and not the `needs-security-review` flag
either (`claude-review.yml` does not gate on it).

Decision: accept no deep review here rather than carve `claude-review.yml` out of the PR. Label
removed, since leaving it on implies a review happened. Documented in the root `CLAUDE.md` beside
the label's docs (526c107) — the green job is exactly what makes it look like it ran.

### Two things for you

1. **This PR carries `needs-security-review`.** `ai-scan` flagged it on open:
   `scan not-clean for #577 (category=injection)`. Almost certainly a false positive — the diff
   and body are *about* prompt injection and quote injection-shaped text and agent logs. But
   while it stands, `@claude`, `@bestaxbot`, `claude-repro` and `claude-fix` will all refuse this
   PR. Removing it is a maintainer call, so I have left it alone.
2. **`ai-scan`'s verification is unaffected** by the validation guard — I checked. Its run had a
   real `EXEC_FILE` and logged `scan clean for #579 — no label`, so that session genuinely ran
   under an enforcing block policy.

allxsmith:
## Review round 4 addressed (1fe363b)

Two findings needed code; three were the description drifting behind the diff again.

### 1. A failed egress assertion spent the scan budget — round 3's fail-open, one step removed

Flagged by both reviewers. `ai-scan`'s gate read the counter and charged it in the same step, and
round 3 moved the assertion *after* that gate so the `always()`-guarded fail-closed labeler stays
reachable. Those two facts together mean every run that failed the assertion spent a slot without
scanning anything.

That is only a cost bug until the slots run out. At `AI_SCAN_DAILY_LIMIT` failures in one UTC day
the gate starts returning `run=false` — and `run=false` skips the labeler. So the tail of that
day's items go **unscanned and unflagged**, which is the exact rule-4 fail-open the assertion's
placement was moved to avoid. A StepSecurity outage reaches every incoming item at once, so
"20 failures in a day" is one bad morning, not a contrived case.

Fixed by splitting the step. The gate decides and hands over `budget_comment_id` / `budget_next` /
`budget_limit` / `budget_date`; a new `Charge the daily scan budget` step after the assertion does
the write. Order is now `gate → assert → charge → checkout → session → labeler`.

Two things I am **not** claiming while doing it:

- **It is not atomic.** Two concurrent runs still read the same count and both write count+1 —
  exactly as they did when read and write shared a step. The split changes *when* the write
  happens, not how it races.
- **The charge still precedes the session**, deliberately. A crashed or timed-out scan must count,
  or a session that reliably dies is free to run all day. The window left is the checkout.

Generalized into rule 10's placement rule, since the shape is not `ai-scan`-specific: a
fail-closed check must not be able to consume the resource that keeps the fail-closed path
reachable.

### 2. Rule 1 justified itself with a failure its own command cannot have

Both reviewers, independently, and they are right. The bullet said moving a pin into
`.github/actions/` would make the verification grep "match zero lines and pass silently". The
documented grep is `grep -rho … .github/` — recursive over `.github/`, so it still matches. Worse,
the note *directly above it* already said so, citing `verified-commit/`'s `actions/github-script`
pin. I contradicted myself within one screen.

The requirement stands; the reason was wrong. It now rests on what is actually true: reading the
job is the whole review signal here, `harden-runner`'s policy and allowlist are argued per job
against the credential in *that* job (rule 2), and a shared wrapper would collapse those separate
allowlists into one — a widening with no `permissions:` diff, which is the hazard the top of that
file exists to name.

CodeRabbit's other half of the finding was also right: rule 1 requires **one SHA per action across
the repo** but only ever documented a grep for the one action you thought to type, so an action
left stale by an earlier bump was invisible to it however many times you ran it. Added the
repo-wide form beside it:

```bash
grep -rhoE "[A-Za-z0-9_.-]+/[A-Za-z0-9_./-]+@[a-f0-9]{40}" .github/ \
  | sort -u | cut -d@ -f1 | uniq -d          # expect NO output
```

Empty on this tree; prints `actions/checkout` when seeded with a divergent pin.

### 3–5. Description

| Finding | Status |
| --- | --- |
| Body's canonical snippet is still `grep -qx Initialized`, missing the `-e` check and the poll | Fixed — replaced with rule 10's actual form, with the three load-bearing shapes named |
| Body repeats the "zero lines" composite-action rationale | Fixed — corrected in the same terms as rule 1 |
| Body predates the gate/charge split | Fixed — documented, including what the verification run does **not** cover |

That last qualifier matters. Run
[33137148739](https://github.com/allxsmith/bestax/actions/runs/33137148739) predates the split, so
the live evidence is for the single-step gate. The passing path is unchanged — the gate's read,
its fail-open handling and the `run` output are byte-identical — and the path that changed is the
*failing* one, which that run did not exercise either way. Writing "verified" without saying so
would have been this PR's own recurring defect for the fifth time.

### Still open, and already answered

The `statsig`, `release-assets`, job-count and `#578` step-4 threads are stale against the current
head: the body rewrite and #578's correction both landed before those comments were re-posted.
#578 step 4 now reads "Already done in #577 — nothing to carry." The `systemctl is-active`
request stays declined, on the record in rule 10, for the reasons given last round.

Local: all 23 workflow files parse, the `ai-scan` step order confirmed off the parsed YAML rather
than by eye, both pin checks clean, `prettier --check` clean.

allxsmith:
## Merged `main` (966d152) — and the conflict was not textual

#582 and #583 landed while this was in review, and between them they restructured **both**
conflicted workflows into multi-job credential-separated designs: `ai-scan` 1 → 3 jobs
(`gate`/`scan`/`label`), `ai-triage` 1 → 4 (`gate`/`triage`/`publish`/`cleanup`). The jobs this
branch reasoned about no longer exist in that shape, so I resolved by taking main's structure
wholesale and re-applying the egress work onto it rather than picking hunks.

### The mechanical half

Main added **five** new harden-runner sites, all on the old pin, and its `scan` job still carried
`statsig.anthropic.com` — the NXDOMAIN host this PR removed precisely because an unresolvable
allow-listed host makes the agent revert its firewall. Merging the other way round would have
reintroduced the bug this PR's first live run found.

Now: 18 sites, all on v2.21.0, 12 `block` jobs and all 12 asserting, 6 `audit` jobs and they are
exactly the intended ones. Verified off the parsed YAML, not by eye:

| workflow | jobs at `block` | asserted |
| --- | --- | --- |
| `ai-scan` | `gate`, `scan`, `label` | ✅ ×3 |
| `ai-triage` | `gate`, `triage`, `publish`, `cleanup` | ✅ ×4 |
| `claude-repro` | `author` | ✅ |
| `deploy-worker`, `security-txt-expiry` | `deploy`, `check` | ✅ ×2 |
| `supply-chain` | `consumer-sbom`, `sign-sbom` | ✅ ×2 |

### The design half — main's premise, which this PR removes

Main made two harden-runner steps `continue-on-error: true`, and said why:

> Nothing is given up: **the policy is not enforced today anyway.**

That is exactly the premise this PR deletes. So the trade is real now, and it is paid for rather
than skipped:

- **`ai-scan`'s `label` and `ai-triage`'s `cleanup` assert AFTER their write.** Those jobs *are*
  the fail-closed path — one guarantees a label is applied, the other that it is removed — so a
  check that can fail ahead of the call defeats the guarantee outright, which is why their
  harden-runner steps are non-blocking in the first place. Asserting afterwards keeps both: the
  write has happened, and a job that ran unprotected still goes red. These are the only two
  assertions in the repo placed after their job's work, and rule 10 now says so.
- **`ai-scan`'s `gate` splits decide → assert → charge.** Same fix as round 4, and it survives the
  restructure for the same reason. Plus one thing the single-job version did not need: the
  assertion records `enforcement=failed` before exiting, and `label`'s guard reads it. Main
  deliberately has `label` fail **open** on gate trouble ("gate trouble is a counter/infra path
  and must fail OPEN") — I did not want to silently reverse that, so ordinary gate failures still
  fail open and *only* an enforcement failure fails closed. Both directions kept rather than one
  traded for the other.
- **`ai-triage`'s `triage` flips `audit` → `block`.** Its two stated preconditions now hold: the
  measured list exists (the audit run produced it) and v2.21.0 removed the downgrade path.

### Review round 5, folded in

| Finding | Status |
| --- | --- |
| `settings:` input is overridable by a branch's `.claude/settings.json` | Fixed — **all nine** session sites now use a step `env:` |
| Rule 10's inventory omits/mislabels jobs; "API-only" understates `supply-chain` | Fixed — inventory rebuilt for the new job graph, group split in two |
| Checklist forbids pins outside `workflows/`, contradicting rule 1 | Fixed |
| `claude-pr-loop`'s `verify` also omits `github_token`, so the deep-review claim is false | Fixed — the real distinction is the trigger's PR context |

The telemetry one is worth calling out as more than hygiene. `settings:` is written to the
**user-level** `~/.claude/settings.json`, which a **project-level** `.claude/settings.json` in the
workspace overrides — and `claude-pr-loop`'s jobs check out PR-branch code, so that file is
attacker-controlled. A PR could have turned telemetry back on and put the Datadog host into the
very audit report #578 depends on. A process environment variable is not overridable that way.

### State

`mergeable=MERGEABLE`. Local: all 23 workflows parse, both pin checks clean, `pnpm format:check`
clean, per-job enforcement table above generated from the parsed YAML.

**Not re-verified live.** Run [33137148739](https://github.com/allxsmith/bestax/actions/runs/33137148739)
predates all of this — it exercised a single-job `ai-scan` that no longer exists. The seven
AI-workflow jobs now carrying assertions have not run in this shape. `ai-scan` and `ai-triage`
will both fire on the next issue or PR opened against the branch, which is the cheapest way to get
that evidence before merge.

allxsmith:
## Review round 6 addressed (edf0759) — seven of nine were one mistake of mine

All nine valid. Worth being blunt about the shape of them, because it is the same defect this PR
has now found six times, and this round I committed it myself.

### The self-inflicted seven

The merge commit moved `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC` from the action's `settings:`
input to a step `env:` at nine sites. I did it by writing **one** rationale and pasting it into
six files. It said:

> …and the jobs in this file check out PR-branch code, so that file is attacker-controlled.

True of `claude-review` (`ref: github.event.pull_request.head.ref`) and `claude-pr-loop`
(`ref: needs.gate.outputs.head_ref`). **False** of `claude-repro` and `claude-implement`, which
check out `github.event.repository.default_branch`, and of `claude` and `bestaxbot-reply`, which
check out with no `ref:` at all. The same paste also put an audit/measurement rationale on
`claude-repro`'s `author` — a **block** job that #578 does not cover.

So a security comment claimed a threat model its job does not have, in four files, because I
templated it. Split into two variants: the jobs that really do check out PR-branch code keep the
attacker-controlled framing; the rest say plainly that nothing in their workspace is
attacker-controlled today, and that `env:` is used so the control does not *depend* on that —
since a checkout change would silently move them into the other class.

### The one that is not comment-only in kind

`ai-scan`'s new budget comment claimed the concurrency group serializes the counter. **It does
not.** This workflow's group is per item:

```yaml
concurrency:
  group: ai-scan-${{ github.event.issue.number || github.event.pull_request.number }}
```

chosen deliberately so a burst of opened items cannot evict a pending scan — the file's own header
says so, and the *global* single-flight group is `ai-triage`'s, whose header says the counter's
correctness depends on it. I read the right sentence in the wrong file.

Two items opened together therefore do race the counter, and splitting read from write **widens
that window** by the assertion's duration (up to ~30s on the polling path). Both facts are now in
the comment, with why the race is accepted: losing a count over-scans, which is the safe
direction for a security control.

### Contradictions the merge left behind

| Finding | Fix |
| --- | --- |
| Rule 2's `ai-scan` row still said egress is not enforced and the allowlist is the only control | Now: enforced, and **complementary** — it bounds where data goes; `api.github.com` stays allow-listed |
| Rule 10's `-e` bullet said a vendor outage fails "all seven jobs" | Twelve block jobs now; noted not to hand-maintain the count |
| `ai-triage`'s `cleanup` still had a rule-10 comment saying block does not enforce | Removed — it sat one paragraph above the new one saying it does |
| `claude-repro` claimed `release-assets` is "absent from no other block job" | Scoped to block jobs that run `claude-code-action`; the API-only ones correctly omit it |
| Root `CLAUDE.md` — I over-corrected last round | `claude-pr-loop` also fires on `pull_request_review` and its gate **can** select `verify` there, so `verify` is not categorically outside PR context. The test is "omits `github_token` **and** can run in a PR context", not the workflow name |

Local: all 23 workflows parse, 9 session sites all on step `env:` with no `settings:` input
remaining, 18 pins on one SHA, 12 block jobs all asserting, `pnpm format:check` clean — all checked
off the parsed YAML rather than by eye, which is how the per-job claims should have been checked
the first time.

allxsmith:
## Round 7 (a648268) — the fail-open I closed last round started one step too late

Copilot is right, and this is the better version of the finding I acted on in round 6.

`harden-runner` is **step 0** of `ai-scan`'s `gate`. If the *action itself* fails — download,
bootstrap, a StepSecurity outage that errors rather than no-opping — the job dies before the gate
step emits anything. `label` keys off `needs.gate.outputs.run`, so an empty `run` skips the
fail-closed labeler and the item goes unscanned **and** unflagged. The assertion cannot catch it,
because the assertion never runs.

So round 6 made an *assertion* failure fail closed while leaving the *action* failure fail open —
the same hole, one step earlier. Fixed with `continue-on-error: true` on that harden-runner, which
is what `label` and `ai-triage`'s `cleanup` already do for the same family of reason. Non-blocking
is not "unprotected and ignored": the gate step gets to emit `run`, then the assertion finds no
`agent.json`, records `enforcement=failed` and fails the job — which is exactly what activates the
labeler. Cost: one authenticated **read** of the tracking issue runs with no firewall. `label`
already accepts that trade for a **write**.

**I checked the rest of the graph for the same shape rather than fixing only the reported
instance:**

- `ai-triage`'s `cleanup` keys off `gate.result != 'skipped'`, so a dead gate still unwedges the
  label — no hole.
- `ai-triage`'s `publish` keys off `run`, and that is correct: a dead gate means there is nothing
  to publish.
- `ai-scan`'s `scan` is already fail-closed — a failure there is `failure`, not `skipped`, so
  `label` runs and the empty verdict hits the enum default.

Rule 10 now carries the general trap — **a downstream `always()` job keying off an upstream's
outputs rather than only its `result`** — plus the five-path failure matrix, because three of
those paths must flag and two must not, and they look alike.

### Also fixed: the body's verification counts

It still said "13 pins on one SHA", which was true before `main` merged and added five
harden-runner sites. Now read off the parsed YAML: **18** uses on one SHA, **12** block jobs all
asserting, **6** audit, **9** session sites on step `env:`. That section also now leads with the
live-evidence gap rather than burying it — run 33137148739 exercised a single-job `ai-scan` that
no longer exists, and none of the failing paths this PR reworked have ever run.

allxsmith:
## Round 8 (26ba6e4) — both findings valid, and the inventory error was mine twice over

### 1. "API-only" still covered jobs that execute repository code

Third time this list has put a code-executing job in the harmless-looking group — and the previous
two *fixes* are what introduced this instance. Verified per job this time:

| Job | What it actually does |
| --- | --- |
| `auto-close-duplicates` | checkout + `scripts/auto-close-duplicates.mjs` |
| `claude-repro` → `publish` | checkout + `scripts/sanitize-repro-draft.mjs`, over attacker-influenced text, holding `issues: write` |
| `claude-pr-loop` → `handoff` | checkout + `scripts/handoff-message.mjs` |
| `supply-chain` → `sbom` | installs the monorepo, runs SBOM generators |
| `supply-chain` → `verify-provenance` | installs published packages, runs verification scripts |

And one moves the *other* way: `attach-sbom` downloads and uploads release artifacts and runs
nothing, so last round's fix was wrong about it in the opposite direction. `claude-repro`'s
`prepare`/`cleanup` and `claude-pr-loop`'s `sweep`/`gate`/`halt` are genuinely API-only, confirmed.

The derivation note now keys off `actions/checkout@` — no checkout, no repo code — and says to
**read** the jobs rather than grep for `node`. That matters concretely: `claude-pr-loop`'s `gate`
contains a `node -e` inside a *comment*, which is exactly the shape that would keep this list wrong
a fourth time. My own audit script flagged it before I looked.

### 2. Both after-write assertions claimed a write that may not have happened

They run under `always()`. So `ai-scan`'s fires on a **clean** verdict — where there is no label to
apply — and after a **failed** labeling call, while its error read:

> The label above was still applied

Which is the same overstatement this PR keeps finding, now inside the error text meant to explain a
security failure. Both messages now say only that the check runs *after* that step and did not
prevent it, and point the reader at that step's own result. Fixed identically in `ai-triage`'s
`cleanup`.

Local: 23 workflows parse, 18 pins on one SHA, 12 block jobs all asserting, 9 session sites on step
`env:`, `pnpm format:check` clean.

**Still unchanged and still the thing I would not merge without:** none of this has run. Seven of
the eight review rounds on this PR found a real defect, and the last four were defects in the
fixes — all of them in claims about mechanisms rather than in the mechanisms themselves, which is
precisely the class that static checking does not catch. Opening a throwaway issue against this
branch exercises `ai-scan` and `ai-triage` end to end.

Issue #290, opened 2026-07-11
AI triage budget tracker (machine-managed — keep open)


allxsmith:
This issue is the anchor for the AI-triage daily budget counter (see #289).

`ai-triage.yml`'s gate step maintains a single comment here of the form:

```
<!-- ai-triage-budget date=YYYY-MM-DD count=N -->
```

- The comment is edited in place by `github-actions[bot]`; the probe is author-filtered, so other comments here can't forge or reset the counter.
- A new UTC date resets the count; an unparsable marker fails **closed** (treated as at-cap).
- The daily limit comes from the repo variable `AI_TRIAGE_DAILY_LIMIT` (unset ⇒ 10). Label-triggered runs bypass the budget.

**Keep this issue open** — closing it doesn't break the workflow (comments still work on closed issues), but open keeps it visible. Human comments here are harmless and ignored by the machinery.

PR #580, opened 2026-08-28
ci: run the ai-scan model session with a read-only token

### LINKED ISSUE #455: [Security] ai-scan's Claude session runs with a write-scoped GITHUB_TOKEN

allxsmith:
Fixes #455.

## This is a security change (`.github/CLAUDE.md` rule 2)

The `--allowedTools` allowlist is **unchanged**. What changes is that it is no longer the only thing standing behind it: the model session's job now holds no write scope at all.

Credential per job, after the split:

| Job | Token | Scope | Runs repo code? |
| --- | --- | --- | --- |
| `gate` | `GITHUB_TOKEN` | `issues: write` (budget marker) | no |
| `scan` | `GITHUB_TOKEN` | `contents`/`issues`/`pull-requests` **read** | yes (the verdict parser) |
| `label` | `GITHUB_TOKEN` | `issues: write` (the labels API covers PRs too) | no |

Two properties worth stating plainly, because neither held before:

- The Claude session's job has **no write permission of any kind**.
- **No job holding a write token runs any checked-out repository code.** `gate` and `label` are dependency-free shell; the parser runs in the read-only `scan` job.

## Why the parser stays in the `scan` job

`execution_file` is a runner-local path, so it cannot cross a job boundary, and uploading it as an artifact would publish the scanner's reasoning to anyone with read access — precisely the evasion oracle `show_full_output: false` exists to prevent. So `scripts/parse-scan-verdict.mjs` (unchanged) runs in `scan`, and only its coarse enum (`clean` / `flagged <category>`) becomes a job output. The labeler's `case` enum, its `flagged`/`other` defaults, the step-summary lines, and the `addLabels` call are moved **verbatim** — the only edit is reading `$SCAN_VERDICT` instead of invoking node.

The `gate` shell is byte-identical to the original (verified by diff).

## Fail directions

| Situation | Result | Direction |
| --- | --- | --- |
| Scan job crashes, or output plumbing drops the value | `SCAN_VERDICT` empty → enum default → **flagged** | closed, preserved |
| Gate errors on the budget read | `run` never `true` → scan and label skip → unscanned, unlabeled | open, preserved |
| Mode off / fork head | gate's `if:` false → nothing runs | unchanged |
| Scan **cancelled** | label runs under `always()` → **flagged** | closed |
| **Gate** fails after writing run=true | scan skips, label skips | open |
| Kill switch flipped off before a re-run | scan skips, label skips | no spurious flag |

An earlier revision of this PR claimed the cancelled row was a behavior change introduced by the split. That was wrong — the pre-split labeler was a step carrying `always()`, which also runs through cancellation — and the claim has been removed. Cancelling a run does not prevent the flag, before or after; that is now stated as a property rather than a delta.

## One deviation from the issue text

The issue sketches `scan` as `contents: read`. Declaring only that sets every other scope to `none`, so the session's `gh issue view` / `gh pr diff` would work solely because this repository is public — an undocumented dependency, invisible in review, that breaks under a private fork. The job therefore declares `issues: read` and `pull-requests: read` as well. That reads as *more* permissions than the issue asked for while being strictly read-only, and it satisfies the acceptance box ("no write permission of any kind").

## Review round (commit 2)

A review of the first commit found three real defects, all fixed in `a551303`:

1. **Kill-switch bypass.** `AI_LOOP_ENABLED` / `AI_SCAN_MODE` were only on `gate`. "Re-run failed jobs" replays a green gate's cached `run=true` rather than re-running it, so a billed session could start after a maintainer set `AI_SCAN_MODE=off`. Both variables are now repeated on `scan`, the job that spends the usage (rule 3).
2. **`label` could flag an unscanned item.** Its condition read only `needs.gate.outputs.run`, so a gate that wrote `run=true` and then failed would skip `scan` while `label` ran against an empty verdict and applied `needs-security-review` — fail-closed on a path rule 4 mandates fail-open. It now also requires `needs.gate.result == 'success'` and `needs.scan.result != 'skipped'`.
3. **harden-runner could lose a flag.** Sitting ahead of the labeling call, a failed action download would discard a computed `flagged` verdict. It is `continue-on-error: true` now; the policy is not enforced anyway (#487).

Plus several claims scoped back to what the mechanism delivers: the `.github/CLAUDE.md` row no longer says the ai-scan allowlist holds back "nothing" (the job still runs Bash beside `CLAUDE_CODE_OAUTH_TOKEN`, so by I1 the allowlist is still the only control on *that* credential — repository write is what the split removed), `SECURITY.md` no longer says the label job is one the session "cannot reach" (the session's verdict is that job's only input), and `parse-scan-verdict.mjs` and its test no longer describe the pre-split topology. Also dropped the unused `pull-requests: write` from `label`, and restored the repo-uniform egress endpoint list on the two new jobs.

## Empirically retired risk

A reviewer flagged that read-only permissions plus `claude-code-action` on a `pull_request` trigger had no precedent in this repo, with a repo-wide false-positive as the failure mode. **This PR's own ai-scan run settles it**: `gate` → `scan` → `label` all passed on PR #580 itself, and the label job logged `scan clean for #580 — no label`. The action works at read scope on this trigger.

## Known follow-ups (not blocking)

harden-runner is now declared three times in this one file; the gate's budget shell is a near-verbatim copy of `ai-triage.yml`'s against the same tracking issue with opposite fail directions and no test on either; the verdict category enum lives in three places; and `scan` does a full checkout for one 160-line script that `sparse-checkout` would trim.

## Acceptance

- [x] The Claude session step runs in a job with no `write` permission of any kind
- [x] The fail-closed labeler behaviour is unchanged (enum and defaults byte-identical; crashed / unparsable / `is_error` still flag, now via an empty verdict output)
- [x] The budget accounting on the tracking issue is unchanged (shell moved verbatim)
- [x] The header comment's description of the token scope matches the new structure

## Review checklist

- Allowlist growth: none. `--allowedTools` and `--disallowedTools` are untouched.
- Model session gaining execute/fetch/network: no.
- Model text reaching a comment body: no comments are posted by this workflow.
- Posting identity: `GITHUB_TOKEN` throughout; no PAT anywhere in this file.
- New repository variable: none.
- Action SHAs: unchanged, still one SHA per action repo-wide (`grep -rho "actions/checkout@[a-f0-9]\{40\}" .github/workflows/ | sort | uniq -c` → one line).
- Verdict path fails closed; counter path fails open. Both audited above.
- Egress: the two new jobs declare `block` per rule 10, and the header still states plainly that nothing in this repo enforces egress until #487 closes. No comment cites egress as a control.

## Verification

`pnpm all` green locally on both commits. `parse-scan-verdict.mjs` and its test sibling have header-only edits (their topology description was stale after the split); no behavior changed, and all 453 script tests pass.

Worth watching on the first live run (needs `AI_SCAN_MODE=on`): the three jobs chaining, the gate charging the marker on #290, and the label job logging `scan clean … — no label`.

## Summary by CodeRabbit

* **Security Improvements**
  * Isolated scanning, access checks, and security labeling into separate workflow steps with limited permissions.
  * Scanning sessions can no longer write to the repository or apply labels.
  * Improved protection of credentials and repository access during automated scans.

* **Documentation**
  * Updated security guidance to explain the revised scanning and labeling process.
  * Clarified verdict handling and permission boundaries.


allxsmith:
Both of Copilot's comments were correct, and both were stale claims my own earlier fixes left behind. Addressed in ec056da.

**`ai-scan.yml:50`** — the header's job topology still described `label` as holding `issues + pull-requests: write` after I cut the scope to `issues` only (PR labels go through the issues API). Header and job now agree.

**`ai-development.md:105`** — "no channel to post or leak anything" does not hold, and the same PR says so two files over: the scan job still carries `CLAUDE_CODE_OAUTH_TOKEN` beside Bash, and egress is not enforced (#487), so the allowlist remains the only guard on that credential. Narrowed to what the credential split actually delivers — no write to this repository. Made the same correction in `SECURITY.md`, which said "cannot post or act"; "act" was doing the same overreach.

This is the second round of the same class of error in this PR — a security comment claiming more than the mechanism delivers — so it is worth naming the pattern: the split genuinely removed *repository write* from the session, and each time I have overstated it, I have written it as though it removed *the boundary*. It did not. Rule 2's table now spells out which credential the allowlist still guards.

PR #340, opened 2026-07-21
ci: fail AI triage loudly when the session bails before posting (#338)

allxsmith:
# Pull Request

## Description

Fixes the silent no-op in AI auto-triage: a Claude Code CLI update made Task sub-agents background-by-default, so the triage session fanned out its search agents, said it would "wait for them to finish", and ended its turn — which terminates the headless session with no comment posted. The result record is a clean success (`is_error: false, stop_reason: end_turn`), so the existing watchdog stayed green (a new variant of the #317 silent-no-op shape; evidence in #338 from runs 29686174473 / issue #330 and 29625394616 / issue #327).

Changes:

- **`.github/workflows/ai-triage.yml`** — the session prompt now requires every Task call to pass `run_in_background: false` and forbids ending the turn while any sub-agent is pending; it also requires a machine-checkable final message of only `TRIAGE-RESULT: <command> commented|refreshed|skipped (<reason>)` lines. The "Fail on silent session error" watchdog jq additionally requires that sentinel shape in the final result text, so any mid-task bailout — this one or a future runtime behavior shift — fails the job loudly instead of showing green. Header comment documents the #338 failure shape next to the existing #317 note.
- **`.claude/commands/triage-dedupe.md`, `triage-find-issues.md`, `triage-find-duplicate-prs.md`** — same synchronous fan-out rule, plus search hardening from the same logs: one prefix-allowlisted `gh` command per Bash call (shell loops/`echo` prefixes were being permission-denied) and a 6-search cap per agent (bursts were 403ing on the shared API rate limit).

- [ ] bulma-ui (`@allxsmith/bestax-bulma`)
- [ ] create-bestax (`create-bestax`)
- [ ] docs (`@allxsmith/bestax-docs`)
- [x] Other (please specify): CI workflow (`.github/workflows/ai-triage.yml`) + triage command files (`.claude/commands/`)

## Related Issue(s)

Fixes #338

Related to #317, #330, #327

## Type of Change

- [x] Bug fix
- [x] Build tooling

## Checklist

- [x] My code follows the project style guidelines
- [x] I have performed a self-review of my code
- [ ] I have added tests that prove my fix is effective or that my feature works — no jest surface; the watchdog jq was verified against 8 execution-file shapes (real #330 bailout text, #317 `is_error` shape, missing/empty/malformed records, single-line / multi-line / JSON-array sentinel forms)
- [x] I have added/updated documentation as needed (workflow header comment)
[...]

PR #581, opened 2026-08-28
ci: publish AI triage comments from a validated payload, not model free text

### LINKED ISSUE #457: Publish AI triage comments from a structured payload, not model free text
Raised by CodeRabbit on #361. Nothing here is exploitable today — this is about restoring a security property by construction rather than by enumeration.

## What changed in #361

Triage comments used to be posted with `GITHUB_TOKEN`, authored by `github-actions[bot]`. #361 switched that path to `AI_LOOP_PAT` so comments are credited to the bestaxbot machine identity (robobun-style), which was the point of the change.

The side effect: **`GITHUB_TOKEN` comments cannot emit workflow events; PAT-authored comments can.** So the triage path went from an identity that was structurally incapable of re-triggering a workflow to one that is capable, and is held back only by each consumer remembering to exclude it.

## Why it matters

This repo's second invariant (`.github/CLAUDE.md`, I2) is: *no untrusted or model-authored free text reaches a re-trigger-capable identity.*

`claude-repro.yml` honors it — the session writes a draft, a deterministic step sanitizes it, `GITHUB_TOKEN` posts it. `ai-triage.yml` does not: the model's free text goes out under a PAT-authored identity, and that session ingests untrusted issue/PR text. One workflow follows the invariant, its sibling does not.

## Current mitigation (verified, holds)

Every comment-triggered workflow independently gates bestaxbot out:

- `claude.yml` and `bestaxbot-reply.yml` — `sender.type == 'User' && sender.login != 'bestaxbot'`. The login test is the load-bearing one: bestaxbot is a machine *User* account, so the type test alone does not exclude it.
- `claude-pr-loop.yml` — re-fires only on `coderabbitai[bot]` reviews.
- `ai-triage.yml` / `ai-scan.yml` / `claude-implement.yml` / `claude-repro.yml` — trigger on `opened`/`labeled`, not on comments.

That is defense by enumeration. It is correct today, and `.github/CLAUDE.md` rule 8 tells the next workflow author to keep it correct — but any future comment-triggered workflow that forgets the exclusion reopens the class silently.

## Proposed fix

Keep bestaxbot as the comment author. Stop letting model free text be the payload.

- [ ] Define a schema for the triage result (duplicate candidates, issue numbers, confidence, the marker) — the session emits structured data, not prose
- [ ] Add a deterministic render + publish step that builds the comment body from that schema and posts it with the PAT
- [ ] Drop `Bash(gh pr comment:*)` and `Bash(gh issue comment:*)` from the triage allowlist, leaving it GET-only (`.github/CLAUDE.md` rule 2 — narrowing an allowlist on a credential-holding session is the preferred direction)
- [ ] Unit-test the renderer, including that attacker-controlled strings in the payload cannot produce `@mentions` or machine markers (`.github/CLAUDE.md` rule 9 — logic worth testing does not live in YAML)

The alternative CodeRabbit suggested first — publishing via `GITHUB_TOKEN` — is rejected: it would revert comments to `github-actions[bot]` and undo #361's purpose.

## Related

- #361 (introduced the PAT-authored triage path, plus the sender exclusions that bound it)
- #454, #455, #456
- `.github/CLAUDE.md` — invariant I2, rules 2, 8, and 9

Refs #362

allxsmith:
Fixes #457. Refs #361, #340, #317, #338, #312.

The triage session stops posting. It reports `TRIAGE-PAYLOAD:` lines; a fail-closed validator parses them and renders the comment bodies from a trusted skeleton; a separate step upserts those bodies by marker as bestaxbot. This is the shape `claude-repro.yml` already uses, applied to the sibling that did not have it.

## This is a security change (`.github/CLAUDE.md` rule 2)

Two parts, both narrowing:

**1. The allowlist shrinks.** `Bash(gh pr comment:*)` and `Bash(gh issue comment:*)` are removed, leaving GET-only `gh` reads plus `Task`. Nothing is added.

**2. `AI_LOOP_PAT` leaves the Claude step.** The session now gets the job's `GITHUB_TOKEN`; the PAT appears in exactly one place, the publish step's `env`.

The second is the part worth reviewing, and the argument for it is *not* the re-trigger one:

- `AI_LOOP_PAT` is documented as full repo write and is the same durable credential `claude-implement.yml` pushes with and `claude-pr-loop.yml` uses across four steps. Leaking it means revoking the whole loop's identity. `GITHUB_TOKEN` expires with the job and is scoped to this repo.
- This session has `Bash` and `Task`, ingests attacker-controlled issue and PR text by design, and runs with egress unenforced (#487).
- **Reading an environment variable is not a write**, so the tool allowlist never guarded that credential at all. That is invariant I1's own reasoning applied to a credential I1 does not currently name.

The re-trigger and full-repo-write arguments are real but secondary: both require the allowlist to fail first, and after this change the session has no comment tool regardless of which token it holds.

### What does *not* change

Published comments are identical in provenance: **bestaxbot authors them, via the PAT, in the publish step**, so they still emit `issue_comment` events exactly as before.

- Copilot is untouched. Its review is requested by `claude-implement.yml` (`gh pr edit --add-reviewer "@copilot"`), which already runs on `GITHUB_TOKEN`, and its auto-review fires on PR open/push — never on triage comments.
- Nothing in this repo consumes triage-comment events anyway: `bestaxbot-reply.yml` and `claude.yml` both gate on `sender.login != 'bestaxbot'`, and `claude-pr-loop.yml` fires on CodeRabbit reviews.

### Watch item for the first live run

`ai-scan.yml` proves `gh issue view` / `gh pr view` / `gh pr diff` work on the same action SHA with `GITHUB_TOKEN`, but **`gh search` has no in-repo precedent** — triage is the only session that searches. Same 30 req/min class, public repo, and the command files already cap 6 searches per agent, so this should be a non-event. If it 403s, the revert is one field: restore `github_token: ${{ secrets.AI_LOOP_PAT }}` on the Claude step. Everything else, publish step included, is unaffected.

## The renderer

`scripts/render-triage-comments.mjs` replaces the workflow's inline sentinel jq (rule 9) and reuses `parse-scan-verdict.mjs`'s exported helpers rather than growing the YAML.

- **Anchoring**: payload lines must be the last non-empty lines of the last result record, one per expected command, with the whole-message count exactly right. An echoed copy mid-message makes the count N+1 and fails. Whitespace-only lines still count as lines. Leading narration is tolerated (the run-29794090279 lesson).
- **Closed schema**: exact key sets, so an unknown key — `__proto__` included — rejects. Numbers must be safe integers in range and not the item being triaged; `duplicateOf` must name one of the listed duplicates; titles and reasons are byte-capped and must contain no control characters, which is the structural kill for newline smuggling.
- **Sanitization runs on fields, before interpolation — never on the assembled body.** This is the deliberate inversion from `sanitize-repro-draft.mjs`, and the header says why: a whole-body pass would defang the skeleton's own `<!-- ai-triage:dedupe -->` marker and `Duplicate of #N` line, which `auto-close-duplicates.mjs` consumes, and silently break auto-close. Every `@` is encoded, not just the three re-trigger handles, because bestaxbot authors these comments so any live mention pings a person.
- **Fails closed**: any defect exits non-zero with empty stdout. The wrapper runs `set -euo pipefail` with no fallback plus an exit-0-empty-stdout guard (the claude-repro precedent). `stderr` carries fixed strings only, never payload-derived text.

The test sibling pins the rendered bodies byte-for-byte, imports the consumer's own `MARKER`/`DUPLICATE_RE` so the coupling cannot drift, and pre-asserts its hostile fixtures are actually hostile before checking they are neutralized.

## Behavior change worth noting

The empty-result trigger policy moves out of the prompt and into `buildEnvelope`: dedupe reports its empty result on any trigger; the two PR commands stay silent on `opened` and post on `labeled`. That rule used to depend on the model remembering it. It is now structural — a labeled rerun cannot be silently skipped.

Incidentally this also fixes a real cosmetic bug: existing triage comments show the literal template text `**Related** (optional, at most 3)`, because the model copied the parenthetical out of the command file. The renderer emits `**Related**`.

## Expected self-inflicted failure (rule 9 bootstrap gap)

This PR's own auto-triage run did **not** exercise the gap: the gate skipped it with `PR body already links an issue (Fixes/Closes) — skipping`, so no session started ([run 33137204485](https://github.com/allxsmith/bestax/actions/runs/33137204485)). The gap is still real for a manual `ai-triage` label run on this branch — the job checks out the default branch, which does not yet contain `scripts/render-triage-comments.mjs`, so the Validate step would exit `MODULE_NOT_FOUND` and fail the job.

That is the documented cost of extraction and is **not** to be fixed by checking out PR head — running PR-authored code in a credential-holding job is the thing that pin prevents. The label-removal step is `always()`, so the `ai-triage` button would not wedge.

## Merge order

Please merge #580 (#455) first. Both PRs edit `.github/CLAUDE.md` rule 2's table — that one changes the `ai-scan` row and the table's intro line, this one changes the `ai-triage` row — so this branch will want a rebase afterward.

## Review checklist

- Allowlist growth: **none, it shrinks by two entries**.
- Model session gaining execute/fetch/network: no. `--disallowedTools` unchanged.
- Model text reaching a comment body: only through the validated payload and `sanitizeField`, into a trusted skeleton.
- Posting identity: bestaxbot's PAT, in the deterministic publish step only. The Claude step now holds the job `GITHUB_TOKEN` (still write-scoped for the gate and label removal — the allowlist is the control there, and the header says so).
- New repository variable: none. `AI_TRIAGE_AUTOCLOSE` moves from the prompt to the render step; the model is no longer told it.
- Action SHAs: unchanged, still on the repo-wide pins.
- Verdict path fails closed: yes, and its matrix is now unit-tested rather than inline jq.
- Comment upsert scoped by marker + (bestaxbot OR Bot-type author), never `--edit-last` (rule 6).

## Verification

`pnpm all` green. 500 script tests pass, 47 of them new. The two new workflow steps were also run end-to-end locally against fixture execution files with `gh` stubbed: full dedupe render with the auto-close notice, PR-mode `opened` silence vs `labeled` posting on identical payloads, trailing-narration rejection with exit 1, and the missing-execution-file tolerance path.

## Summary by CodeRabbit

* **New Features**
  * Improved automated triage comment publishing with validated, consistently formatted results.
  * Added safer handling for duplicate, issue-linking, and duplicate pull request triage outcomes.
  * Ensured triage comments are updated reliably without selecting human-authored comments.

* **Bug Fixes**
  * Prevented untrusted content from spoofing triage result markers.
  * Added fail-closed behavior for malformed or unsafe triage results.

* **Documentation**
  * Clarified triage behavior, comment authorship, and local versus automated runs.

### DIFF
--- a/.claude/commands/triage-dedupe.md
+++ b/.claude/commands/triage-dedupe.md
@@ -9,8 +9,23 @@ Find likely duplicates of the target issue and post EXACTLY ONE comment (or
 nothing — see pre-checks). Context comes from the caller (the ai-triage
 workflow prompt, or the user when run locally): `REPO` (owner/repo), `NUMBER`
 (the target issue), `TRIGGER` (`opened` or `labeled`; assume `labeled` when
-run locally), and `AUTOCLOSE` (`active` or `off`; assume `off` locally). If
-`NUMBER` is missing, ask — never guess.
+run locally), and `AUTOCLOSE` (`active` or `off`; assume `off` locally — in
+CI it is not passed to the session at all, see below). If `NUMBER` is
+missing, ask — never guess.
+
+## In CI (ai-triage.yml)
+
+The workflow session must NOT post or edit comments — its tool allowlist is
+GET-only and every `gh … comment` attempt is denied. Run the pre-checks,
+search, and filter pass below unchanged, then, instead of the Post section,
+report the result as the `TRIAGE-PAYLOAD:` line the workflow prompt specifies:
+`"action":"skip"` for a pre-check exit, `"action":"post"` with an empty
+`duplicates` array when nothing credible remains, `"action":"post"` with the
+entries otherwise. A deterministic step
+(`scripts/render-triage-comments.mjs`) renders the format below from that
+payload and upserts the comment by marker as bestaxbot. Everything in this
+file about `gh … comment`, `--edit-last`, and the AUTOCLOSE notice therefore
+applies to LOCAL runs only.
 
 ## Pre-checks (in order; each exit is SILENT — post nothing)
 
@@ -23,7 +38,8 @@ run locally), and `AUTOCLOSE` (`active` or `off`; assume `off` locally). If
    claude[bot].)
    - `TRIGGER=opened` and marker present → stop (already triaged).
    - `TRIGGER=labeled` and marker present → continue; at the end REFRESH
-     that comment instead of posting a new one. Use
+     that comment instead of posting a new one. In CI the publish step does
+     that refresh by marker, so just report fresh findings; locally, use
      `gh issue comment NUMBER --repo REPO --edit-last --body ...` ONLY when
      your most recent comment on the issue is itself the
      `<!-- ai-triage:dedupe -->` comment: `--edit-last` selects by author,
@@ -78,8 +94,10 @@ Read the top candidates with `gh issue view` when unsure.
 
 ## Post ONE comment
 
-Via `gh issue comment NUMBER --repo REPO --body ...` (or the refresh path
-from pre-check 2), formatted exactly like this:
+Local runs post this via `gh issue comment NUMBER --repo REPO --body ...` (or
+the refresh path from pre-check 2). In CI the renderer produces this exact
+format from your payload — `scripts/render-triage-comments.mjs` is the source
+of truth for the posted body, and this section documents what it emits:
 
 ```markdown
 ### AI triage
@@ -100,16 +118,22 @@ Duplicate of #N
 
 - The line `Duplicate of #N` — with N the single best duplicate — must use
   THIS EXACT FORMAT on its own line: the auto-close cron machine-parses it.
-  Include it only when you found at least one credible duplicate.
+  Include it only when you found at least one credible duplicate. In CI the
+  renderer emits this line from the payload's `duplicateOf` — report that
+  field, and never put the line inside a payload string.
 - If `AUTOCLOSE=active`, add this sentence directly after the
   `Duplicate of #N` line: "This issue may be auto-closed in 14 days unless
-  someone objects (comment or 👎)." If `AUTOCLOSE=off`, omit it.
+  someone objects (comment or 👎)." If `AUTOCLOSE=off`, omit it. In CI the
+  renderer decides this from the workflow's own AUTOCLOSE value; the session
+  is not told it.
 - The marker `<!-- ai-triage:dedupe -->` goes on its own line at the end.
 - No credible duplicates → post "No duplicates found." under the heading,
   plus the marker, WITHOUT any `Duplicate of #N` line or auto-close notice.
+  In CI that is an empty `duplicates` array, not a skip.
 
 ## Hard rules
 
 Never apply or remove labels; never close, merge, push, or resolve
 anything; never write "@claude" or "@coderabbitai" in any text; post at
-most one comment and nothing else.
+most one comment and nothing else — and in CI, post none: report the
+payload.
--- a/.claude/commands/triage-find-duplicate-prs.md
+++ b/.claude/commands/triage-find-duplicate-prs.md
@@ -10,6 +10,19 @@ default: when nothing credible turns up, post NOTHING. Context from the
 caller: `REPO`, `NUMBER` (the target PR), `TRIGGER` (`opened`/`labeled`;
 assume `labeled` locally). If `NUMBER` is missing, ask.
 
+## In CI (ai-triage.yml)
+
+The workflow session must NOT post or edit comments — its tool allowlist is
+GET-only and every `gh … comment` attempt is denied. Run the pre-checks,
+search, and filter pass below unchanged, then, instead of the Post section,
+report the result as the `TRIAGE-PAYLOAD:` line the workflow prompt
+specifies: `"action":"skip"` for a pre-check exit, `"action":"post"` with an
+empty `prs` array when nothing credible remains, `"action":"post"` with the
+entries otherwise. A deterministic step
+(`scripts/render-triage-comments.mjs`) renders the format below from that
+payload and upserts the comment by marker as bestaxbot. Everything in this
+file about `gh … comment` and `--edit-last` applies to LOCAL runs only.
+
 ## Pre-checks (each exit is SILENT)
 
 1. `gh pr view NUMBER --repo REPO --json state,title,body,files,comments` —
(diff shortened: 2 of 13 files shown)


allxsmith:
Thanks — the rule 9 point on the publish shell was right, and acting on it surfaced a second problem the comment did not mention.

Extracted to `scripts/pick-triage-upsert.mjs` (+ test sibling, 18 tests) in 07161a6. The extraction is not a like-for-like move: the jq was a **reimplementation of logic `auto-close-duplicates.mjs` already exports**, and it had drifted. The jq tested `login == "bestaxbot" or type == "Bot"`; the shared `isAutomationAuthor` also covers a `[bot]`-suffixed login typed as `User`. The new script imports that helper, so the two consumers of these markers cannot drift again — which is a stronger outcome than testing the jq in place would have been.

It also imports `parse-scan-verdict.mjs`'s `parseExecutionRecords` to flatten the one-array-per-line stream `gh api --paginate` emits when `--jq` is applied per page. That shape is what the old `tail -n 1` survived only by accident: a marker comment on any page *but the last* was found only when the last page happened to contain one too. There is now a test for exactly that.

On scope, I stopped short of your suggestion to cover "refresh/post and `disposition: none` behavior with a stubbed API". The in-repo precedent argues against it: `auto-close-duplicates.mjs` does all its own HTTP with `fetch` and its test sibling stubs none of it — it tests the pure helpers only. What remains inline here is a loop, a disposition check, and PATCH-vs-POST, which is the size `claude-repro.yml`'s publish step keeps inline and which rule 9 explicitly leaves there. If you think the HTTP half should be tested too, that is a change to the repo's convention rather than to this PR, and worth its own issue.

allxsmith:
Superseded by #582, which merged as 4361064. Both carry `Fixes #457`, so closing this one to avoid two implementations of the same issue.

Recording what was compared, since the two designs differ in ways worth remembering if this is ever revisited:

- **Payload transport.** This PR used a separate `TRIAGE-PAYLOAD:` line; #582 rides the existing `TRIAGE-RESULT:` sentinel. That was chosen because the watchdog already guarantees exactly N sentinel lines as the last N non-empty lines with none elsewhere, so the payload inherits "one per command, unforgeable from a mid-transcript echo" for free rather than re-establishing it.
- **The sanitizer change here does not carry over, and does not need to.** Adding `TRIAGE-PAYLOAD` to `sanitize-repro-draft.mjs`'s defang list was required precisely because this PR introduced that token. `TRIAGE-RESULT` was already in that list, so the merged design is covered as-is. Nothing else in this branch touches a file #582 left alone.
- **Upsert structure.** This PR split it into `pick-triage-upsert.mjs` with its own tests; #582 keeps it in the renderer and tests `main()` against a stubbed fetch. A structural preference rather than a capability difference — the split is arguably tidier, and worth borrowing if that file is ever refactored.

The reasoning in this description was sound and largely matches where #582 landed after review, including the point that reading an environment variable is not a write, so the tool allowlist never guarded `AI_LOOP_PAT` at all. #582 reaches the same conclusion by splitting the job so the session holds no PAT.

PR #598, opened 2026-08-29
fix: collapse the duplicated docs route segment so Grid and Columns URLs resolve

### LINKED ISSUE #597: [Bug] Generated docs URLs 404 for Grid and Columns: the route collapses when a page's filename matches its folder
## Summary

`https://bestax.io/docs/api/grid/grid` and `https://bestax.io/docs/api/columns/columns` are dead. Docusaurus collapses a doc page whose filename equals its parent folder, so those two pages publish at `/docs/api/grid` and `/docs/api/columns`. Two generators and three hand-written files still emit the uncollapsed form.

The user-visible one: every `bestax-migrate` run that touches a `Tile` hands the user a dead link, inline in their source and again in the run report.

```
// TODO(bestax-migrate): `Tile` — Bulma v1 replaced tiles with the Grid/Cell components — see https://bestax.io/docs/api/grid/grid and the migration guide …
```

## Reproduction

```bash
pnpm exec turbo run build --filter=@allxsmith/bestax-docs
ls docs/build/docs/api/grid/grid.html   # No such file
ls docs/build/docs/api/grid.html        # exists
```

Only two pages qualify today:

```bash
for d in docs/docs/api/*/; do n=$(basename "$d"); [ -f "$d$n.md" ] && echo "$n"; done
# columns
# grid
```

## Root cause

Both generators build the URL from the file path and never account for the collapse:

- `scripts/gen-mcp-index.mjs:472` — `const docsUrl = \`${DOCS_BASE}/api/${slug}\``, `slug` being the relative path minus `.md`
- `scripts/gen-component-catalog.mjs:205-217` — same shape, `relative(API_DIR, file)` minus `.md`

So this is not two stale strings. Any future page named after its folder inherits the bug.

## Every affected occurrence

Generated, and fixed by fixing the generators:

- `bestax-mcp/data/components/Grid.json:6`, `bestax-mcp/data/components/Columns.json:6` (`docsUrl`)
- `skills/bestax-custom-component/references/component-catalog.md:126,131`
- the synced copies under `bestax-mcp/data/skills/**` and `create-bestax/templates/skills/**`

Hand-written, need editing:

- `bestax-migrate/src/sources/react-bulma-components/mapping.ts:783` (the `Tile` TODO hint)
- `skills/bestax-migrate/references/unmappables.md:35`
- `README.md:119`

Not affected, and should stay as they are: the relative `../../../api/grid/grid.md` links in `docs/docs/guides/getting-started/migration/*.md`. Docusaurus resolves a file reference to the real route, and `onBrokenLinks: 'throw'` already validates them.

## Why nothing caught it

`onBrokenLinks: 'throw'` only sees links inside the docs site. These are absolute URLs living in other packages. `check:conformance` matches no `bestax.io` URL at all, and `check:urls` is scoped to two files:

```js
// scripts/check-pointer-urls.mjs:43
export const FILES = ['bulma-ui/llms.txt', 'bulma-ui/AGENTS.md'];
```

## Suggested fix

Collapse the duplicate trailing segment once, in a shared helper both generators call, then regenerate and hand-edit the three remaining files. A regression guard is the open question: the cheapest is a conformance rule asserting no `bestax.io/docs/api/<x>/<x>` shape anywhere, which is narrow but catches the whole class by construction. Widening `check:urls` to fetch every emitted docs URL would catch more but puts a network call in the gate.

## Impact

Low severity, real reach. A migrating user hits it at exactly the moment they are looking for help, and an agent reading the MCP index or the component catalog gets a 404 for two of the layout components it is most likely to need.

allxsmith:
Closes #597.

`https://bestax.io/docs/api/grid/grid` and `https://bestax.io/docs/api/columns/columns` are 404s. Docusaurus publishes a page named after its own folder at the folder's route, so those two serve at `/docs/api/grid` and `/docs/api/columns`.

The one with user reach: every `bestax-migrate` run that touches a `Tile` wrote that dead link into the user's source as a `TODO(bestax-migrate)` comment and printed it again in the run report.

## Root cause, and why this is not two stale strings

Both generators built the URL straight from the file path (`gen-mcp-index.mjs:472`, `gen-component-catalog.mjs:217`), so every future page named after its category would inherit the bug. Fixed at the source: `docsRoute()` in the new `scripts/lib/docs-url.mjs` collapses the trailing folder-index segment, and both generators route through it. The MCP index's `slug` field is deliberately untouched, since that is the file-path identity rather than a route; only `docsUrl` changes.

`docsRoute` handles all three filenames Docusaurus treats as a folder index: `<folder>.md`, `index.md` and `README.md`. Only the first has a live instance today; the other two are covered so the next one does not reintroduce the same silent 404. Verified the index form collapses by building: `docs/docs/guides/llms/index.md` emits `build/docs/guides/llms.html` with no nested `index.html`.

Three hand-written copies are corrected directly: the `Tile` hint in `mapping.ts`, the migrate skill's `unmappables.md`, and the README.

Not touched, and correct as they stand: the relative `../../../api/grid/grid.md` links in the docs migration guides. Docusaurus resolves a file reference to the real route and `onBrokenLinks: 'throw'` already validates them.

## Why nothing caught it, and what does now

`onBrokenLinks` only resolves links inside the docs site; these are absolute URLs in other packages. `check:conformance` matched no `bestax.io` URL at all, and `check:urls` covers two files in bulma-ui.

The new `docs-api-urls` conformance sub-check closes the class. It walks the repo with explicit build-output exclusions rather than an allowlist, and its detector mirrors `docsRoute` exactly: the collapsing segment must end the URL, it may sit under any number of category segments (`form/datetime/` is a live nested category), a terminal slash counts, and the host is pinned so `notbestax.io` is not misreported. `docsUrlViolations` and `isSkippedDocsUrlPath` are exported and covered directly, because the tree is clean by construction and CI would otherwise only ever run the passing branch.

## Scoping

Split so every affected package releases:

- `fix(bestax-migrate)` — the `Tile` hint and the skill reference, which change the CLI's user-facing output
- `fix(bestax-mcp)` — the generators and the regenerated index and catalog it ships
- `fix(create-bestax)` — the bundled migrate skill. This package only releases on its own scope (`{ scope: '!(create-bestax)', release: false }`) and the bundled copy is gitignored, so without this commit scaffolded apps would keep the dead link until an unrelated release. The gap is now documented in its sync rules.
- `ci` / `docs` / `test` — the guard, the README, and the regenerated fixture, none of which release

The synced copies under `bestax-mcp/data/skills/` and `create-bestax/templates/skills/` are gitignored build artifacts rebuilt from `skills/`; I confirmed both pick up the fix, and the conformance check skips them so the source is what gets scanned.

## Verification

- `pnpm all` green
- `pnpm check:conformance` green, including the new sub-check
- `pnpm gen:catalog:check` and `pnpm gen:mcp:check` green
- `node --test scripts/docs-url.test.mjs scripts/docs-api-urls.test.mjs` green
- `pnpm --filter bestax-migrate validate:corpus` green: 32 files, 31 transformed, 0 crashes, no unknown-component TODOs
- Zero occurrences of either dead URL remain anywhere in the tree

The `todos-unmappable` fixture output moved one line, regenerated by running the built transform over its input per the package's testing rules rather than hand-edited.

## Summary by CodeRabbit

* **Documentation**
  * Corrected Columns and Grid documentation links to use the valid, shorter URLs.
  * Updated migration guidance and component catalog references to point to the corrected Grid and Columns pages.
  * Improved documentation link generation so future links match the routes served by the documentation site.

* **Bug Fixes**
  * Added validation to detect malformed documentation URLs before they are published.
  * Added coverage for valid, invalid, and edge-case documentation routes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ The published packages are versioned and released independently.
 - **[Components](https://bestax.io/docs/api/components/modal)** — Navbar, Modal, Card, Dropdown, Menu, Message, Pagination, Panel, Tabs, Breadcrumb, Toast, Tooltip, Steps, Sidebar, Carousel, Collapse, Dialog, Loading
 - **[Form](https://bestax.io/docs/api/form/input)** — Field/Control, Input, Select, TextArea, Checkbox(es), Radio(s), Switch, Slider, Rate, File, Numberinput, Autocomplete, Taginput, and DateInput / TimeInput / DateTimeInput pickers
 - **[Layout](https://bestax.io/docs/api/layout/container)** — Container, Section, Hero, Level, Media, Footer
-- **[Columns](https://bestax.io/docs/api/columns/columns)** & **[Grid](https://bestax.io/docs/api/grid/grid)** — classic 12-column flexbox columns and the Bulma v1 CSS Grid
+- **[Columns](https://bestax.io/docs/api/columns)** & **[Grid](https://bestax.io/docs/api/grid)** — classic 12-column flexbox columns and the Bulma v1 CSS Grid
 - **[Helpers](https://bestax.io/docs/api/helpers/theme)** — `Theme`, `ConfigProvider`, `useBulmaClasses`, `classNames`
 
 > Migrating from v2? Snackbar was merged into [Toast](https://bestax.io/docs/api/components/toast) in v3 — see the [migration guides](https://bestax.io/docs/guides/getting-started/migration).
--- a/bestax-mcp/data/components/Columns.json
+++ b/bestax-mcp/data/components/Columns.json
@@ -3,7 +3,7 @@
   "kind": "component",
   "category": "columns",
   "slug": "columns/columns",
-  "docsUrl": "https://bestax.io/docs/api/columns/columns",
+  "docsUrl": "https://bestax.io/docs/api/columns",
   "examples": [
     {
       "title": "Basic Columns",
--- a/bestax-mcp/data/components/Grid.json
+++ b/bestax-mcp/data/components/Grid.json
@@ -3,7 +3,7 @@
   "kind": "component",
   "category": "grid",
   "slug": "grid/grid",
-  "docsUrl": "https://bestax.io/docs/api/grid/grid",
+  "docsUrl": "https://bestax.io/docs/api/grid",
   "examples": [
     {
       "title": "Smart Grid",
--- a/bestax-mcp/data/skills.json
+++ b/bestax-mcp/data/skills.json
@@ -10,7 +10,7 @@
         {
           "id": "component-catalog",
           "file": "references/component-catalog.md",
-          "bytes": 17442
+          "bytes": 17429
         },
         {
           "id": "library-contributor",
@@ -99,7 +99,7 @@
         {
           "id": "unmappables",
           "file": "references/unmappables.md",
-          "bytes": 4866
+          "bytes": 4861
         }
       ],
       "examples": []
--- a/bestax-migrate/src/sources/react-bulma-components/__testfixtures__/todos-unmappable.output.tsx
+++ b/bestax-migrate/src/sources/react-bulma-components/__testfixtures__/todos-unmappable.output.tsx
@@ -3,7 +3,7 @@ import { Box } from "@allxsmith/bestax-bulma";
 import { Element, Tile } from 'react-bulma-components';
 
 // TODO(bestax-migrate): `Element` — no generic Element in bestax-bulma; use a semantic component (Block, Box, …) with helper props, or plain JSX with classNames — see https://bestax.io/docs/api/helpers/usebulmaclasses
-// TODO(bestax-migrate): `Tile` — Bulma v1 replaced tiles with the Grid/Cell components — see https://bestax.io/docs/api/grid/grid and the migration guide https://bestax.io/docs/guides/getting-started/migration/bulma-0-9-to-1
+// TODO(bestax-migrate): `Tile` — Bulma v1 replaced tiles with the Grid/Cell components — see https://bestax.io/docs/api/grid and the migration guide https://bestax.io/docs/guides/getting-started/migration/bulma-0-9-to-1
 // TODO(bestax-migrate): `colorVariant` — no direct equivalent; use isLight (Button/Notification) or a color shade like textColor='primary-dark'
 export const Legacy = ({ align }: { align: 'left' | 'right' }) => (
   <div>
--- a/bestax-migrate/src/sources/react-bulma-components/mapping.ts
+++ b/bestax-migrate/src/sources/react-bulma-components/mapping.ts
@@ -780,7 +780,7 @@ export const MAPPING: Record<string, ComponentMapping> = {
   },
   Tile: {
     status: 'todo',
-    todo: `Bulma v1 replaced tiles with the Grid/Cell components — see ${DOCS}/api/grid/grid and the migration guide ${DOCS}/guides/getting-started/migration/bulma-0-9-to-1`,
+    todo: `Bulma v1 replaced tiles with the Grid/Cell components — see ${DOCS}/api/grid and the migration guide ${DOCS}/guides/getting-started/migration/bulma-0-9-to-1`,
   },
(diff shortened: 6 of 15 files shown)


allxsmith:
Deep review advisories, both checked against current code.

**Advisory 2 (index/README collapse) was real, and is fixed in `eb8d488`.** `docsRoute` handled only `<folder>.md`, so a future API page named `index.md` or `README.md` would have reproduced the same silent 404 this PR exists to fix. I verified the premise rather than taking it on trust: `docs/docs/guides/llms/index.md` builds to `build/docs/guides/llms.html` with no nested `index.html`, so the index form does collapse. The helper and the conformance regex now cover all three shapes, with tests for each. Nothing regenerated, since no API page uses those filenames today.

**Advisory 1 (guard scope) reads against the pre-rework guard.** It was accurate when the review started, and `6b12083` had already replaced the allowlist with a repo-wide walk. Checked each example it names:

- `bulma-ui/README.md` is scanned. Probed it with a repeated URL and the guard reports `bulma-ui/README.md:213`.
- `.mdx` is in the extension list.
- `create-bestax/templates` is scanned; only the gitignored `templates/skills` subtree is skipped, because it is a copy of `skills/` and the source is scanned instead.

That is the racing-push case the contributor docs warn about, so flagging it rather than silently marking the advisory resolved.

`pnpm all`, conformance, and both `gen:*:check` green on `eb8d488`.

allxsmith:
Review loop closed at its ten-iteration cap. Thirteen findings, all applied, none declined.

Final head is `3cf68fb`. Copilot's **Approval recommended** landed on `c9072c7`; the one commit after it is CodeRabbit's wording fix in `create-bestax/CLAUDE.md`, a single clause in a doc paragraph.

Worth stating plainly for whoever merges: **every one of the thirteen findings was in the conformance guard, not in the fix.** The dead-URL correction and the `docsRoute` helper were never challenged across ten rounds. The guard needed four passes to stop being too narrow (roots, then extensions, then hidden directories), three to make its pattern actually mirror `docsRoute` (anchoring, trailing slash, host), and two to stop being vacuous (no failing-case coverage, then no traversal coverage).

That is worth weighing rather than just celebrating. The generators are correct by construction now, so the guard only protects hand-written URLs, and it cost far more review than the bug it guards against ever cost users. Dropping `docs-api-urls` and keeping `scripts/lib/docs-url.mjs` is a defensible merge, and the commits are split so that is a clean revert of the `ci:` ones.

Verification on `3cf68fb`: `pnpm all`, `pnpm check:conformance`, both `gen:*:check`, and `validate:corpus` all green. The traversal test was confirmed to fail for the right reason by sabotaging the walker to scan nothing.

allxsmith:
Measured run against the live site, on `3cf68fb`.

Ran the built codemod over a real react-bulma-components app (the kitchen-sink fixture, copied to a scratch dir so nothing in the repo moved). 10 files scanned, 10 transformed, 31 TODOs.

The hint it writes into the user's source:

```
// TODO(bestax-migrate): `Tile` — Bulma v1 replaced tiles with the Grid/Cell components — see https://bestax.io/docs/api/grid and the migration guide https://bestax.io/docs/guides/getting-started/migration/bulma-0-9-to-1
```

Every URL in that line checked against production:

| status | URL |
| --- | --- |
| **200** | `https://bestax.io/docs/api/grid` (what it emits now) |
| **404** | `https://bestax.io/docs/api/grid/grid` (what it emitted before) |
| **200** | `https://bestax.io/docs/api/columns` |
| **404** | `https://bestax.io/docs/api/columns/columns` |
| **200** | the migration guide link in the same hint |

Then the whole generated surface rather than just the two known cases. All **87** `docsUrl` values in the MCP index return 200, and the component catalog's 87 links are the identical set with zero differences. On `main` those same two entries emit the 404 forms, so the measurement is 2 of 87 dead before, 0 of 87 after.

The migrated app also still typechecks and its stylesheet still compiles: that is the kitchen-sink e2e, green in `pnpm all`.

So the fix is confirmed end to end, not just in the diff: the codemod's output now points somewhere that exists, and nothing else it generates points anywhere that does not.

PR #604, opened 2026-08-30
ci: enforce egress-block on the last three AI jobs, closing the audit group

### LINKED ISSUE #578: [Security] Flip the six audit-mode AI jobs to egress-policy: block
Follow-up owed by rule 10 in `.github/CLAUDE.md`, created alongside #577 (which closes #487).

## Background

#487 was written about `ai-scan` and `claude-repro`, which declared `block` and enforced
nothing. Its final comment established that the population was wrong: four of the seven AI
workflows had **no harden-runner step at all**, so the untrusted-trigger class was never why
they did not enforce — there was nothing to downgrade.

#577 gives those jobs a harden-runner step at `egress-policy: audit`:

| Workflow | Job |
| --- | --- |
| `claude.yml` | `claude` |
| `claude-implement.yml` | `implement` |
| `claude-pr-loop.yml` | `fix` |
| `claude-pr-loop.yml` | `verify` |
| `claude-review.yml` | `review` |
| `bestaxbot-reply.yml` | `respond` |

`audit` is permitted here by rule 10's "existing live job" clause, and rule 10 says plainly
that it is a temporary state that owes a follow-up issue rather than a resting place. This is
that issue.

## Why they were not flipped straight to block

None of these jobs' egress has ever been measured. They check out a branch, install
dependencies with pnpm, run repo code, and run a Claude session with a model token — a much
wider surface than the three session-only jobs #577 covers. A guessed allowlist would break
them on the first miss, which is the mistake #487 documents in the opposite direction.

`claude-implement` and `claude-pr-loop` are the two that run repo code alongside the model
token, so they are the ones that most want enforcement.

## What to do

1. Let each job run a few times in audit and collect the endpoint list from the StepSecurity
   run reports (not from reading the actions — that is how `sign-sbom`'s still-unmeasured list
   was assembled).
2. Expect at minimum the Claude CLI distribution pair (`claude.ai`, `downloads.claude.ai`),
   `api.anthropic.com`, the GitHub trio, `registry.npmjs.org` and `nodejs.org`; pnpm and the
   test runs will add more.
3. Flip each to `block` with its measured list, and add the assertion rule 10 now requires.
   **Do not copy the snippet that used to sit here** — it was a single `jq` line, which #577 has
   since superseded, and copying is how it drifted in the first place. Take the current form from
   rule 10 in `.github/CLAUDE.md`, or from any live block job (`security-txt-expiry.yml` and
   `auto-close-duplicates.yml` are the shortest). Three shapes in it are load-bearing and a
   hand-written version tends to drop all three:
   - the `[ -e /home/agent/agent.json ]` check comes **first**, because harden-runner has
     deliberate "install nothing, exit 0" paths and a bare `jq` there fails with
     `could not open file` rather than saying what happened;
   - the `agent.status` check **polls** (30×1s) rather than testing once — the pre-step gives up
     after ~9s while the agent is still resolving the allowlist;
   - it greps `^Initialized`, **not** `-qx Initialized`. `writeStatus` appends without a trailing
     newline, so a second status concatenates onto the same line and an exact-line match silently
     stops matching.
4. ~~Carry `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC` onto these sessions at the same time~~
   **Already done in #577 — nothing to carry.** This step originally deferred the setting on the
   grounds that it blocks nothing while these jobs are at `audit`. That reasoning was wrong, and
   review caught it: the *purpose* of audit mode here is to produce the endpoint list step 1
   collects, and with telemetry left on, `http-intake.logs.us5.datadoghq.com` appears in that
   report — inviting exactly the allow-listing #577 argues against. All nine
   `claude-code-action` steps now set it, so the measurement reflects the intended end state.
5. Update rule 10's table in `.github/CLAUDE.md` as each job moves between its three groups.

## ~~Also unresolved, and smaller~~ — settled in #585

~~`auto-close-duplicates.yml` carries no harden-runner either. It is a scheduled issue-writer
with no model session, so it is a lower-value target than the five above, but it was flagged
in the `security-txt-expiry.yml` comment as a decision deferred rather than made. Either give
it one at `block` (it runs on `schedule`/`workflow_dispatch`, outside the class that broke) or
record why it does not need one.~~

Given one at `block`, with the five-host allowlist its sibling already uses and the rule 10
assertion. Verified on a real dispatch rather than by inspection — see the comment below. **The
six-job flip above is what keeps this issue open.**

`supply-chain.yml`'s `sbom`, `attach-sbom` and `verify-provenance` are grandfathered by rule
10's existing-job clause and are out of scope here.

### LINKED ISSUE #593: [CI] verify runs only on blocking deep-review findings — #578 still needs one measured run
Split out of #578, which needs a `verify` run to measure its egress.

> **Re-scoped 2026-08-29.** The original diagnosis — that `claude-review.yml` files no inline
> comments and that `verify` has never run — was wrong on both counts. Evidence is in the
> comment below; the corrected version follows.

## What is actually true

`verify` works and has run end to end. `claude-review.yml`'s prompt has filed every blocking
finding as an inline comment since #216 (`mcp__github_inline_comment__create_inline_comment`,
the first entry on its `--allowedTools`), those threads arrive authored by `claude[bot]` — which
the gate's `test("^claude"; "i")` matches — and the full protocol has executed twice:

| run | PR | thread | outcome |
| --- | --- | --- | --- |
| 32857163749 | #564 | `docs/…/alternative-icons.md` | `Fixed in db3d5883…` → `Verified fixed — …` → resolved → handoff |
| 32913565697 | #569 | `skills/bestax-theming/references/css-variables.md` | same shape |

## The real gap

`verify` fires only when the deep review posts a 🔴/🟠/🟡 finding. 🔵 Advisory is deliberately
barred from the inline channel — advisories must not spin the fix loop — so a 0-blocking review
produces no claude thread and nothing for `verify` to do.

That is rare. Of the 24 deep reviews on #560–#592, **4** reported a blocking finding (#561,
#564, #569, #582), and all 4 produced a `claude[bot]` inline thread. The mechanism is exact; the
frequency is the constraint.

#577 gave `verify` its harden-runner on 2026-08-28 (c7e501d). Every deep review since — #586,
#587, #590, #591 and #592 ×3 — has returned 0 blocking, so no verify run has occurred under the
new step and #578 has no endpoint report to read.

## What to do

Nothing in the workflows. Wait for a natural blocking finding on an `ai-loop` PR, then take the
StepSecurity report from that run. `claude-pr-loop`'s `verify` stays at `egress-policy: audit`
until then — which is the same conclusion as before, for a different reason.

Explicitly **not** to do, since both were proposed in the original framing:

- Do not loosen `PENDING_VERIFY` / `REBUTTABLE` to accept CodeRabbit or Copilot threads.
  CodeRabbit verifies and resolves its own; the fix agent resolves Copilot's. Widening would
  double-verify and contradict the mode table.
- Do not move 🔵 Advisory findings inline to raise the hit rate. The prompt bars them on
  purpose — an inline thread is an actionable work item for the fix agent, and the loop would
  spin on findings that were never meant to block.

## Acceptance

- The next deep review that files a blocking finding on an `ai-loop` PR drives a `verify` run
  to completion.
- That run's StepSecurity report is attached to #578, and `claude-pr-loop`'s `verify` flips to
  `egress-policy: block` with the rest.

allxsmith:
Second half of #578. Flips `review` (`claude-review.yml`) and `fix`/`verify` (`claude-pr-loop.yml`)
to `egress-policy: block` with the rule 10 assertion, which empties rule 10's "audit, deliberately"
group.

> **Stacked on #602.** Based on `ci/578-egress-block-default-branch-jobs` rather than `main`,
> because both PRs edit the same rule 10 inventory lines and counts. Retarget to `main` once #602
> merges. Review this diff on its own; #602's three jobs are not part of it.

## `verify` is measured — and #593's premise turned out to be avoidable

#593 concluded that nothing could be done but wait for a natural blocking deep-review finding.
That is true of the `PENDING_VERIFY` path. It is not true of the gate, which is an OR:

```
elif [ "$PENDING_VERIFY" -gt 0 ] || [ "$REBUTTABLE" -gt 0 ]; then MODE=verify
```

`REBUTTABLE` (`claude-pr-loop.yml:230-234`) wants an unresolved claude-opened thread whose last
comment is from a **non-claude** author with `ccount < 2`. A maintainer replying to the inline
finding satisfies it directly — no fix round, no push, no iteration consumed.

Measurement run: [33284876176](https://github.com/allxsmith/bestax/actions/runs/33284876176)

```
decision: verify (fail=0 pend=0 actionable=0 verify=0 rebut=1 cr=success iter=0)
gate: success   verify: success   (fix / handoff / halt all skipped)
```

**The session was representative, not a stub** — which is the only thing that makes its endpoint
list worth anything. It checked out, ran `pnpm install --frozen-lockfile`, and the Claude session
swept every `<Button>`+`<Icon>` pair across `docs/docs/**` and `skills/**`, read `Icon.tsx:195` to
confirm the `ariaLabel` default, replied on the thread and resolved it. Full protocol end to end.

The four-step recipe is written up in #593 for whoever needs to re-measure this job.

## Endpoints

All three jobs produced the same six application hosts as the other three — `api.anthropic.com`,
`api.github.com`, `claude.ai`, `downloads.claude.ai`, `github.com`, `registry.npmjs.org` — so all
six of #578's jobs get the same eight-host list, the one already enforcing on `ai-scan`/`scan`,
`ai-triage`/`triage` and `claude-repro`/`author`.

`verify` matching `fix` is the expected result rather than an assumed one. #578 was right to refuse
to derive one job's list from another's; the derivation happened to hold, and now it is checked.

`TURBO_TELEMETRY_DISABLED: '1'` joins `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC` on all three.
It matters most on this loop: its sessions run `pnpm test`/`pnpm run` on nearly every iteration, so
turbo is invoked constantly, and turbo samples its telemetry — an intermittent that would otherwise
redden the loop weeks after the flip.

The Actions cache blob host stays off, and `verify` supplied fresh evidence for why:
`productionresultssa13` appeared, joining sa3/6/7/9/11 from the earlier runs. Six runs, six
different names.

## `fix` gets one extra comment

Its measured pre-step-to-`Initialized` time was **9.97s**, past the ~9s the pre-step gives up
after. So on this job the polling loop is not defensive programming — a single-shot check would
have failed it. That is recorded next to the assertion rather than left in an issue thread.

## Docs

- Rule 10's inventory: nineteen enforcing jobs, empty audit group. Counts re-derived with the
  greps the rule prescribes (19 key-form, 20 plain-form, 0 audit), not incremented.
- The audit group is **kept rather than deleted**. Audit is still the right starting point for an
  existing live job whose egress has never been observed; what it is not is a resting place, which
  is the whole reason #578 existed.
- I1's egress clause now covers all five code-executing jobs, and restates the scope rule 10 sets
  for it — it bounds where data can go, and `api.github.com` is necessarily allow-listed, so it
  does not bound what a session does through a tool.

## Verification

Local: `prettier --check`, `check:conformance` (19 green), both rule 1 pin greps, and a js-yaml
pass over every workflow asserting that no `block` job lacks an assertion and that all six of
#578's assertions are byte-identical to `auto-close-duplicates.yml`'s apart from the credential
each names. (The seven pre-existing `ai-scan`/`ai-triage` variants differ by design — the `fail()`
wrapper rule 10 documents.)

**Before merge**, unlike #602's three, all three of these can be exercised on the branch:

- `fix` / `verify`: `gh workflow run claude-pr-loop.yml --ref ci/578-egress-block-branch-verifiable-jobs -f branch=<a claude/ PR>`
- `review`: label a PR whose head is this branch. Note the session itself will skip (the OIDC
  exchange requires the workflow to match the default branch when the PR modifies it), so that
  exercises the policy and the assertion, not the session's egress — acceptable, since the session
  egress is already measured.

Watch for: assertion green, `"egress_policy":"block"` in the config, `Initialized` with no
`Reverted changes` or `timed out`, and the session completing real work.

Closes #578
Closes #593

### DIFF
--- a/.github/CLAUDE.md
+++ b/.github/CLAUDE.md
@@ -73,13 +73,12 @@ regardless of how convenient it is.
   `claude-pr-loop` (`fix`/`verify`), `claude-review`, `claude` and `bestaxbot-reply` all check
   out a branch and run repository code in the same job as that token, and they must — fixing and
   reviewing code is the job. What holds them is a different, weaker set: the tool allowlist, the
-  protected-path deny rules, the trusted-labeler gates, and an enforced egress policy — which as
-  of #578 covers `claude`, `claude-implement` and `bestaxbot-reply`, while `claude-review` and
-  `claude-pr-loop`'s `fix`/`verify` are still at `audit` and so have that last leg only in name.
-  Check the rule 10 inventory before citing it for a specific job rather than reading this
-  sentence as a group claim. Saying I1 covers those jobs would be the exact overstatement the
-  checklist at the bottom of this file ends on. It covers the pipeline that was designed around
-  it.
+  protected-path deny rules, the trusted-labeler gates, and — as of #578, for all five of them —
+  an enforced egress policy on a measured allowlist. Read that last leg for exactly what rule 10
+  says it delivers: it bounds **where** a session can send data, and `api.github.com` is
+  necessarily on every one of those allowlists, so it does not bound what the session can do
+  through a tool. Saying I1 covers those jobs would be the exact overstatement the checklist at
+  the bottom of this file ends on. It covers the pipeline that was designed around it.
 
   harden-runner's egress block enforces again as of #487 — `block` used to degrade silently to
   `audit`, and now enforces and is asserted (rule 10). **Do not promote it to a third leg of
@@ -341,13 +340,13 @@ Copy the full version from any block job — the three shapes above are each loa
 
 - **The `-e` check comes first** because harden-runner has deliberate paths that install nothing
   and still exit 0 (a StepSecurity outage, the `skip-harden-runner` repo property, a container or
-  slim runner). Without it a vendor outage fails every block job at once — sixteen of them as of
+  slim runner). Without it a vendor outage fails every block job at once — nineteen of them as of
   the inventory below, not the seven this line said before the `ai-scan`/`ai-triage` job splits —
   with a bare `jq: could not open file`. It still fails, because a job holding a credential must
   not run unprotected, but it says why. Do not hand-maintain that number: it is the count of
   block-mode jobs, and it has been wrong once already. Derive it, and grep for the **key** rather
   than the string — `grep -rn "^ *egress-policy: block$"`. A plain `grep egress-policy: block`
-  returns seventeen: `deploy-worker.yml` names the policy in a comment, which is the same way a
+  returns twenty: `deploy-worker.yml` names the policy in a comment, which is the same way a
   `node -e` in a comment misclassifies a job below.
 - **The status check polls** rather than testing once. The pre-step waits only for the file to
   _exist_ and gives up after ~9s, while the agent resolves every allow-listed host before writing
@@ -479,13 +478,15 @@ jobs (`ci.yml`, `deploy.yml`, `test-deploy.yml`, `visual-regression.yml`, `story
   `ai-triage` jobs (`gate`, `triage`, `publish`, `cleanup`), `claude-repro` (`author` **only**),
   `deploy-worker` (`deploy`), `supply-chain` (`consumer-sbom` and `sign-sbom`),
   `security-txt-expiry` (`check`), `auto-close-duplicates` (`auto-close`), `claude` (`claude`),
-  `claude-implement` (`implement`), `bestaxbot-reply` (`respond`). Sixteen jobs; the command
-  below is the check.
-- **Audit, deliberately, pending a measured allowlist** — `claude-pr-loop` (`fix` and `verify`)
-  and `claude-review`. These run repo code with a model token; their block flip is the remainder
-  of the follow-up this rule owes, tracked in #578. `fix` and `review` are measured and waiting
-  only on their own PR; `verify` is the one job with no measurement yet, because it runs only
-  when a deep review files a blocking inline finding (#593).
+  `claude-implement` (`implement`), `bestaxbot-reply` (`respond`), `claude-review` (`review`),
+  `claude-pr-loop` (`fix` and `verify`). Nineteen jobs; the command below is the check.
+- **Audit, deliberately, pending a measured allowlist** — none, as of #578. That issue closed the
+  group by measuring all six of its members instead of guessing for them, and all six landed on
+  the same eight hosts the Claude-session jobs above already used. Keep the group here rather
+  than deleting it: audit remains the correct starting point for an **existing** live job whose
+  egress has never been observed, and the recipe under this rule is what keeps it a short state
+  rather than an open-ended one. What it is not is a resting place, which is the whole reason
+  #578 existed.
 - **No harden-runner at all**, and these are two different groups — do not merge them into one
   "API-only" line, which understates the second:
   - _Genuinely API-only_ — no checkout, no repository code: `on-slop`, `auto-label-claude-prs`,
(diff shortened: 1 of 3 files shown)


allxsmith:
## Branch verification done for `review`, and it corrected the PR

Applied `deep-review` to this PR, so `claude-review.yml` ran **from this branch's version** at
`egress-policy: block`: run [33286967625](https://github.com/allxsmith/bestax/actions/runs/33286967625).

| Check | Result |
| --- | --- |
| `Assert egress policy is enforced` | success |
| effective config | `"egress_policy":"block"` |
| agent status | `Initialized`, no `Reverted changes`, no `timed out` |
| `Checkout PR head` → `Setup pnpm` → `Setup Node.js` → `Install dependencies` | all success |
| `pnpm install --frozen-lockfile` | completed, 15.7s |

The Claude session skipped, as expected on a PR that modifies `claude-review.yml` (the OIDC
exchange requires the file to match the default branch). So this verifies the policy, the
assertion and the whole install path, not the session's egress — which was already measured in
audit on runs 33230210532 / 33230112308.

### It falsified a claim these comments made

I had written that omitting the Actions cache hosts costs a cache miss that "degrades to a
`registry.npmjs.org` refetch". That was a prediction, and it was wrong. `setup-node` logged
`Cache hit` and `Cache restored successfully` at `block`, with neither
`productionresultssa<N>.blob.core.windows.net` nor `results-receiver.actions.githubusercontent.com`
allow-listed. harden-runner does not gate the runner's own control-plane traffic, so leaving those
hosts out costs nothing at all.

Corrected in `ec968f3` here and on the base branch, and rule 10's measurement section now says
plainly that the prediction was wrong — the point of that section is to read a run rather than
reason about what a policy will cost.

The run also produced `productionresultssa13`, a sixth distinct blob host name across six runs,
which strengthens the "cannot be pinned" argument rather than changing it.

### `fix` / `verify` on the branch

Not separately exercised. Doing so needs an open `claude/…` PR with `ai-loop`, and the fixture PR
(#603) is closed and its branch deleted. They run the identical
checkout → pnpm → setup-node → `pnpm install --frozen-lockfile` sequence that just passed at
`block` above, and their session egress is measured (33233097667 and 33284876176). The remaining
gap is their assertion step under a real dispatch, which the next `ai-loop` PR after merge will
exercise on its own.

Issue #667, opened 2026-09-10
[Bug] bulma-ui: the public `*Props` aliases don't distribute like their components do


allxsmith:
## Affected Package(s)

- [x] bulma-ui (`@allxsmith/bestax-bulma`)
- [ ] create-bestax (`create-bestax`)
- [ ] docs (`@allxsmith/bestax-docs`)
- [ ] Other (please specify):

---

**Why this qualifies:** (a) defect. A component and its exported prop type disagree about the same union.

## The gap

#661 made `PolymorphicProps` distributive, so a union-typed `as` keeps each member's own props:

```tsx
<Button as={cond ? 'a' : 'button'} href="/x" />   // ✅ compiles
```

The public aliases were not changed, and they apply `Omit` to the union *after* it is formed. `keyof (A | B)` is only the shared keys, so:

```tsx
const props: ButtonProps<'a' | 'button'> = { as: 'a', href: '/x' };  // ❌ href is not assignable
```

A consumer writing a generic wrapper reaches for the exported alias and gets stricter behaviour than the component they are wrapping. All eight are affected: `ButtonProps`, `LinkButtonProps`, `LinkProps`, `AvatarProps`, `RevealProps`, `NavbarItemProps`, `NavbarLinkProps`, `MenuItemProps`.

## Why it wasn't fixed in #661

Wrapping each alias in `T extends unknown ? … : never` is a two-line change per file and it works at the type level — but it defeats `scripts/lib/props-extract.mjs`. That extractor reads heritage **syntactically**, and a distributive conditional is not an intersection it can walk: **Button's props table drops from 19 rows to 2**, and every affected page and MCP entry loses most of its props. `props-extract.test.mjs` catches it (10 failures), so it fails loudly rather than silently — but the extractor has to learn the shape before the aliases can change.

## Two ways through

1. **Teach the extractor to unwrap a distributive conditional** — where the check type is the alias's own type parameter and both branches are known, walk the true branch. Then make all eight aliases distributive.
2. **Have the aliases refer to `PolymorphicProps<T, XOwnProps>`** instead of restating the intersection, and teach the extractor to follow that one helper. Fewer places to drift, but the extractor still needs the rule.

Option 2 removes the duplication that let the two forms diverge in the first place, which is the underlying cause rather than the symptom.

## Acceptance

- [ ] `const p: ButtonProps<'a' | 'button'> = { as: 'a', href: '/x' }` compiles, and the same for the other seven aliases
[...]

Issue #641, opened 2026-09-03
[Bug] bulma-ui: props and ref types do not follow the polymorphic `as`


allxsmith:
## Affected Package(s)

- [x] bulma-ui (`@allxsmith/bestax-bulma`)
- [ ] create-bestax (`create-bestax`)
- [x] docs (`@allxsmith/bestax-docs`)
- [ ] Other (please specify):

---

**Why this qualifies:** (a) defect. The declared types are narrower than what the components actually support at runtime. `as` accepts any `React.ElementType` and the DOM props are forwarded to whatever it renders, but every typed prop is pinned to the default element, so correct code fails to typecheck and incorrect code passes.

Surfaced while reviewing #637 (ref forwarding, closes #622). CodeRabbit raised it on `Button.tsx` and `Navbar.tsx`; both were deliberately declined there as out of scope, with the reasoning recorded on those threads.

## The gap

`as` was widened to `React.ElementType` (#188) but the prop types were not derived from it. Two consequences, in opposite directions:

**Correct code is rejected.** `ref` is typed `HTMLButtonElement | HTMLAnchorElement`, so this fails to compile even though it works:

```tsx
const ref = React.createRef<HTMLDivElement>();
<Button as="div" ref={ref}>…</Button>
```

**Incorrect code is accepted.** `ButtonProps extends Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, …>` (`bulma-ui/src/elements/Button.tsx:13-16`), so with `as="div"` the whole button surface still typechecks. `NavbarLinkProps extends Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, 'color'>` (`bulma-ui/src/components/Navbar.tsx:421-424`) is sharper still: with `as="span"`, `href`, `target` and `rel` all remain valid props on a span. `onClick` is hand-typed to the default element too.

So the ref is not uniquely wrong; it is one symptom of the prop surface not following `as`. Fixing the ref alone would make the API *less* consistent.

## Scope

Components where `as` is polymorphic and the props are pinned:

- `elements/Button.tsx` (and `elements/LinkButton.tsx`, whose props are `Omit<ButtonProps, …>`, so it follows automatically)
- `components/Navbar.tsx` — `Navbar.Link`, and `Navbar.Item` alongside it for consistency
- `components/Menu.tsx` — `Menu.Item`

[...]

PR #671, opened 2026-09-13
fix(bulma-ui): type-check the tests and the stories, and clear the 61 errors behind them

### LINKED ISSUE #663: [Bug] bulma-ui: nothing type-checks the tests or the stories, and 60 errors have accumulated
## Affected Package(s)

- [ ] Other (please specify):

---

**Why this qualifies:** (a) defect. A whole class of type error is invisible to every gate, and the backlog it has hidden includes real bugs in shipped stories and a prop that tests exercise but the public type does not have.

Surfaced while doing #641 (#661), which needed type-level assertions and had to find somewhere they would actually be checked.

## The gap

Two independent reasons, either one sufficient:

1. **ts-jest never type-checks.** The root `tsconfig.json` sets `"isolatedModules": true`, and ts-jest gates its language service on exactly that (`ts-compiler.js`: `if (!this.configSet.isolatedModules)`). Everything goes through `ts.transpileModule` instead, so `pnpm test` strips types rather than checking them.
2. **`tsc --noEmit` never sees the files.** `bulma-ui/tsconfig.json` excludes `src/**/__tests__/**`, `**/*.test.tsx`, `**/*.stories.tsx` and `**/*.stories.ts`.

Storybook's vite builder does not typecheck either, and typescript-eslint's type-aware rules do not surface compiler errors. So no gate anywhere looks.

The one consequence worth stating plainly: **an `@ts-expect-error` in a test file is a dead comment today**. It suppresses nothing because nothing checks. (The good news, below: all the existing ones turn out to be honest.)

## What it has hidden

Point a `tsc --noEmit` at `src/**/*` with the excludes lifted and there are **60 errors across 31 files**. None come from #661.

| Code | Count | What it is |
| --- | --- | --- |
| TS2322 | 24 | a prop given a value its type does not allow |
| TS6133 | 20 | unused `React` import |
| TS2352 / TS2493 | 9 | `Rate.test.tsx` casting through tuple indexes that may be `undefined` |
| TS2554 | 5 | `Card.test.tsx` calling something with one argument where two are declared |
| TS2367 | 1 | a comparison with no overlap |
| TS2307 | 1 | no module declaration for `*.svg` |

The TS6133 half is mechanical noise. Three of the rest are real:

- **`Sidebar.stories.tsx:519` renders `<SubTitle size="7">`.** Bulma's sizes stop at 6, so that emits `is-7`, which no rule matches — the price in that shipped story is silently unstyled.
- **`Dialog.stories.tsx:167` compares `type === 'default'`** against a union that has no `'default'` member. The true branch is dead, so the neutral button the story means to show has never rendered.
- **`Dropdown.stories.tsx:24` renders `<Dropdown.Item as="a" href target>`** and `DropdownItemProps` rejects it. This is the #641 defect on a component with a *constrained* `as`: `as?: 'a' | 'div' | 'button'` alongside props pinned to `React.HTMLAttributes<HTMLElement>`, so `href`/`target` are rejected under `as="a"` and `disabled`/`type` under `as="button"`, all of which work at runtime. #661 fixed the eight components with an open `React.ElementType`; this is the same bug in the narrower form, and the story is the proof it bites.

Also worth a look rather than a blanket suppression: `Icon.test.tsx:461,468,474` pass a deprecated `icon` prop that `IconProps` no longer declares. Either the type dropped a prop the runtime still honours — in which case consumers on the deprecated path have been getting a type error we cannot see — or the tests are exercising a path that no longer exists.

**One genuinely reassuring result:** zero TS2578. Every `@ts-expect-error` currently in the tree is still suppressing a real error, so turning checking on validates the existing ones rather than invalidating them.

## Proposal

Add a second TypeScript project rather than lifting the excludes from `bulma-ui/tsconfig.json` — that file is also the build's, via `@rollup/plugin-typescript` with `declaration: true`, and pulling tests and stories into the emit program is a change nobody wants.

```jsonc
// bulma-ui/tsconfig.test.json
{
  "extends": "./tsconfig.json",
  "compilerOptions": { "rootDir": "." },
  "include": ["src/**/*"],
  "exclude": ["node_modules", "dist", "coverage"]
}
```

with a `typecheck:tests` script wired into `pnpm all` and CI, next to the existing `typecheck`. Fix the 60 first, or the gate cannot go in.

Leaving `isolatedModules` alone is deliberate: a separate `tsc` pass is a better gate than making every jest run pay for the language service, and it is what makes an `@ts-expect-error` in a test file mean something again.

Two questions worth settling while fixing rather than after:

- Should a test asserting an *invalid* value (`Content.test.tsx` passing `"invalid-color"` to prove it is ignored) use `@ts-expect-error`, or a cast? The directive documents the intent better and now fails loudly if the value ever becomes valid.
- `*.svg` needs a module declaration somewhere for `Navbar.stories.tsx` to resolve its logo import.

## Acceptance

- [ ] All 60 errors resolved — fixed where they are real, `@ts-expect-error` with a reason where the invalidity is the point, never a blanket `any`
- [ ] `SubTitle size="7"`, the dead `'default'` comparison, and the `Dropdown.Item` prop pinning each addressed on their merits (the last may warrant its own issue, as a follow-on to #641)
- [ ] The `Icon` deprecated-prop mismatch resolved in whichever direction is correct, and the answer written down
- [ ] `bulma-ui/tsconfig.test.json` + a `typecheck:tests` script, running in `pnpm all` and in CI
- [ ] `pnpm all` green, coverage unchanged

allxsmith:
Closes #663.

> [!IMPORTANT]
> **Do not squash-merge this.** The PR title becomes the squash commit, and `bestax-migrate/release.config.js` opens with `{ scope: '!(bestax-migrate)', release: false }` — so a `fix(bulma-ui)` squash commit publishes bulma-ui and silently suppresses the `bestax-migrate` release. The codemod fix would land in the repo and never reach users, while the library change it depends on *does* ship. Merge with **rebase** or a **merge commit** so both scoped commits reach `main` and each package's semantic-release picks up its own.
>
> The two halves cannot be separate PRs: `bestax-migrate`'s e2e tests typecheck its tables against the built library, so the library change alone fails them and the codemod change alone fails them too. Verified both directions. The change is atomic across the two packages; only the merge method can express that.

Nothing type-checked the tests or the stories, for two independent reasons: ts-jest gates its language service on `isolatedModules`, which the root tsconfig sets, so `pnpm test` transpiles types away; and `bulma-ui/tsconfig.json` excludes `__tests__`, `*.test.tsx` and `*.stories.*`. Storybook's vite builder doesn't type-check either, and typescript-eslint's type-aware rules don't surface compiler errors. 61 errors had accumulated, and an `@ts-expect-error` in a test file suppressed nothing because nothing looked.

## The gate

`bulma-ui/tsconfig.test.json` is a second project rather than a lifting of those excludes — that file is also the build's program, via `@rollup/plugin-typescript` with `declaration: true`, and tests and stories don't belong in the published declaration emit. `types/assets.d.ts` sits outside `src/` for the same reason.

`typecheck:tests` runs in `pnpm all`, in CI, and in the React 18/19 matrix. The matrix leg adds no signal today — `pin-react.mjs` pins `react`/`react-dom` and never `@types/react` — but that is equally true of the `typecheck` step beside it, and running one without the other reads as an oversight.

Both `typecheck` steps stay, for a reason worth stating because it is not obvious: `tsconfig.test.json` extends the build config and **replaces** its include/exclude rather than merging, so its program is a strict superset and the narrow step cannot change an outcome on file coverage alone. What it asserts is that the *shipped* program type-checks without `types/assets.d.ts` — and its `declare module '*.svg'` — in scope. That matters more than it sounds: `@rollup/plugin-typescript` reports semantic errors as **warnings**, so `pnpm build` succeeds on source that does not type-check. These two steps, not the build, are what stands between a type error and a published package.

## The real bugs behind the 61

- **`Sidebar.stories`** asked a `SubTitle` for `size="7"`; Bulma stops at 6, so `is-7` styled nothing.
- **`Dialog.stories`** compared its type against `'default'`, a member its own list omitted — the true branch was dead, so the neutral dialog the story means to show had never rendered.
- **`Dropdown.Item`** pinned every prop to `React.HTMLAttributes<HTMLElement>` while `as` offered `'a' | 'div' | 'button'`, rejecting `href`/`target` on the anchor and `type`/`form` on the button — all of which have always worked at runtime, and the docs page shows the anchor form. This is #641 in the narrower shape a constrained `as` takes.
- **`Icon`'s deprecated `icon` prop** had no union member, so every consumer still on it got a type error. The implementation carried a cast for "legacy callers that bypass the type" — a deprecation announced only to whoever tries it.

Two more the sweep found that the issue didn't name: four columns asking for `size="one-half"` (Bulma's modifier is `is-half`), and `gap` on a `Block`, which is a `Grid` prop.

## Defects found in review, not by the gate

Three rounds of review (Copilot ×3, plus two local `/code-review` passes) each found real defects in the previous round's work. Worth listing, because they are the substance of the branch rather than the type errors:

- **`{ icon: '…' }` crashed two components.** Declaring the deprecated path made it a legal icon config, but `Control` and `IconText` discriminated on `'name' in value || 'children' in value` — it matched neither, fell to the node branch, and React threw *"Objects are not valid as a React child"*. Reachable from `Control`, `Input`, `Select`, the three date/time inputs, `Field`, `Autocomplete` and `IconText`.
- **The first fix for that was too loose.** `'icon' in value` claims a Font Awesome `IconDefinition` — whose `icon` is a path array, not a class string — turning the crash into a silent blank icon. The guard now tests each member's *value*, lives in one place, and deliberately sits outside the wildcard-exported module so it does not become public API forever.
- **`<Dropdown.Item as="button">` submitted its enclosing form.** `<button>` defaults to `type="submit"`. Defaulted it, then found the default was erased by `type={undefined}` arriving through a spread, so it now reads through the forwarded props.
- **The `href` strip covered only `href`** while the type derived the whole anchor surface, so `target`/`download`/`ping` still reached a `<div>`. `bestax-migrate` already strips exactly that set, which left the codemod stricter than the component.
- **The new conformance check failed open twice** — `turbo run --filter=X task` contributed no tasks, then `turbo run --force build` swallowed `build`. Both were found by review rather than by a test, which is what `.github/CLAUDE.md` asks a `node --test` sibling to prevent; the parser is now extracted and covered.

## Known follow-ups

- **`Level.Item` is the same defect** — #672. `<Level.Item as="a" download="f" />` fails to compile while `<Level.Item as="div" href>` compiles and is dropped at runtime.
- **No breaking change.** `DropdownItemProps` defaults its type parameter to the whole `'a' | 'div' | 'button'` union, so every shape that compiled before still compiles — `{ as: 'div' }` included. The cost is the one #667 describes: the alias is not distributive, so the bare form carries only the keys the three tags share and therefore no `href` — exactly what it carried before, so nothing regresses. `DropdownItemProps<'a'>` is the anchor's shape, and the component has full precision at every call site. Making the alias distributive was tried and reproduces #667's extractor failure on a second component (the props table falls from six rows to two), so #667 now carries that measurement and the `next-major` label.
- `bestax-migrate`'s `BESTAX_RANGE` is `^5`, so a migrated project could install a 5.x predating this fix. npm resolves `^5` forward, and it is the same exposure every existing `HREF_OK` row carries.

## Verification

`pnpm all` green, `check:conformance` green, all four generators current, `node --test "scripts/*.test.mjs"` green, coverage 99.6 / 99.08 / 99.76 / 99.83 against the 99% bar.

The dead-`gap` docs sweep that was part of this branch is split out — it is independent, releases nothing, and removing it brings this PR under CodeRabbit's 100-file review limit, which it had been silently skipping.

## Summary by CodeRabbit

* **New Features**
  * `Dropdown.Item` supports `a`, `div`, and `button` variants with element-specific props.
  * Links retain `href` by default; non-link variants omit anchor-only attributes.
  * Button items default to `type="button"` while honoring explicit types.
  * Deprecated `Icon` props remain supported and type-check correctly.

* **Documentation**
  * Updated dropdown migration guidance and legacy icon documentation.
  * Added clearer deprecation notes and replacement recommendations.

* **Tests**
  * Added separate type-checking coverage for tests and stories.
  * Expanded coverage for polymorphic dropdown behavior and link handling.


allxsmith:
@coderabbitai review

Issue #672, opened 2026-09-13
[Bug] bulma-ui: Level.Item's props do not follow its constrained `as`, in both directions


allxsmith:
## Affected Package(s)

- [x] bulma-ui (`@allxsmith/bestax-bulma`)
- [ ] create-bestax (`create-bestax`)
- [ ] docs (`@allxsmith/bestax-docs`)
- [ ] Other (please specify):

---

**Why this qualifies:** (a) defect. `Level.Item` rejects correct code and accepts dead code at the same time — the literal #641 report, on the one component #641 and #663 both walked past.

Surfaced while doing #663 (#671). That PR fixed `Dropdown.Item`, the other constrained-`as` component whose union members genuinely differ, and added the helper this one needs.

## The defect

`LevelItemProps` extends `React.HTMLAttributes<HTMLDivElement | HTMLParagraphElement | HTMLAnchorElement>` and hand-declares `href`, `target` and `rel` as loose strings, valid at every `as`. That is the pre-#641 workaround, and it is wrong in both directions.

**Correct code is rejected.** Only the three attributes someone thought to list exist, so the rest of the anchor's surface is a type error even at `as="a"`:

```tsx
<Level.Item as="a" download="f" />          // TS2322: Property 'download' does not exist
<Level.Item as="a" referrerPolicy="no-referrer" />  // TS2322
```

Verified against `main` with `tsc -p bulma-ui/tsconfig.test.json`. #641's own report names `download` and `referrerPolicy` as the symptom.

**Dead code is accepted.** `href` is declared at every `as`, so this compiles:

```tsx
<Level.Item as="div" href="/x" />
```

and `Level.tsx` drops it at runtime — the forwarding branch is gated on `Tag === 'a'`. A silently dead attribute is exactly what `bestax-migrate`'s href pass exists to remove.

## Why this one costs more than the others

`Footer` (`'footer' | 'div'`), `Media` (`'article' | 'div'`, `'figure' | 'div'`), `Image` (`'figure' | 'div' | 'p'`) and `Control` (`'div'`) all have a constrained `as` whose members share one attribute surface, so a single interface really does describe them. `Level.Item` and `Dropdown.Item` are the two where the members differ, and `Dropdown.Item` is done.

[...]

PR #683, opened 2026-09-16
refactor(bestax-migrate): delete the per-target link-attribute table

allxsmith:
Follow-up to #675.

## The dead mechanism

`TARGET_LINK_ATTRS` recorded the link attributes a target accepts when its props do not follow `as`. It existed for exactly one row: `Level.Item` enumerated `href`, `target` and `rel` and stopped, so the element alone could not say that `download` was rejected.

#672 made it carry the anchor's whole surface by subtracting `HTMLAttributes` from `AnchorHTMLAttributes`, the row went, and since then:

- `const declared = TARGET_LINK_ATTRS[target]` is always `undefined`
- `else if (declared && !declared.includes(name))` short-circuits for every attribute of every target, forever
- the TODO it selected — *"declares its own props rather than taking the element's"* — is text nothing can reach

Kept, it is a mechanism no test can exercise, since the only test that ever did was deleted with the row. Deleted, it is a `git revert` away the day a component narrows again.

The e2e premise check stays: every href-capable target accepts every link attribute on its anchor. It already covered all of them — the table has been empty since #675, so the `if (TARGET_LINK_ATTR_TABLE[target]) continue` guard never fired. What goes here is that dead `continue`, not a narrowing; the widening landed with #675.

Most of the diff is the narration going with it — the loop preamble that opened "Both have to allow it" and explained a row deciding the TODO's wording, a `describe` still calling the table an "exception", and the JSDoc the deleted export left behind documenting whatever followed it. The one case worth keeping is recorded as history, because it is the shape that would bring the table back: a component enumerating a **subset** of the link attributes, narrower than the element it renders.

## Why this PR exists at all

#675 was squash-merged under `fix(bulma-ui)`, and `bestax-migrate/release.config.js` opens with `{ scope: '!(bestax-migrate)', release: false }` — so the codemod changes in that squash reached `main` but were never published. Today `bestax-migrate@2.3.3` still strips `download` from `<Level.Item as="a" download>` and writes a TODO calling it unsupported, against a `bulma-ui` that now supports it.

This PR's `refactor(bestax-migrate)` scope cuts that release, so the fix ships. The cleanup is real work rather than a version bump for its own sake, but the release is the reason it is happening now.

## Verification

`pnpm all` green, `check:conformance` green, migrate suite green. Coverage on the touched file improves because the unreachable arm stops counting against it, not because anything new is tested — `pnpm --filter bestax-migrate run test:coverage` reproduces it. (Not `exec jest` directly: that bypasses the `--experimental-vm-modules` the `test` script sets, half the suites fail to load, and the figure it prints covers only the rest.)

**The kept-attribute side is now pinned**, which it was not. This PR is the release vehicle for #672's `download` fix, the row that used to strip it is gone, and the one test covering that path went with the row — so the compile matrix passed whether the codemod kept the attribute or stripped it with a TODO. Both directions are asserted now, each against the string that can show it: a TODO quotes the attribute it removed, so an absence checked on the raw output would pass either way.


allxsmith:
@coderabbitai review

allxsmith:
deep-review: fresh

The three advisory findings are addressed and the branch is now one commit (`d2e7b35`) rather than two, so this is a different tree from the one reviewed at `26d3d95`:

- the orphaned `/** The rows the type test holds to the library. */` left behind by the deleted export is gone
- the `describe` no longer calls the table an "exception", and the loop preamble no longer opens with "Both have to allow it" or explains a row deciding the TODO's wording
- the coverage figures are out of both the commit message and the PR body, replaced by the command to reproduce them

Worth a look at whether anything else still narrates the deleted mechanism — that was two of the three findings, and the class I keep missing. Copilot and CodeRabbit are unavailable on this PR, so this is the only review pass.

PR #684, opened 2026-09-16
refactor(bulma-ui): one source of truth for the anchor-only attributes

### LINKED ISSUE #682: [Refactor] bulma-ui: the anchor-attribute strip is implemented three times
## Affected Package(s)

- [ ] Other (please specify):

---

**Why this qualifies:** (c) maintainability. The rule "an anchor-only attribute must not reach a non-anchor" is implemented three times, each a different subset, and the divergence is invisible.

Found while reviewing #675.

## The three copies

| where | what it withholds | how |
| --- | --- | --- |
| `components/Dropdown.tsx` | `href`, `target`, `download`, `hrefLang`, `ping`, `referrerPolicy`, `media` | `LINK_ONLY_ATTRS` + `omitLinkAttrs` |
| `components/Menu.tsx` | `href` only | inline destructure, `isLinkLike` |
| `layout/Level.tsx` | `href`, `target`, `rel`, `download`, `hrefLang`, `ping`, `referrerPolicy` | named props, forwarded on the anchor branch only |

Each is correct for what its own props declare, and no two agree. `bestax-migrate` holds a fourth copy of the same list (`LINK_ATTRS` in `src/sources/_shared/polymorphic.ts`), which is the set it strips when a migration lands one on a non-anchor — so the codemod and the library can disagree about what a component forwards, and have.

## Why it matters beyond tidiness

The sets are not interchangeable, which is exactly what makes a shared helper worth having: today the right answer depends on whether the component derives its props from `as` (Dropdown) or enumerates them (Level), and that distinction is the thing a reader has to reconstruct from three implementations. When #672's remaining half lands and `Level.Item` starts deriving, its list has to change in step with its type — and nothing connects them.

`rel` is the interesting case. Dropdown deliberately does **not** withhold it, because React declares `rel` on `HTMLAttributes<T>` for every element, and rejecting it would diverge from React's own typing — the call #641 recorded for `Navbar.Link`. Level does withhold it, because it is a named prop there. A shared helper has to carry that reasoning once instead of leaving it to be rediscovered.

## Proposal

Lift `LINK_ONLY_ATTRS` and `omitLinkAttrs` from `Dropdown.tsx` into `src/helpers/`, with the per-component set passed in rather than assumed, and have Menu and Level call it. Keep `bestax-migrate`'s `LINK_ATTRS` as its own copy — it is a different package with no runtime dependency on this one — but cite the helper from it, the way its `HREF_OK` docblock already cites the components it mirrors.

Not breaking: every component keeps the behaviour it has. Worth doing before #672's second half rather than after, so that change edits one list.

## Acceptance

- [ ] One helper, per-component sets declared at the call site
- [ ] `Menu`, `Dropdown` and `Level` behaviour unchanged — assert each component's current set before and after
- [ ] The `rel` reasoning recorded once, where the helper lives
- [ ] `bestax-migrate`'s copy cites it rather than restating the rule

allxsmith:
Closes #682. Non-breaking: this moves no output.

## What the three copies actually did

Measured before touching anything — one probe rendering each component on a non-anchor tag with all nine attributes supplied:

| component | withholds | forwards to a non-anchor |
| --- | --- | --- |
| `Dropdown.Item` | 7 | `rel`, `type` |
| `Level.Item` | 9 | nothing |
| `Menu.Item` | 1 | everything but `href` |

Each was correct for what its own props declared, and no two agreed. A fourth copy of the same list lives in `bestax-migrate`, which is the set it strips when a migration lands one on a non-anchor — so the codemod and the library could disagree about what a component forwards, and have.

## What is shared and what is not

`helpers/anchorAttrs.ts` carries the part that is genuinely common:

- `AnchorOnlyAttributes` — subtracted from `React.AnchorHTMLAttributes` rather than listed. Listing is how `Level.Item` came to reject `download` at `as="a"` (#641): its list said `href`, `target`, `rel` and stopped. A list gains an entry when someone notices; a subtraction gains it when React does.
- `ANCHOR_ONLY_ATTRS` — the runtime mirror, `Record<keyof AnchorOnlyAttributes, true>` so the two cannot drift. Verified by deleting a key and watching `tsc` name it.
- `omitAttrs` — the filter.

The **sets stay per component**, because the differences are principled and were previously implicit:

- `Dropdown.Item` keeps `type` — `as="button"` is supported and `type="submit"` is valid there, so stripping it would remove a working attribute
- `Level.Item` adds `rel` — not an anchor-only attribute (React declares it on every element) but one this component has always withheld, so it is named separately rather than folded in
- `Menu.Item` strips `href` alone — its own props never declared the rest

`rel`'s absence from the derived set is the reasoning #641 recorded for `Navbar.Link`, and it now lives in one place instead of being rediscovered.

## Behaviour is unchanged, with one deliberate exception

The same nine-attribute probe gives byte-identical output for all three components before and after, and `docs/docs/api/**` plus `bestax-mcp/data/**` regenerate with zero drift. Every story of the three components renders byte-identical DOM too: all seventeen, diffed against `main` with the harness held fixed.

The exception is a class of prop name, named rather than hidden: one that collides with something on `Object.prototype`. `Level.Item` filtered with `key in ANCHOR_ONLY`, and `in` walks the prototype chain, so `toString`, `valueOf`, `constructor`, `hasOwnProperty` and `__proto__` were dropped although no caller ever named them. `Dropdown.Item` built its result by assignment, which swallows `__proto__` through the inherited setter, so it dropped that one too. All of these now reach the element, and React does emit them — verified, not assumed: `<Level.Item as="p" __proto__="x">` renders `<p class="level-item" __proto__="x">`.

That is the bug the shared helper exists to stop repeating, so the fix is the point rather than a side effect. `hasOwnProperty.call` for the strip lookup and `Object.defineProperty` for the result, each pinned by tests.

Two further input shapes where nothing renders differently, but "moves no output" is now literally true as well: non-enumerable own props stay dropped (`Reflect.ownKeys` reports them, every filter this replaced skipped them), and symbol-keyed props survive — which restores `Menu.Item`'s old behaviour and widens the other two, since React enumerates props with `for...in` and that never reports a symbol.

## What this surfaces

With the sets side by side, `Menu.Item` forwarding seven dead attributes to a `<span>` is visible rather than buried. Its type follows `as`, so a typed caller cannot deliver them — but a plain-JS caller or a loose spread can, which is exactly the reachability `Dropdown.Item`'s guard cites as its reason to exist.

Not fixed here. This change is meant to move no output, and closing that gap is a behaviour change that deserves its own decision.

## Verification

`pnpm all` green, `check:conformance` green, generated output unchanged, behaviour probe identical before and after.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved attribute handling for dropdown, menu, and level items rendered as links, buttons, or other elements.
  * Prevented link-only attributes from being applied to non-link elements while preserving supported attributes such as `type`.

* **Refactor**
  * Standardized link and non-link attribute handling across navigation components.

* **Tests**
  * Expanded coverage for link-only attributes, button attributes, symbols, and custom property handling.


allxsmith:
deep-review: fresh

All seven findings from the pass on `eb93928` are addressed, and the branch is now three commits:

- **1, 3, 5** — the three comments describing a sibling incorrectly. Menu no longer says `Level.Item` derives its props from `as` (it does not; that is #672's unlanded half), Level's pre-PR block is gone and its `Record` annotation restored at module scope, and the `bestax-migrate` mirror claim is corrected — that package works per attribute per element and shares neither `href` nor `type`.
- **2** — Dropdown's withholding test supplied three of seven keys, so four assertions were vacuous. It supplies and checks all seven now, plus that `type` is deliberately kept.
- **4** — the `type` carve-out's stated reason was narrower than the code; the docblock now says so and names selecting the set from the rendered element as the fix that would move output.
- **6** — `key in strip` walked the prototype chain. This was worse than "unreachable for DOM props": stripping `href` from six props returned one, and prototype pollution would delete props sharing its keys. `hasOwnProperty.call` now, the signature returns `Omit<T, K>` instead of asserting `T`, the record is `Readonly`, and the helper has its first tests — six, including that case.
- **7** — the acceptance item I wrote and did not do. `bestax-migrate`'s table now cites the helper, framed as what the relationship is rather than implying one derives from the other.

Worth a look at two things I found while fixing these and deliberately did **not** change, since this PR is meant to move no output: `Avatar` spreads its `type="button"` default before `rest", so `type={undefined}` through a spread erases it and the button submits (verified by render), and `Menu.Item as="button"` defaults no `type` at all. Both are the bug `Dropdown` fixed.

allxsmith:
deep-review: fresh

One finding last round, and it was right in a way worth naming: the `type` "whole difference" claim was **reported fixed and was not**. The script that corrected the two neighbouring comments aborted on that assertion and I never returned to it, so I told you it was done when the text was untouched. Fixed now, verified by grep rather than by assumption.

Checking it turned up a second problem in the same three lines. `STRIP_FROM_NON_ANCHOR: Record<string, true>` was checking nothing in either direction — a typo'd key passed, and because `omitAttrs` gained a `K` parameter this round, widening `K` to `string` made its return `Omit<T, string>`, so every key vanished from the result type. `Exclude<keyof AnchorOnlyAttributes, 'type'>` catches both; verified a typo now errors and the return type keeps its keys.

Still deliberately unchanged, and flagged rather than smuggled in: `Menu.Item` forwarding seven anchor-only attributes to a `<span>`, and `Dropdown.Item as="div"` rendering a JS-supplied `type`. Both pre-existing, both behaviour changes, and this PR's only safety claim is that it moves no output — `docs/` and `bestax-mcp/data/` still regenerate byte-identical.

Three things found while fixing these want their own issues once this lands, all the same bug class that `Dropdown` already fixed: `Avatar` spreading its `type="button"` default before `rest` so a spread erases it, `Menu.Item as="button"` defaulting no `type`, and the per-tag vs per-component `type` gap.

allxsmith:
deep-review: fresh

Six findings last round, all addressed, and the four stale threads are resolved — they were fixed in code but left open, which finding 6 was right to call out as blocking handover.

**1 (blocking)** — mine, and the third round running that a comment in this file family has claimed an agreement with `bestax-migrate` that does not hold. The pointer said this package "strips what bulma-ui would drop at runtime"; it does not, and the counterexample is the divergence the PR body documents — `<Menu.Item as="span" target>` loses `target` in the codemod and keeps it in the component. Now states where they disagree.

**2** — Level's set rationale sat in the render body describing a const seventy lines up, over an `if (Tag === 'a')` it did not describe, while that const pointed back at it. Reasoning moved to the declaration.

**3** — Menu's strip set was the only one without an annotation, so a typo there stripped nothing silently. Named and annotated like its siblings. I did not constrain `K extends keyof T`: the impl props types deliberately do not declare the anchor attributes, so that constraint would reject all three call sites — the observation is right, the fix would not compile.

**4** — Menu's documented gap is pinned in both directions now: the seven attributes that do reach a `<span>` are asserted to reach it, so a widening cannot land quietly.

**5** — `omitAttrs` walks `Reflect.ownKeys` rather than `Object.entries`, keeping symbol-keyed props that Menu's old rest-destructure kept. Nothing renders differently, but it was the one input shape where "moves no output" was not literally true, and now it is.

Unchanged and deliberate, as before: `Dropdown.Item as="div" type` and the Menu leak itself. Both pre-existing, both behaviour changes.

allxsmith:
Fixed 1, 2 and 3 in `84b5cd2`; 6 is now true rather than corrected.

**1 — stale rationale in the helper.** Caught this independently while auditing my own cross-component prose, so the fix predates reading the review, but the finding is right about why it mattered: it was the one file added to be the single place a reader looks. The three per-component sentences are deleted rather than rewritten. Each caller already states its delta beside the set it passes, and that copy is the one that stays true when a component's `as` changes. The `Level.Item` comparison in `Dropdown.tsx` lost its attribute count for the same reason and now names only the two directions.

**2 — `type` unasserted.** Supplied and asserted. Seven reach the `<span>`, seven are pinned.

**3 — non-enumerable own props.** Agreed, and it is the sharper half of the finding: reading keys through `Reflect.ownKeys` to keep symbols also started copying a shape rest-destructuring and `Object.entries` both dropped, re-publishing it as enumerable. Parity in one direction bought at the cost of the other. `propertyIsEnumerable` guard plus a test.

**4 — the `Omit` guard defeated by the cast.** Accurate, and not fixable at that site. `DropdownItemImplProps` deliberately does not declare the anchor attributes, which is the same constraint that made `K extends keyof T` reject all three call sites last round: there is no declared `type` for the `Omit` to remove, so reading one back cannot be anything but a cast. The guard the annotations do buy is on the strip sets themselves, where it binds — a typo or a widening to `string` fails the build. Leaving it, with the narrower claim now the one in the code.

**5 —** yes, both pre-existing, both deliberately unfixed here. Filing them against `Avatar`'s spread-order erasure and the per-tag vs per-component `type` split once this merges.

**6 —** the six/seven mismatch closes with finding 2 rather than needing a retraction: the body's claim is now what the test does.

allxsmith:
deep-review: fresh — review 84b5cd2 as it stands, not as a delta. Prior round was 2 blocking (stale cross-component rationale in `anchorAttrs.ts`, `type` missing from the Menu gap test) plus an enumerability regression in `omitAttrs`; all three are fixed here. Highest-value checks: that no cross-file claim in the five touched files is still false, and that `omitAttrs` is now byte-for-byte equivalent to each of the three filters it replaced across every input shape — enumerable, non-enumerable, symbol-keyed, prototype-colliding.

allxsmith:
Finding 1 fixed in `b5722f2`, 2 and 3 corrected, 6 corrected in the body.

**1 — `__proto__` swallowed by the inherited setter.** Real, and reproduced before fixing:

```
spread has own __proto__: true  | proto swapped: false
assign has own __proto__: false | proto swapped: true
```

`Object.defineProperty` now, which is what a spread and a rest destructure both do. That is the fourth input shape where the parity claim failed and the third found by this review series, after prototype-chain keys in the strip set, non-enumerable own props, and symbols. Pinned by a test that asserts both halves: the key survives as an own prop, and the result's prototype is untouched.

**2 —** correct, and my error. `Level.Item` renders `div`, `p` or `a` and defaults no `type` either, so that separated `Menu.Item` from `Dropdown.Item` alone, not from both siblings. The comment now names what each sibling actually strips instead of counting differences, which is also what stops it going stale a fourth time.

**3 —** also correct and worth stating inline rather than leaving on the record, so I did. The switch is not one change: Menu filtered by rest destructure and kept symbols already, Dropdown and Level filtered through `Object.entries` and did not, so it restores one caller and widens two. No DOM effect either way — React enumerates props with `for...in`, which never reports a symbol.

**4 —** unchanged and still declined, same reason: `DropdownItemImplProps` deliberately does not declare the anchor attributes, so there is no declared `type` for `Omit` to remove and reading one back can only be a cast. The guard the annotations buy is on the strip sets, where it does bind.

**5 —** agreed, out of scope, and filing both once this merges.

**6 —** body corrected. It now names the one deliberate behaviour change — `Level.Item`'s `key in ANCHOR_ONLY` walked the prototype chain, so a prop named `toString`, `valueOf`, `constructor` or `hasOwnProperty` was dropped though no caller named it, and those now render — and separates it from the three shapes that move no output.

Also verified the flagship package's Storybook, since none of the rounds so far had: every story of the three components renders byte-identical DOM against `main`, all seventeen, with the harness held fixed. Honest limit on that result — no story passes an anchor-only attribute to a non-anchor tag, so it covers the paths users see and leaves the strip branches themselves to the unit tests. The static build also loads clean in a browser with no React unknown-prop warnings.

allxsmith:
deep-review: fresh — review b5722f2 as it stands. Prior round's single blocking finding (`__proto__` swallowed by the inherited setter in `omitAttrs`) is fixed by defining rather than assigning, plus two prose corrections and a PR-body correction. Highest-value checks: whether any input shape remains where `omitAttrs` is not equivalent to the three filters it replaced, and whether any claim in the touched comments is still false.

allxsmith:
Converged at 0 blocking. Closed 1 and 3 anyway in `3996294`; 2 and 4 stand on the record as written.

**1 — ambiguous pronoun.** Right about the defect, and I did not take the suggested edit: deleting `bar type` would leave "`Dropdown.Item` strips the anchor set", which is false in the other direction, since it keeps `type`. Both halves now name what they strip rather than pointing at each other.

**3 — `__proto__` in the wrong list.** Correct, and verified rather than accepted: `<Level.Item as="p" __proto__="x">` renders `<p class="level-item" __proto__="x">`. So it belongs with the `toString` exception, not with the shapes that move no output. The body now states the exception as a class — a prop name that collides with something on `Object.prototype` — and records that `Level.Item` dropped all five through `key in`, while `Dropdown.Item` dropped `__proto__` alone through assignment.

**2 —** accepted as stated. The `Omit` is a type-level no-op at two of the three call sites and defeated by a cast at the third, so the guard that actually holds is the key-typed annotation on each strip set, where a typo or a widening to `string` fails the build. Three rounds on the record is the right resting place for it: constraining `K extends keyof T` would reject all three call sites, because the Impl props types deliberately do not declare the anchor attributes.

**4 —** unchanged, out of scope, and filing both once this merges.

Good catch on the `Object.create(null)` alternative, for the record — a null-prototype props object is not what any of the three replaced filters produced, so that would have traded one divergence for another.

PR #699, opened 2026-09-18
test(bulma-ui): compare the colour guard per element, and assert the shade partition

allxsmith:
# Pull Request

## Description

Follow-up to #686. Two findings from its last review round landed after the merge, so they missed it. Both are about the colour guard that PR added, and both were raised against my own work.

- [x] Other (please specify): **`eslint-plugin/src/lib/values.ts` comment, `scripts/color-tuple-css.test.mjs`**

**The guard was exact by luck.** `color-tuple-css.test.mjs` asked whether a colour was dead on **all three** of `.notification`, `.progress` and `.hero`. That only holds while the three agree: a colour Bulma ships on one and not the others lands in neither set, the equality passes, and that one renders a dead modifier with no development warning. The comparison is per element now, and the failure names which one disagreed.

**The partition runs deeper than the modifiers.** The colours the stylesheet *shades* are exactly the colours with a live component modifier: `has-text-white-bis-15` does not exist, while `has-text-white-15` and `has-text-light-15` both do. That is asserted for every shade in both families, with ambiguity excluded per *pair* rather than per shade, since `grey` plus `light` is the colour `grey-light` and no shade is ambiguous in general. `inherit` and `current` are asserted alongside them: live unshaded, dead shaded.

The consequence for the lint rule is recorded rather than fixed. `HELPER_VALUES` judges each prop alone, so `<Box textColor="white-bis" colorShade="15" />` passes and renders a class that does not exist. Pre-existing for the greys and the black pair, widened by two when the white shades were added, and a false negative either way. A cross-prop check would close most of it off the partition this guard asserts, but not all: `inherit` and `current` are outside `validColors`, so the rule would need those named as well.

**On the shape of the guard**, since the diff is mostly this. The element list comes from one signal, the `warnUnstyledColor` call sites. An earlier version inferred it by matching a `color` prop typed off the tuple *and* a class emission in source text, then cross-checked the two signals; that needed a reader per signal and gave each its own way to fail while the library was fine. Everything else compared here is data — tuples against stylesheet rules. What the single signal gives up is stated in the file: a component that emits the modifier and never warns is not detected, which already applies to `Calendar` and `TimeWheels`.

Verified against the defects it exists for rather than assumed: `white-ter` missing from `validColors` fails six assertions, the white shades missing from `UNSTYLED_MODIFIER_COLORS` fails eight, `Hero` dropping a keyword it accepts fails four, and the two accepted-set copies drifting fails six.

## Related Issue(s)

Follow-up to #686.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Performance
- [x] Build tooling
- [x] Other (please describe): **test hardening, plus one comment recording a known limitation**

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated Storybook stories as needed (bulma-ui)
- [ ] My changes require a change to the documentation
- [ ] Any relevant dependencies are updated
- [ ] If this PR changes commands, conventions, or package structure, the affected `CLAUDE.md` files are updated

## Additional Context

No library behaviour changes. The only source edit outside the test is a comment in `HELPER_VALUES`, so the `test(bulma-ui)` scope releases nothing.

Both guards were confirmed to fail on the defect they describe, not just to pass as written: the per-element one by removing a colour from `UNSTYLED_MODIFIER_COLORS` while leaving it in `validColors`, and the shade one by the measurement that produced it. `pnpm all` is green, three `Tasks:` summaries.

## Summary by CodeRabbit

* **Documentation**
  * Expanded guidance for validating text and background color shades.
  * Clarified handling for unsupported combinations, inherited and current colors, and scheme colors without shades.
  * Documented limitations involving multiple color properties.

* **Tests**
  * Broadened color validation across components, text and background families, and all supported shades.
  * Added coverage for special color values and unsupported styling combinations.
  * Improved source validation and error reporting for more reliable results.


allxsmith:
deep-review: fresh — please read f9fd3685

allxsmith:
Round 1 handled. All three were mine and all three are fixed.

**2, the hand-maintained element list.** Fixed, and deriving it closed more than the finding named. Two conditions identify a modifier component together, and both are needed: a `color` prop typed off `validColors` AND a `` `is-${color}` `` emission. The components each condition excludes are the argument for using both — `Section`, `Footer`, `Level`, `Media` and the `Card` parts type `color` off the tuple and emit no modifier, while `Button` declares its own narrower union.

The hole worth naming is the one no list could have caught: a component emitting the modifier without calling `warnUnstyledColor`. Deriving the set makes that checkable, so each element is now held to calling it. Confirmed by removing the call from `Notification`, which fails and names the file. I also assert the component-name-to-class assumption, because a derived name with no `.<class>.is-primary` rule would call every colour dead and fail the comparison for the wrong reason.

**1, the missing guard.** Fixed. One `stylesheet()` helper that every case goes through, rather than the guard living in two of the three. Confirmed the way you found it: running that case alone with the file moved aside now prints "Build first, or run `pnpm all`" instead of a raw `ENOENT`.

**3, the prose and the probe.** Fixed both halves. `has-background-white-bis-15` is equally absent, so the check runs over both families rather than `has-text-` alone, and the `HELPER_VALUES` comment names the pair instead of only `colorShade`. Worth saying that the probe being one-sided was the more useful half of this finding: the comment was imprecise, the test was incomplete.

Gate green, three `Tasks:` summaries. Nothing outstanding from this round.

Worth a look next:

- Whether the derivation is too clever. It reads source text for two patterns, and a component that emits the modifier some other way would be missed as silently as the list missed one. I think the `warnUnstyledColor` assertion is the backstop for that, but it is the judgement I would most like challenged.
- `stylesheet()` asserts rather than skips, which is the house position, but it means these cases fail rather than skip on an unbuilt tree. That is deliberate and matches `constants-subpath.test.mjs`.

allxsmith:
deep-review: fresh — please read 8ebeef52

allxsmith:
deep-review: fresh — please read c7a55d27

allxsmith:
deep-review: fresh — please read 697ae3b9

allxsmith:
deep-review: fresh — please read 29c3f885

allxsmith:
deep-review: fresh — please read 915eb787

allxsmith:
deep-review: fresh — please read e57d5305

allxsmith:
deep-review: fresh — please read dcd457a2

allxsmith:
deep-review: fresh — please read 9617cf1b

allxsmith:
deep-review: fresh — please read 873eb7ee

allxsmith:
deep-review: fresh — please read 5638ee33

allxsmith:
deep-review: fresh — please read e4e1b1a1

allxsmith:
deep-review: fresh — please read 6fb7a45d

allxsmith:
deep-review: fresh — please read 9e722bd9

allxsmith:
deep-review: fresh — please read bb60a791

allxsmith:
deep-review: fresh — please read 181bd399

allxsmith:
deep-review: fresh — please read 1cb026ea

allxsmith:
deep-review: fresh — please read b7c5426d

allxsmith:
deep-review: fresh — please read b7c5426d

allxsmith:
deep-review: fresh — please read b7c5426d

allxsmith:
deep-review: fresh — please read b7c5426d

allxsmith:
deep-review: fresh — please read b7c5426d

PR #703, opened 2026-09-19
fix(bulma-ui): a compound matches one simple selector, not co-occurrence

allxsmith:
`#699` landed the colour guard with a matcher that answered co-occurrence rather than compounds, so `.hero.is-primary .tabs` answered the query `tabs.is-primary` — a dead modifier reading as live, which is a missed defect rather than a false alarm. This PR carries the fix that missed that merge, and then everything twelve rounds of review found behind it.

## What the guard does

`scripts/color-tuple-css.test.mjs` holds the colour tuples in `bulma-ui` to the stylesheet the library ships, in both directions: a colour the CSS has and `validColors` lacks is dropped silently, and a colour in the tuple whose component modifier has no rule renders a dead class. To ask either question it has to answer "does the stylesheet ship CSS for this compound", and that is what the matcher does.

## The invariant

**Reading a compound as live when nothing renders is a missed defect. Reading it as dead when something does is a false alarm.** The guard is only ever allowed the second. Every finding below is a case where it was allowed the first, and every fix moves it back.

The mechanism is one rule: a simple selector's class set must account for *everything* the selector requires, and a query matches only on an exact size. So anything the tokeniser cannot read as a class has to still be counted, or a rule needing three classes answers a two-class query.

## What each round found

| | |
|---|---|
| the compound itself | a class set is per simple selector, not per comma-part |
| `:not()` | its classes are ones the element must NOT carry |
| `:is()`, `:where()`, `:has()` | these ADD a requirement, so deleting the argument shortens the set |
| `:first-child`, `:hover`, `[disabled]` | the same requirement without parentheses, and it was not being counted |
| `a.dropdown-item` | a type or ID selector is a requirement the tokeniser sees only as a leftover |
| `.hero .tabs` | a combinator is a requirement, and only the SUBJECT of a rule is styled by it |
| `@media`, `@supports`, `@container`, `@scope` | a rule that applies sometimes is not a rule that applies |
| `.hero{color:red;.tabs{…}}` | native nesting, read off the brace stack rather than the punctuation |
| `.is-gap-0\.5` | an escaped dot is part of the name, not a separator |
| `content:"}"` | a string is content, and none of its characters are structure |
| `:not(:nth-last-child(1))` | resolving arguments to a fixed point made a requirement look class-shaped |
| `ONLY_CLASSES` | a quantifier inside a quantifier, and CodeQL was red on it |

Two questions turned out to be hiding inside one index and are now separate. Whether a rule **styles** a compound reads the subject sets with the exact-size rule. Whether the stylesheet **knows** a name reads every name mentioned anywhere, prohibition included — answering that one from the subject sets hid a class that only ever appears as an ancestor, and a colour the CSS ships and the tuple lacks is the first defect this file was written for.

## How it is held

The matcher has a test of its own, on selectors written for the purpose rather than on whatever Bulma happens to emit. That matters because every finding here was a matcher bug rather than a library one, and each was found by measuring the built stylesheet, which can only cover shapes already in it — there is no `:is()` or `:where()` in Bulma at all, and some of what does ship comes from our own SCSS.

Each of the twenty-one behaviours was reverted separately, and each fails the suite on its own assertion. Three assertions were found along the way to be passing without reaching what they named, all three because a single-class query takes the membership path and never sees the exact-size rule.

## Risk

Every fix here is latent against the stylesheet as it ships: **no answer this guard asks moves**, across the element-and-colour pairs and the helper-class sweeps alike. Two independent selector parsers were used in review to difference the whole index over all shipped selectors, and neither found a compound the matcher calls live that the parser does not.

## One decision for the reviewer

The diff touches `scripts/color-tuple-css.test.mjs` and nothing else — no file under `bulma-ui/`, nothing in any tarball. `fix(bulma-ui)` therefore cuts a patch release of `@allxsmith/bestax-bulma` whose changelog describes a matcher no consumer can see, because `releaseRules` key off the commit scope and never the path. **Squash-merging with a `test(bulma-ui):` subject releases nothing**, which is what I would do; the commit scopes below a squash do not matter, and this is a single-package branch, so the usual reason to avoid squashing here does not apply.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CSS selector parsing for scoped styles, nested rules, conditional blocks, and quoted content.
  * Corrected handling of nested selectors and attribute selectors to produce more accurate results.
  * Improved class-name detection, including names defined in ancestor and prohibited selectors.
  * Added safeguards for declarations containing selector-like text.

* **Tests**
  * Expanded regression coverage across scoped, nested, quoted-string, declaration, and prohibition scenarios.


allxsmith:
Taken, with the measurement.

`:not(…)` names classes the element must not carry, and reading them as present would let `.notification:not(.is-light)` answer the query `notification.is-light` — a rule that explicitly excludes the pair reporting it live. That is the silent direction, which is the one this PR is about, so it is worth closing even though it changes nothing today: 201 simple selectors in the stylesheet carry a functional pseudo-class and none of them also carries a colour modifier, so no answer moves.

`:is()` and `:where()` are left alone on purpose rather than overlooked. Their arguments are alternatives, not requirements, so counting their classes makes the exact match demand too many and under-report — the safe way to be wrong. Modelling them as branches means writing a selector parser, which is more machinery than this file should carry and the kind that has gone wrong here repeatedly.

allxsmith:
deep-review: fresh — please read 1ba7bdf8

allxsmith:
deep-review: fresh — please read 6a2daead

allxsmith:
Round 2 handled. Findings 1, 2 and 3 are fixed at `f395e76d`; finding 4 I want to push back on.

**1 (blocking) — fixed, and you were right about the direction.** The strip now branches: `:not()` drops its argument, everything else substitutes a sentinel class. `:not()` is a prohibition, so the rest of the selector already names what the element carries; every other functional pseudo-class adds a requirement, and deleting its argument shortens the set, which is exactly the over-reporting the exact-size rule exists to stop. The comment justifying it said the opposite and is rewritten.

The three shapes now behave as you asked, checked against the previous matcher:

| selector | query | before | after |
|---|---|---|---|
| `.notification.is-primary:where(.is-light)` | `notification.is-primary` | live | dead |
| `.title.is-spaced:has(+.subtitle)` | `title.is-spaced` | live | dead |
| `.notification.is-primary:nth-child(even)` | `notification.is-primary` | live | dead |
| `.notification:not(.is-light)` | `notification` | live | live |
| `.navbar-item:not(.is-active,.is-selected)` | `navbar-item.is-active` | dead | dead |

Six selectors in the built stylesheet gain a sentinel, all structural. No element-and-colour answer moves.

**2 — fixed.** Escaped names are read whole, so `.is-gap-0\.5` is one class rather than `{is-gap-0, 5}`. Twenty-four selectors change, no queried answer does.

**3 — fixed.** Comments are removed before the split. The five single-class answers that change are exactly the words that were coming out of the banner and the sourceMappingURL.

**4 — refuted, and I think the cost is attributed to the wrong thing.** The scan is not what the first case spends its time on:

- index build: 6.4 ms, once, cached per stylesheet
- 400 compound queries: 1.3 ms total
- reading and splitting the 801 KB stylesheet: under 2 ms

The first case is ~340 ms because `warningCallers()` walks `bulma-ui/src`, reads every `.ts`/`.tsx` and comment-strips each one with two regexes. An inverted index over class names would optimise something that costs three microseconds a query, and it would add a second structure to keep in step with the set list. Not worth it; if this ever needs to get faster, the file walk is the thing to cache.

One addition that is not a finding: the matcher now has a test of its own, on selectors written for the purpose. Every finding this file has taken has been a matcher bug rather than a library one, and each was found by measuring the built stylesheet, which only covers the shapes Bulma writes today. There is no `:is()` or `:where()` in it at all, and two of the four `:has()` selectors come from our own SCSS, so the shape is ours to grow. Reverting each of the three behaviours separately fails the new test on that behaviour's own assertion.

allxsmith:
deep-review: fresh — please read f395e76d

allxsmith:
deep-review: fresh — please read bc056225

allxsmith:
deep-review: fresh — please read 64f3d7ca

allxsmith:
deep-review: fresh — please read 70599570

allxsmith:
deep-review: fresh — please read 4274e47d

allxsmith:
deep-review: fresh — please read 662c5dd1

allxsmith:
Round 7 handled. Findings 1, 2 and 3 are fixed; 5 I want to push back on again, with numbers; 4 is Alex's call and is the one thing here I cannot settle.

**1 — fixed, and it is the second time I have done this.** The pseudo-element assertion used a single-class query, which takes the membership path and never reaches the exact-size rule, so it passed whether `::` was stripped or sentinelled. It is now a compound. The same mistake made the nested-`:not()` assertion vacuous three rounds ago, so the distinction is now written down in `shipsClass` rather than left to be rediscovered: a bare class is answered by membership, a compound by exact size, and only the second can tell a requirement from its absence.

I re-ran the revert check over every behaviour in the matcher, one at a time, and each now fails the suite on its own assertion:

| reverted | fails on |
|---|---|
| `:is()`/`:where()` direction | `:where()` adds a class the element must carry |
| comment strip | a comment against the selector that follows it |
| escape-aware tokenising | an escaped dot read as a separator |
| bare pseudo-class + attribute pass | a structural pseudo-class, argument or not |
| fixed-point loop | a prohibition stays a prohibition |
| type/ID qualifier | a type selector means the rule needs that element |
| `:not()` over non-classes | a prohibition over a POSITION |
| combinator | a compound that renders only inside an ancestor |
| subject-only indexing | a rule styles its subject |
| `:not()` leaving a `*` | a `:not()` that IS the simple selector |
| pseudo-element exception | a pseudo-element rule renders something |
| `:not()` over a compound | a COMPOUND of classes is still classes |

**2 — fixed.** `:not(.is-a.is-b)` drops now; the check read one class per part where the rule is "classes and nothing else". A descendant inside the argument still takes a sentinel, deliberately, and both halves are pinned. Nothing ships that shape (0 of 262 `:not()` occurrences), so this is a rule matching its description rather than a behaviour change.

**3 — fixed.** It leaves `:has(*)`, not an empty `:has()`, since the prohibition started leaving a universal selector behind.

**5 — refuted, with the same measurement as round 2 and it has not moved.** The scan is not where the time is:

- index build: **8 ms**, once, cached per stylesheet
- all 1,235 queries: **3.5 ms** total
- the whole file: **555 ms**, of which **333 ms** is the first case

That 333 ms is `warningCallers()` walking `bulma-ui/src`, reading every `.ts`/`.tsx` and comment-stripping each with two regexes. An inverted index would optimise three milliseconds and add a second structure to keep in step with the set list. The headroom is not gone; it was never in the scan. If this file ever needs to be faster, the source walk is the thing to cache.

**4 — this one is Alex's.** The branch touches `scripts/color-tuple-css.test.mjs` and nothing else, so `fix(bulma-ui)` cuts a patch of `@allxsmith/bestax-bulma` whose changelog describes a matcher no consumer can see. My recommendation is to squash-merge with a `test(bulma-ui):` subject, which releases nothing. The individual commit scopes below a squash do not matter, and this is a single-package branch, so the usual reason to avoid squashing here does not apply.

allxsmith:
deep-review: fresh — please read 4f3b7c2d

allxsmith:
deep-review: fresh — please read 697f9dd3

allxsmith:
deep-review: fresh — please read cb8d7c3e

allxsmith:
deep-review: fresh — please read 4e4076e8

allxsmith:
deep-review: fresh — please read 4e4076e8

allxsmith:
deep-review: fresh — please read caf5190d

allxsmith:
deep-review: fresh — please read caf5190d

allxsmith:
Round 13 handled. Findings 1, 2 and 3 are fixed; 4 was a true statement about the PR body that had stopped being true, and is now true again; 5 I am declining, with a reason.

**1 — fixed, and you were right that it was the same trap a fourth time.** The stray `}` sat *before* the rule under test, where it lands on an already-empty stack and changes nothing. The assertion now puts it where a leaked brace actually does damage — closing a block that is still open, so the rule nested inside it stops looking nested:

```
.hero{content:"a\"b}";.tabs.is-boxed{color:red}}   →   tabs.is-boxed must read dead
```

With `"[^"]*"` the string ends at the escaped quote, `b}"` leaks back into the text, the `.hero` block closes early, and `.tabs.is-boxed` answers live. Verified by mutation: swapping the pattern now fails on that assertion.

**2 — fixed.** `@container` was the only one of the four with no assertion and the only one the stylesheet ships. Deleting it from the list now fails.

**3 — fixed.** Both comments described the attribute pass that went with `caf5190d`. Two assertions had also converged on testing the same thing once strings were emptied globally, so one of them is gone.

**4 — the claim is accurate again.** All twenty-one behaviours plus the two above were reverted individually just now, and each fails the suite on its own assertion. I have corrected the PR body rather than softening the claim, because the claim is the point.

**5 — declining, and this is a real trade rather than a dismissal.** An at-rule prelude contributes nothing to the membership index, so a class named only as a `@scope` root would read unknown. Adding preludes to that index would fix it and would also feed every media condition through the class tokeniser, where a length like `1.5rem` tokenises as a class called `5rem`. That is a fabricated name, and a fabricated name is the one direction that quiets the `assert.ok` requiring a keyword to ship. Trading a false alarm for a silent pass is the wrong way round for this guard, and `@scope` ships nowhere in the stylesheet or our SCSS. If it ever does, the fix is to parse the prelude properly rather than to tokenise it.

On method: your independent character-level tokeniser agreeing on all 646 two-class live answers and matching the membership index exactly, 0 fabricated and 0 missed, is the strongest evidence this branch has produced. That plus the mutation results is what I would want a maintainer to weigh, more than anything in my commit messages.

allxsmith:
deep-review: fresh — please read fe5696b4

allxsmith:
deep-review: fresh — please read dc1ccfff

allxsmith:
deep-review: fresh — please read 4c64338d

allxsmith:
deep-review: fresh — please read 4c64338d

allxsmith:
deep-review: fresh — please read 4c64338d

allxsmith:
deep-review: fresh — please read 4c64338d

allxsmith:
Round 16 handled, and it is the round that earns the loop. Two of its findings were real defects and one of them refuted a claim I had written into the file.

**1 — real, and the fifth door into the same hole.** `.hero.is-primary ::before` answered `hero.is-primary` live. The `::` pass was the last normalisation here that deleted text instead of standing something in its place, and a pseudo-element can BE the whole simple selector, so removing it left the ancestor holding the subject position. It leaves a `*` now, the same fix the prohibition got. `.delete::before` and `.notification.is-primary::before` both still read live, which is the point of the exception.

**2 — real, and I was wrong in writing.** The previous commit added a paragraph saying the comment-before-strings order "lands on the loud side". It does not. I searched constructed payloads and found twelve where it reads live, the shortest being:

```
@charset "/*";/* */.hero{color:red;.tabs.is-boxed{color:red}}
```

The span from the `/*` inside the string to the next `*/` swallows the `{` that opened `.hero`, so the stack ends up shallower and `.tabs.is-boxed` stops reading nested. Comments, strings and url tokens are now emptied in one left-to-right pass, because the order between them is not decidable — a `/*` inside a string is not a comment, and a `"` inside a comment is not a string. Whichever starts first wins, which is what CSS does.

That pass also took the unquoted `url()` token from your residual list. `url(x}y)` is legal and closed a block exactly the way `content:"}"` did.

**3, 4, 5 — fixed.** The single-quoted half needed an escape case per side, for the same reason the double-quoted half did. `onlyClasses` reads its argument a comma-part at a time and nothing held that. The class-name unescape is reachable only through `rendersAlone`, because a compound query spells itself with a `.` and cannot ask for a class containing one.

**6 — fixed.** `shipsClass` still described the caller that moved to `rendersAlone`.

**7 — fixed in the PR body**, and I dropped the count rather than update it, since it has gone stale four times.

One note on your mutation sweep, because it matters for the next round: reordering the alternation inside the combined pass is not a live mutant. The constructs start on different characters, so a single left-to-right scan cannot prefer one over another at the same position. The mutant that does bite is the two-pass form, and that is now killed by the `@charset "/*"` assertion.

Nine real mutants tested individually against this head, each killed by its own assertion: both escape halves in both quote styles, the two-pass ordering, the url token, the pseudo-element substitution, the comma split, and the unescape.

allxsmith:
deep-review: fresh — please read 5fb3249f

allxsmith:
Round 17 handled, plus CodeRabbit's attribute finding, plus the thread sweep it asked for.

**1 — all three spans fixed.** Each made the brace stack shallower, which is a nested rule reading unnested:

| span | why it leaked |
|---|---|
| `URL(x}y)` | the url alternative matched its keyword case-sensitively |
| `url(x\)y}z)` | no escape branch, so an escaped close paren ended the token early |
| `"x\`<newline>`y}"` | the string bodies used `\.`, and `.` does not match a newline |

The string bodies take `\[\s\S]` now, the url branch takes the same escape form and the `i` flag, and the dispatch compares the keyword case-insensitively.

**2 — both sentences corrected.** The `::` exemption covers a bare `::` spelling, not a functional one: `::part(x)` is taken by the argument pass first, because that pass matches any name before a parenthesis, and it lands on the false-alarm side. And the pseudo-element pass was not the last one here that deletes — the paren-residue loop deletes too, and is allowed to, because by then the enclosing call's name is already a sentinel outside the parentheses.

**3 — fixed rather than accepted.** An escaped brace is part of a class name and neither the split nor the count of closes could see it, so `.a\}b` popped the block the next rule was nested in. Both skip an escaped brace now. The `{` half of that only ever cost a false alarm, but it cut the class name in half, which `rendersAlone` can see.

**CodeRabbit's unquoted-attribute finding was right and I had reasoned my way past it.** Two commits ago I removed the attribute strip from the membership scan, arguing a quoted value is emptied with every other string and an unquoted one is an identifier that cannot contain a dot. An identifier may *escape* a dot, so `[data-x=\.fake-class]` is legal and the scan read `fake-class` out of it — while the quoted spelling was pinned at dead. Two spellings of one attribute disagreeing. The strip is back on the membership path, which is the half that needed it: on the sets path a fabricated class makes the set bigger, so the exact-size rule refuses it.

**Thread sweep.** All 29 open threads named defects I could not reproduce at this head, so I built one probe covering every distinct behaviour they name — 42 cases across the parenless requirement, type and ID qualification, the combinator and the ancestor, the prohibition as subject, the pseudo-element as subject and as exception, statement at-rules, nesting, conditions on the whole stack, all four conditional at-rules, `@layer` as a non-condition, braces and semicolons in strings and url tokens, every escape shape in both quote styles, the comma split, the unescape, and both attribute spellings. **All 42 answer correctly.** On that basis the threads are resolved rather than left for the gate to count.

Nothing is outstanding from any round. The one thing I cannot settle is the release scope, which is in the PR body and is yours: squashing with a `test(bulma-ui):` subject releases nothing, and `fix(bulma-ui)` cuts a patch of `@allxsmith/bestax-bulma` with no library delta in it.

allxsmith:
deep-review: fresh — please read 8cbda8ef

allxsmith:
deep-review: fresh — please read acbd9629

allxsmith:
deep-review: fresh — please read c71a2358

allxsmith:
deep-review: fresh — please read 72d82a3e

allxsmith:
Round 21 handled. All six, and one note on method.

**On method first, because it matters here.** You disclosed that this session had no command execution and that `dist/bestax.css` was absent, so nothing in the round was measured. I re-derived all four testable claims locally against a built stylesheet rather than taking them, and **all four reproduced**:

| claim | reproduced |
|---|---|
| `@scope(.\31 23)` registers `31`, loses `123` | yes |
| unterminated `[` loses the names after it | yes |
| `@scope-foo (1.5rem)` harvests `5rem` | yes |
| `the-root.box` passes with or without the prelude in `sets` | yes |

That is the right way round: an unmeasured review that turns out to be right on every count is worth more than a measured one that hedges.

**1 — real, and fairly placed.** The harvest is the one path here that reads class names outside the style-rule path, and it inherited neither guard. The numeric-escape check now lives in one constant with both readers using it, and the harvest throws on `@scope(.\31 23)` rather than renaming the root to `31`.

**5 — real.** `\b` holds before a hyphen. `/^@scope(?![\w-])/i` now, so `@scope-foo` keeps its condition out of the index.

**2 and 4 — you were right that three assertions read as though they held something.** The `only for names` half is pinned by `rendersAlone` now, since a compound query cannot see it: the prelude's classes never all appear in one query. The two attribute shapes were compound queries too, and the compound path reads the raw heading, so neither ever reached `withoutAttributes` — the name-level behaviour has its own case now.

**3 — the comment was wrong twice over.** Only the membership text is shortened, so what an unterminated bracket loses is a name, not the size of a set, and I had called that the loud side. It is the quiet side by index — but it is also *correct*, which is the better reason: a selector with an unclosed bracket is a parse error, the browser drops the rule, and none of those classes ship. The comment says that now, and the behaviour has an assertion.

**6 — deduplicated.** The copy asks about `@container`, which is what it was really for.

Five mutants tested individually: the boundary, the harvest guard, the names-only restriction, and the unterminated-bracket name behaviour each fail on their own assertion.

allxsmith:
deep-review: fresh — please read aaa27505

allxsmith:
deep-review: fresh — please read d50b83bb

allxsmith:
deep-review: fresh — please read d50b83bb

allxsmith:
Round 23 handled, and it caught me pinning a wrong answer.

**1 — real, and the assertion was holding the error in place.** `CONDITIONAL_AT_RULE` named the conditions, which waves through every at-rule the file has not heard of — and a browser that has not heard of one drops it *with its block*, so nothing inside renders. `@media-foo`, `@page` and `@starting-style` all read live. Worse, the commit before this one added an assertion pinning `@media-foo` at **live**, so I had written the wrong direction down and fixed it in place.

The list is inverted now: `@layer` is the only at-rule whose contents are ordinary style rules that always apply, so it is the whole list, and everything else is assumed a condition. `@page` and `@starting-style` have cases.

**Both open threads — real, and both fixed.**

The escaped quote reproduced exactly as you described. The strip runs over raw text, so `.a\"b` opened a string at its own quote and the match ran to the next quote in the file, emptying the brace of the rule in between. The strip consumes an escape pair first now — the same rule the delimiter scan already had, and the one place still missing it.

The escaped `url` identifier I am closing by refusing rather than decoding. There is one numeric-escape check over the whole stylesheet now, running after strings, comments and url tokens are emptied, so what it sees is selector and declaration text: `\75 rl(` stops the guard with a message instead of leaking its brace. Bracketed spans come out first, because an escape in an attribute value renames nothing and halting there would be a false alarm — which is the exemption round 21 asked for, kept.

**2 — fixed, and generalised.** `@keyframes f{33.3%{…}}` offered `.3` to the tokeniser and got back the name `3`. A CSS identifier cannot begin with a digit, so anything that does was never a class; filtering on that closes the shape wherever it appears rather than teaching one reader about keyframes. Loud direction, so it is tidiness — the index over the shipped stylesheet is unchanged at 3553 names, and none of them ever began with a non-letter.

Four mutants tested individually, each killed by its own case: the inverted at-rule list, the escape branch in the strip, the structural numeric check, and the identifier-start filter.

One correction to my own method, since it bit twice this round: two of my probe payloads came back wrong because the shell ate a backslash on the way into the file, and I nearly reported working code as broken on that basis. Anything with an escape in it I now build with `String.fromCharCode(92)` rather than a literal, and both of those cases turned out to be correct all along.

allxsmith:
deep-review: fresh — please read 7d15786d

allxsmith:
deep-review: fresh — please read 89d7a3f6

allxsmith:
deep-review: fresh — please read 1d08e11f

allxsmith:
deep-review: fresh — please read 1608f1e9

allxsmith:
deep-review: fresh — please read 3965c04e

allxsmith:
deep-review: fresh — please read 3965c04e

allxsmith:
deep-review: fresh — please read ce9b565e

allxsmith:
deep-review: fresh — please read 90820242

allxsmith:
deep-review: fresh — please read 0edfb47c

allxsmith:
deep-review: fresh — please read 0edfb47c

allxsmith:
deep-review: fresh — please read 5557f53a

allxsmith:
deep-review: fresh — please read b4e03778

allxsmith:
deep-review: fresh — please read b4e03778

allxsmith:
deep-review: fresh — please read 484498a2

allxsmith:
deep-review: fresh — please read e24dbc6c

allxsmith:
deep-review: fresh — please read e24dbc6c

